PR #1, opened 2016-06-07
bugfix: chunk compression should use actual size of chunk

funkey:
The actual size of a chunk can be smaller than the one reported in the
VolumeChunkSpecification, if the requested chunk was not fully contained
in the volume.


funkey:
I signed the CLA.

funkey:
I added the commit e-mail address as alternate e-mail address.

PR #2, opened 2016-06-14
ndstore backend: set cube_dimension from info as chunkSize

funkey:


funkey:
I am not 100% sure if the cube_dimension field in the ndstore info API is a suggestion for efficient chunk access. However, the [example response](http://docs.neurodata.io/ndstore/api/info_api.html) seems to be in line with that.

jbms:
There is a lot to say about this, so I wrote up a document describing how the SliceView works (much of which you may have already figured out) and the trade-offs related to the chunk size.  It is in src/neuroglancer/sliceview/README.md

What it comes down to is that I don't think it is a good idea to, by default, set the chunk size to the cube_dimension provided, since there is no guarantee that it will be an efficient size for Neuroglancer to use.  It could be used as a guideline, though, in that Neuroglancer would attempt to align its chunks to fall on cube_dimension boundaries when reasonable.

If we are talking about a server specifically for the purpose of providing data to Neuroglancer, like your server for the CREMI website, or something like the prototype server in python/neuroglancer.py, then obviously the cube_dimension could be chosen correctly for Neuroglancer.

Another option is to provide the set of chunk sizes to use as a query string parameter included in the data source URL.  I recently added support for this syntax to ndstore in order to support an encoding=raw parameter for use with the prototype python server.  This would be a bit awkward for specifying different chunk sizes for each resolution, but for Python integration you probably only will have one resolution, and the NDstore info format is also limited in that it would only allow specifying a single chunk size, rather than multiple.

PR #4, opened 2016-06-20
improved dvid backend support

stephenplaza:
DVID does not have a specific datatype for multi-scale chunk data.  However, a power of two pyramid is created by convention.  For example, for a data instance 'grayscale', 'grayscale_1', 'grayscale_2', etc represents multiple levels.  I made a small modification to the dvid frontend to check for this convention.

As noted by the author, DVID also does not indicate which tiles are available for its imagetile datatype (XY, YZ, and/or XZ).  However, by default, DVID returns a blank tile if nothing is available.  I changed the default to expect all orientations given that the performance is really slow anyway for XZ and YZ if only XY is assumed.

I added some basic parsing of the DVID dag so that the version history can be observed (segmentation can now be viewed at different points in time).  In the future, it might make sense to allow the user to specify the repo id and then choose a specific version.  For now, I just show all the version nodes for every repository in DVID.

### DIFF
--- a/src/neuroglancer/datasource/dvid/backend.ts
+++ b/src/neuroglancer/datasource/dvid/backend.ts
@@ -42,7 +42,12 @@ class VolumeChunkSource extends GenericVolumeChunkSource {
       // computeChunkBounds.
       let chunkPosition = this.computeChunkBounds(chunk);
       let {chunkDataSize} = chunk;
-      path = `/api/node/${params['nodeKey']}/${params['dataInstanceKey']}/raw/0_1_2/${chunkDataSize[0]}_${chunkDataSize[1]}_${chunkDataSize[2]}/${chunkPosition[0]}_${chunkPosition[1]}_${chunkPosition[2]}/nd`;
+ 
+      if (params['level'] == "0") {
+        path = `/api/node/${params['nodeKey']}/${params['dataInstanceKey']}/raw/0_1_2/${chunkDataSize[0]}_${chunkDataSize[1]}_${chunkDataSize[2]}/${chunkPosition[0]}_${chunkPosition[1]}_${chunkPosition[2]}/nd`;
+      } else {
+        path = `/api/node/${params['nodeKey']}/${params['dataInstanceKey']}_${params['level']}/raw/0_1_2/${chunkDataSize[0]}_${chunkDataSize[1]}_${chunkDataSize[2]}/${chunkPosition[0]}_${chunkPosition[1]}_${chunkPosition[2]}/nd`;
+      }
     }
     handleChunkDownloadPromise(
         chunk, sendHttpRequest(openShardedHttpRequest(params.baseUrls, path), 'arraybuffer'),
--- a/src/neuroglancer/datasource/dvid/backend.ts
+++ b/src/neuroglancer/datasource/dvid/backend.ts
@@ -15,12 +15,12 @@
  */
 
 import {handleChunkDownloadPromise} from 'neuroglancer/chunk_manager/backend';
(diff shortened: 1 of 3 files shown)


stephenplaza:
I signed the CLA.

src/neuroglancer/datasource/dvid/frontend.ts
```diff
@@ -253,30 +284,59 @@ export class RepositoryInfo {
...
+      // just copy references
+      this.errors = obj.errors;
+      this.dataInstances = obj.dataInstances; 
+      return;  
+    } 
     verifyObject(obj);
     this.alias = verifyObjectProperty(obj, 'Alias', verifyString);
     this.description = verifyObjectProperty(obj, 'Description', verifyString);
     let dataInstanceObjs = verifyObjectProperty(obj, 'DataInstances', verifyObject);
     for (let key of Object.keys(dataInstanceObjs)) {
```
jbms: Since Object.keys(dataInstanceObjs) constructs a new array, it might be better to call it just once and use it both for the loop and for the calls to parseDataInstance.

src/neuroglancer/datasource/dvid/frontend.ts
```diff
@@ -90,21 +101,38 @@ export class VolumeDataInstanceInfo extends DataInstanceInfo {
...
-                }))
-                .map(spec => {
-                  return chunkManager.getChunkSource(
-                      VolumeChunkSource, stableStringify(parameters),
-                      () => new VolumeChunkSource(chunkManager, spec, parameters));
-                })];
+  getSources(chunkManager: ChunkManager, parameters: DVIDSourceParameters) {
+    let sources: VolumeChunkSource[][] = [];
+    for (let level = 0; level < this.numLevels; ++level) {
+      let voxelSize = vec3.clone(this.voxelSize);
```
jbms: I think it would still make sense to call VolumeChunkSpecification.getDefaults, since that automatically enables client-side transcoding to compressed segmentation format, and in the future might intelligently pick the chunkDataSize based on the voxel size.

src/neuroglancer/datasource/dvid/backend.ts
```diff
@@ -42,7 +42,12 @@ class VolumeChunkSource extends GenericVolumeChunkSource {
       // computeChunkBounds.
       let chunkPosition = this.computeChunkBounds(chunk);
       let {chunkDataSize} = chunk;
-      path = `/api/node/${params['nodeKey']}/${params['dataInstanceKey']}/raw/0_1_2/${chunkDataSize[0]}_${chunkDataSize[1]}_${chunkDataSize[2]}/${chunkPosition[0]}_${chunkPosition[1]}_${chunkPosition[2]}/nd`;
+ 
+      if (params['level'] == "0") {
```
jbms: It seems this could be simplified by just setting params['dataInstanceKey'] to the level-specific data instance key.  Then there would not need to be any branching here.

jbms:
Thanks for these improvements, Stephen!

I just have a few small comments.

For style consistency, can you please run clang-format on all changed files and ensure that tslint gives no errors?  You can npm install both clang-format and tslint.  (The npm version of clang-format is kept up to date.)  In particular, I'd like to maintain the use of single quotes for string literals === and !== instead of == and !=, except in cases where the behavior of == and != is specifically needed.  Also the interface definitions don't need a trailing semicolon (actually an existing mistake I made in base.ts), which becomes apparent once you run clang-format.

stephenplaza:
Thanks Jeremy for the feedback!  I believe I have addressed all of your comments and ran clang-format followed cslint for the three changed files.  (I originally thought that clang-format was going to remove the semicolon from the interface hence the extra commit.)

jbms:
Great, thanks!

PR #5, opened 2016-06-22
Ndstore chunk size (second try)

funkey:
This commit adds support for setting the chunkSize in an ndstore data backend. For that, the `info` JSON contains an additional (optional) field `chunk_size` in `dataset`. If it exists, neuroglancer uses it to set the chunk size, otherwise, it falls back to the default.


jbms:
I just added a new "python" datasource so that we don't have to keep working within the limitations of the ndstore protocol.  (It is still very similar to the "ndstore" protocol.)  In addition to supporting user-specified chunk sizes, it also can be used with multi-channel arrays, and will probably soon support on-the-fly generated meshes.

The changes in Nyroglancer to use the "python" protocol instead of "ndstore" should be quite minimal.  I also refactored the Python code so that it might be easier to reuse more of it.

funkey:
Great! That is a very nice solution, especially regarding the multi-channel images.

PR #11, opened 2016-12-23
A first pass at render support 

alexbaden:
This pull request adds support for render (https://github.com/saalfeldlab/render) as a datasource. Major features are complete, including stack name auto-completion in the add dialogue, metadata parsing from render, and of course displaying render tiles. The major incomplete feature is handling multi-scale render data (i.e. render data w/ generated mipmaps). 

decode_jpeg_stack has also been modified to support 3-channel (RGB) jpegs. A test was added in decode_jpeg_stack.spec.ts.

### DIFF
--- a/src/neuroglancer/datasource/render/frontend.ts
+++ b/src/neuroglancer/datasource/render/frontend.ts
@@ -1,0 +1,342 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @file
+ * Support for Render (https://github.com/saalfeldlab/render) servers.
+ */
+
+import {ChunkManager} from 'neuroglancer/chunk_manager/frontend';
+import {CompletionResult, registerDataSourceFactory} from 'neuroglancer/datasource/factory';
+import {TileChunkSourceParameters} from 'neuroglancer/datasource/render/base';
+import {DataType, VolumeChunkSpecification, VolumeSourceOptions, VolumeType} from 'neuroglancer/sliceview/base';
+import {defineParameterizedVolumeChunkSource, MultiscaleVolumeChunkSource as GenericMultiscaleVolumeChunkSource, VolumeChunkSource} from 'neuroglancer/sliceview/frontend';
+import {applyCompletionOffset, getPrefixMatchesWithDescriptions} from 'neuroglancer/util/completion';
+import {Vec3, vec3} from 'neuroglancer/util/geom';
+import {openShardedHttpRequest, sendHttpRequest} from 'neuroglancer/util/http_request';
+import {parseArray, parseQueryStringParameters, verifyFloat, verifyInt, verifyObject, verifyObjectProperty, verifyOptionalString, verifyString} from 'neuroglancer/util/json';
+import {CancellablePromise, cancellableThen} from 'neuroglancer/util/promise';
+
+const VALID_ENCODINGS = new Set<string>(['jpg']);
+
+const TileChunkSource = defineParameterizedVolumeChunkSource(TileChunkSourceParameters);
+
+interface OwnerInfo {
+  owner: string;
+  stacks: Map<string, StackInfo>;
+}
+
+interface StackInfo {
+  lowerVoxelBound: Vec3;
+  upperVoxelBound: Vec3;
+  voxelResolution: Vec3; /* in nm */
+  mipMapLevels: number;
+  project: string;
+}
+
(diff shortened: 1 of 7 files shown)


src/neuroglancer/datasource/render/frontend.ts
```diff
@@ -0,0 +1,345 @@
...
+  return voxelResolution;
+}
+
+function parseMipMapLevels(currentMipMapPathBuilderObj: any): number {
+  let levels = 0;
+  try {
+    levels = verifyObjectProperty(currentMipMapPathBuilderObj, 'numberOfLevels', verifyInt);
+  } catch (Error) {
+    // TODO: Something better than console.log for passing messages?
+    console.log('No Mip Map Levels specified. Using default of 0.');
```
alexbaden: An info button next to the Help button (say an "!" in addition to the "?") that displays a popup of log messages might do the trick. Specifically, it might be nice to allow the user to try to reload a given tile/volume/chunk in the event of a load failure. We sometimes encounter those in render and ndstore, e.g. if there's a cache miss that times out. 

As far as the Mip Map Levels error message... I think I've figured out how to properly handle Mip Map levels, so this should go away. I'll write that up in a separate post.

src/neuroglancer/datasource/render/frontend.ts
```diff
@@ -0,0 +1,345 @@
...
+  getSources(volumeSourceOptions: VolumeSourceOptions) {
+    // let sources: Array<any> = [];
+    let sources: VolumeChunkSource[][] = [];
+
+    for (let level = 0; level <= this.stackInfo.mipMapLevels; level++) {
+      let voxelSize = vec3.clone(this.stackInfo.voxelResolution);
+      let chunkDataSize = vec3.fromValues(1, 1, 1);
+      // tiles are NxMx1
+      for (let i = 0; i < 2; ++i) {
+        voxelSize[i] = voxelSize[i] / Math.pow(2, level);
```
alexbaden: Yeah, you're right. I think this should be `*`.

alexbaden:
Thanks for the detailed feedback! I've fixed a bunch of items in the datasource files. I'll try and take a look at the jpeg decoder early next week. 

In terms of handling MipMaps, here's what has been suggested: 
Render is designed to materialize the desired 2D view in the most efficient way possible. In some cases, especially with multi-resolution EM data, MipMaps may be generated for some tiles (high res) and not for others (low res). Render knows about this and will materialize a tile with the correct selection of mipmap/non-mipmap tiles. 

Render also caches materialized tiles. So, it seems most efficient to let render do the bulk of the downsampling. The suggestion then is to pick a range (say base resolution up to 1024 x 1024 or 4096 x 4096) in pixel space and have neuroglancer grab new tiles for each subsequent power of 2 beyond the base resolution, up to the limit (and which point requesting new tiles leads to diminishing returns). Neuroglancer won't care about how many mipmap levels exist, especially since mipmaps aren't the full story.

This should be pretty easy to implement. I'll check it out and see what the performance is like. What do you think?

jbms:
Thanks, I'll take a look.

I'm not sure what you mean by "base resolution up to 1024 x 1024 or 4096 x 4096".  It sounds like you are saying that render will generate tiles at arbitrary power-of-2 downsamplings, so I think you might as well just pick a sufficient number of downsampling levels so that the total image dimensions at the lowest resolution are less than some constant, e.g. 512.

jbms:
The changes to datasource/render look good.  One very minor point: "chore" in the commit message would normally refer to changes in the build system or formatting that don't affect the actual program.  In this case you are adding support for more than single channel jpegs, so that could be labeled "feat" for feature.

src/neuroglancer/sliceview/decode_jpeg_stack.ts
```diff
@@ -20,14 +20,41 @@ import {Vec3, vec3} from 'neuroglancer/util/geom';
 export function decodeJpegStack(data: Uint8Array, chunkDataSize: Vec3) {
```
alexbaden: I didn't change this, yet, since it seems like it will involve a lot of file changes across the datasource folder and may be better in a separate pull request. Do you agree?

jbms: Actually, it looks like it is called directly only once, in neuroglancer/sliceview/backend_chunk_decoders/jpeg.ts --- and from there you can get the number of channels as chunk.source!.spec.numChannels

alexbaden: Good call. I modified both files and added an error message for differing number of components and an unexpected number of components.

alexbaden:
I think the merge conflict is a result of same change in the master branch since I last merged. I'm not sure if it's better for me to merge master into my render branch, or to wait until this pull request is merged. 

The only outstanding item (I think) is better handling of multiple resolutions.

jbms:
All the changes look great --- thanks.  The conflict is because a change to the type definitions for gl-matrix allowed vec3, etc. to also serve as types, so I eliminated the Vec3, etc. types and replaced all uses with the lowercase names.  This also affects some of your other code, but I can fix this myself when I merge it.

If you want to work on the multiple scaling levels more, I can wait to merge; or I'm happy to merge this now.

alexbaden:
Sounds good. We're still working through the best way to do multiple scaling levels. So, in the meantime, I changed the parseMipMapLevels functions to always set the number of scaling levels to 0 (I'm not entirely convinced it will work in all cases as previously implemented). Right now neuroglancer will work on any render project by pulling resolution 0 tiles. If that is okay with you, then I'm in favor of merging!

PR #15, opened 2017-01-24
Updates to the Render datasource

alexbaden:
The main feature added in this PR is support for multiple scales in the Render datasource. When loading data from a Render server, neuroglancer will now compute a resolution hierarchy (levels 0 to n) where the nth level corresponds roughly to the user specified tilesize (default 1024) in the largest dimension. If Render is properly configured and mipmaps have been generated, this configuration should result in the best performance. 

In addition, this PR adds a fix to ignore render stacks not in the COMPLETED state (render will fail to return images for these stacks) and a json utility function to verify an optional Int (used to parse the user specified tilesize parameter mentioned above).


alexbaden:
Looks like there's a merge conflict relating to the change in CancellationToken mechanism. Should I try to resolve on my end?

jbms:
Yeah, perhaps you can try to resolve it yourself, but feel free to ask if you have any questions about it.

alexbaden:
Turned out it was pretty easy.

PR #17, opened 2017-02-12
Adding VR capabilities to neuroglancer

Kaju-Bubanja:
This PR adds experimental WebVR capabilities to neuroglancer, accessible as an additional layout (press space bar multiple times).

Shortcomings
    -When comparing the WebVR examples with the neuroglancer VR experience, I find that there is still a small distortion in the neuroglancer VR that might create some discomfort. I do not know yet how to improve it. The key change that implements the VR views is:
    newVertex = {LeftEyeProjectionMatrix from WebVR} * {LeftEyeModelViewMatrix from WebVR} * {ModelViewMatrix from Neuroglancer} * vertex.

    -The dimensions of the canvas that we provide to the Oculus are the dimensions of the browser canvas on screen (currently about 1920,1080). However we think that we should provide a canvas with the dimensions coming from the Oculus canvas (about 1400, 1000 from what we read with leftEye.renderWidth or 1080,960 from the online specifications) but it does not look good. It is also unclear why the canvas sizes differ between our leftEye.renderWidth and the specifications.
   
    -The current VR layout does not display tissue slices. There is experimental code that does it in additional layouts (currently commented out) but the panning/zooming behavior has errors.
   
    -The on-screen mirroring is not working properly (flickering and black after a while)
   
    -It was tested only with Oculus DK2.
   
Suggestions for continuation
    -Object selection by head tracking: We have not had time yet to implement a cursor (center of the VR view) controlled by the head tracking. The machinery is already there with the mouse cursor that can pick objects, it would probably be similar in the VR context (get cursor coordinates, apply the inverse of the webVR ProjectionMatrix, ...).
       
    - Implement navigation with a gamepad, the gamepad webvr API mentioned by Jeremy looks like a good library choice (https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API)

There are only so many files in the PR because of the formatting of gulp. There are 5 main files that implement the changes:
- display_context.ts
- perspective_view/panel.ts
- viewer.ts
- viewer_layouts.ts
- webvr-utils.ts
`npm gulp check-format` throws no errors.
`npm run gulp tslint` throws a few errors, but all in files I did not edit.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "typescript.tsdk": "./node_modules/typescript/lib"
+}
--- a/config/karma.conf.js
+++ b/config/karma.conf.js
@@ -59,7 +59,7 @@ module.exports = function(config) {
     webpack: webpackConfig,
     webpackServer: {noInfo: true},
     browserStack: {
-      // This empty object is required to work around a bug in karma-browserstack-launcher.
+        // This empty object is required to work around a bug in karma-browserstack-launcher.
     },
     browsers: [
       'Chrome',
--- a/config/ts_loader_patch.js
+++ b/config/ts_loader_patch.js
@@ -49,24 +49,23 @@
 const tsLoaderConfigModule = require('ts-loader/dist/config.js');
 const tsLoaderUtilsModule = require('ts-loader/dist/utils.js');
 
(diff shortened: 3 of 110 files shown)


jbms:
Thanks!  I'll try to get a hold of a VR device and try this out.

If you are able to undo the formatting changes to untouched files, that would simplify things.  (My instructions about running the reformatting command should probably be updated.)

Kaju-Bubanja:
Nice, yes I will do that.

PR #20, opened 2017-02-27
Update README.md

kmader:
adding docker section to the readme


jbms:
Can you clarify what the advantage is of using a docker image rather than just relying on, say, a pypi package?

kmader:
The advantage of Docker is that it is entirely self-contained image and won't conflict with any other running services or setups (particularly challenging with Windows machines). It also makes starting many machines running the same software much easier (using autoscaling and tools like Kubernetes this makes having a production ready site much easier). I am also working on the docker image to have it build the nodejs and nvm tools as well so everything can be assembled in one package.

jbms:
Okay, if I understand correctly, your https://github.com/4Quant/neuroglancer-docker repository serves as a useful example that others can modify in order to produce docker images that combine the Neuroglancer Python code with necessary code for starting some service that makes use of it, and possibly some data for it to use.

Do you imagine that others would want to directly use the existing docker images that you published?

kmader:
Probably most people would modify the images for their own datasets / configurations but it is possible to use the images, but we have a maintained published version of the image (https://hub.docker.com/r/4quant/neuroglancer/)

jbms:
I added a link to your neuroglancer-docker project to a new "Related Projects" section of the README.  Thanks!

PR #21, opened 2017-02-28
feat:add swc parsing

westona:
New decoder for parsing swc files into vertices and indices needed for skeleton rendering
Includes simple tests. tslint and formatting have been run.

I'll create a separate PR for the DVID integration example once this is merged.


jbms:
Thanks for putting this together.  Aside from the few line comments, I'd suggest moving from the sliceview/ directory to the skeleton/ directory.  Additionally, it is a nit pick, but camelCase for all variables would be preferable for consistency with the rest of the code.

westona:
@googlebot  I signed it!

westona:
@jbms nit picks are welcome! Gotta keep the codebase clean.

I wasn't sure what you meant by "Aside from the few line comments". Should I remove the comments? Or add more?

jbms:
Oh, I actually was just referring to my own review comments associated with particular lines of code.  I just saw that I left them pending in github's interface, though, and forgot to submit them.

src/neuroglancer/sliceview/decode_swc_skeleton.ts
```diff
@@ -0,0 +1,77 @@
+import {each} from 'lodash';
+import {decodeSkeletonVertexPositionsAndIndices, ParameterizedSkeletonSource, SkeletonChunk} from 'neuroglancer/skeleton/backend';
+import {convertEndian32, Endianness} from 'neuroglancer/util/endian';
+
+export function decodeSwcSkeletonChunk(
+    chunk: SkeletonChunk, swc_str: string, endianness: Endianness) {
```
westona: This means that additionally I don't need to call convertEndian32 below?

jbms: Yeah, that's right.  Endian conversion is only needed for converting binary formats into native endianness.

westona: Ahh, that makes sense. Thanks!

PR #22, opened 2017-03-21
Sliceview refactor & PointChunk 

alexbaden:
This pull request includes a refactoring of SliceView to separate image/segmentation related code from general code required for loading chunks in cross-sectional views. As an application, code for downloading, processing, and rendering vertex data in 2D cross-sections is also included.


src/neuroglancer/point/backend.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.systemMemoryBytes = this.gpuMemoryBytes = this.vertexPositions!.byteLength;
+    super.downloadSucceeded();
+  }
+
+  freeSystemMemory() {
+    this.vertexPositions = null;
+  }
+};
+
+// @registerSharedObject(POINT_SOURCE_RPC_ID)
```
jbms: Remove commented out line.

src/neuroglancer/point/backend.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  downloadSucceeded() {
+    this.systemMemoryBytes = this.gpuMemoryBytes = this.vertexPositions!.byteLength;
+    super.downloadSucceeded();
+  }
+
+  freeSystemMemory() {
+    this.vertexPositions = null;
+  }
+};
```
jbms: Unlike C++, semicolons after class definitions are unnecessary in javascript/typescript.  I've made this mistake elsewhere in neuroglancer, but I'm trying to clean it up as I modify files.

src/neuroglancer/point/backend.ts
```diff
@@ -0,0 +1,114 @@
...
+      chunk = this.getNewChunk_(PointChunk);
+      chunk.initializeVolumeChunk(key, chunkGridPosition);
+      this.addChunk(chunk);
+    }
+    return chunk;
+  }
+};
+
+@registerSharedObject(POINT_RENDERLAYER_RPC_ID)
+export class RenderLayer extends SliceViewRenderLayer implements RenderLayerInterface {
```
jbms: This code is identical to the code for RenderLayer defined in sliceview/volume/backend.ts.  Perhaps it can be refactored?

src/neuroglancer/point/base.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Specifies a chunk layout and voxel size.
+ */
+export class PointChunkSpecification extends SliceViewChunkSpecification {
+  chunkBytes: number;
+
+  constructor(options: PointChunkSpecificationOptions) {
+    super(options);
+
+    let chunkBytes = 10000;  // TODO!  remove??
```
jbms: Remove unused variable declaration.

src/neuroglancer/point/base.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export interface PointChunkSpecificationOptions extends SliceViewChunkSpecificationBaseOptions {
+  chunkDataSize: vec3;
+}
+
+/**
+ * Specifies a chunk layout and voxel size.
+ */
+export class PointChunkSpecification extends SliceViewChunkSpecification {
+  chunkBytes: number;
```
jbms: It seems like this could just be part of the backend chunk class, rather than here in base, since it isn't needed in the frontend.  Additionally, I would guess that the number of bytes may vary between chunks (which would be another reason to make it a variable of the Chunk class, rather than the SliceViewChunkSpecification), rather than being fixed for all chunks for a given source.

src/neuroglancer/point/point_renderlayer.ts
```diff
@@ -0,0 +1,86 @@
...
+    builder.addFragmentCode(glsl_COLORMAPS);
+    builder.setFragmentMainFunction(FRAGMENT_MAIN_START + '\n' + DEFAULT_FRAGMENT_MAIN);
+  }
+
+  beginSlice(sliceView: SliceView) {
+    let shader = super.beginSlice(sliceView);
+    let {gl} = this;
+    gl.uniform1f(shader.uniform('uOpacity'), this.opacity.value);
+    gl.uniform3fv(
+        shader.uniform('uColor'), vec3.fromValues(1.0, 0.0, 0.5));  // TODO accept from user
```
jbms: Supporting custom shaders, as for the image renderlayer (and as of recently, for skeletons, based on an arbitrary set of custom vertex attributes) would be great, but for the first version this is fine.

src/neuroglancer/point_user_layer.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+
+import {ChunkManager} from 'neuroglancer/chunk_manager/frontend';
+import {CoordinateTransform} from 'neuroglancer/coordinate_transform';
+import {getPoint, GetPointOptions} from 'neuroglancer/datasource/factory';
+import {UserLayer, UserLayerDropdown} from 'neuroglancer/layer';
+import {LayerListSpecification, registerLayerType, registerVolumeLayerType} from 'neuroglancer/layer_specification';
+import {getVolumeWithStatusMessage} from 'neuroglancer/layer_specification';
+import {Overlay} from 'neuroglancer/overlay';
+// import {VolumeType} from 'neuroglancer/sliceview/base';
```
jbms: remove commented out code

src/neuroglancer/point_user_layer.ts
```diff
@@ -0,0 +1,129 @@
...
+    x['opacity'] = this.opacity.toJSON();
+    return x;
+  }
+  makeDropdown(element: HTMLDivElement) {
+    return new PointDropDown(element, this);
+  }
+}
+
+/*
+function makeShaderCodeWidget(layer: ImageUserLayer) {
```
jbms: remove commented out code

src/neuroglancer/point_user_layer.ts
```diff
@@ -0,0 +1,129 @@
...
+          let renderLayer = this.renderLayer =
+              new PointRenderLayer(points, {opacity: this.opacity, sourceOptions: {}});
+          this.addRenderLayer(renderLayer);
+        }
+      });
+    }
+  }
+  toJSON() {
+    let x: any = {'type': 'point'};
+    x['point'] = this.pointsPath;
```
jbms: Perhaps make this 'source' rather than 'point' for consistency with other layers.

src/neuroglancer/sliceview/backend.ts
```diff
@@ -135,59 +138,40 @@ export class VolumeChunk extends Chunk {
...
 
-export abstract class VolumeChunkSource extends ChunkSource implements VolumeChunkSourceInterface {
-  spec: VolumeChunkSpecification;
+export abstract class SliceViewChunkSource extends ChunkSource implements
+    SliceViewChunkSourceInterface {
+  spec: SliceViewChunkSpecification;
   constructor(rpc: RPC, options: any) {
     super(rpc, options);
-    this.spec = VolumeChunkSpecification.fromObject(options['spec']);
+    // this.spec = SliceViewChunkSpecification.fromObject(options['spec']);
```
jbms: remove commented out code

src/neuroglancer/sliceview/backend.ts
```diff
@@ -237,34 +221,18 @@ export abstract class VolumeChunkSource extends ChunkSource implements VolumeChu
...
 }
 
-@registerSharedObject(SLICEVIEW_RENDERLAYER_RPC_ID)
-export class RenderLayer extends SharedObjectCounterpart implements RenderLayerInterface {
-  rpcId: number;
-  sources: VolumeChunkSource[][];
+export abstract class RenderLayer extends SharedObjectCounterpart implements RenderLayerInterface {
+  sources: SliceViewChunkSource[][];
   layerChanged = new NullarySignal();
-
```
jbms: Per my previous comment about refactoring this, maybe this stuff can actually stay part of sliceview, rather than be duplicated.

src/neuroglancer/sliceview/renderlayer.ts
```diff
@@ -13,46 +13,23 @@
...
 
-/**
- * If the absolute value of the dot product of a cube edge direction and the viewport plane normal
- * is less than this value, intersections along that cube edge will be exluded.  This needs to be
- * non-zero in order to avoid vertex placement artifacts.
- */
-const ORTHOGONAL_EPSILON = 1e-3;
-
-class SliceViewShaderBuffers extends RefCounted {
+export class SliceViewShaderBuffers extends RefCounted {
```
jbms: This stuff is specific to sliceview/volume, so it would be better to move it there.

src/neuroglancer/sliceview/renderlayer.ts
```diff
@@ -141,258 +118,27 @@ class SliceViewShaderBuffers extends RefCounted {
...
     super();
+
     this.shaderError = shaderError;
     shaderError.value = undefined;
-    this.chunkManager = multiscaleSource.chunkManager;
+    this.chunkManager = chunkManager;
     let gl = this.gl;
-    this.vertexComputationManager = VolumeSliceVertexComputationManager.get(gl);
 
-    let sources = this.sources = multiscaleSource.getSources(volumeSourceOptions);
```
jbms: Maybe the common code for dealing with the sources can actually remain here, rather than be duplicated?

src/neuroglancer/sliceview/renderlayer.ts
```diff
@@ -476,91 +206,11 @@ export class RenderLayer extends GenericRenderLayer {
...
-          vertexComputationManager.drawChunk(gl, shader, chunkPosition);
-        }
-      }
-    }
-    chunkFormat.endDrawing(gl, shader);
-    this.endSlice(shader);
-  }
-}
+  abstract draw(sliceView: SliceView): void 
+}
```
jbms: Fix trailing whitespace, add newline at end of file.

src/neuroglancer/sliceview/volume/backend.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.data = null;
+  }
+
+  serialize(msg: any, transfers: any[]) {
+    super.serialize(msg, transfers);
+    let data = msg['data'] = this.data!;
+
+    transfers.push(data.buffer);
+    this.data = null;
+    // console.log(`Serializing chunk ${this.source.rpcId}:${this.key} with
```
jbms: Delete commented out code

src/neuroglancer/sliceview/volume/renderlayer.ts
```diff
@@ -0,0 +1,449 @@
...
+        let denom = vec3.dot(vDir, uPlaneNormal);
+        if (Math.abs(denom) > ORTHOGONAL_EPSILON) {
+          let lambda = (uPlaneDistance - vec3.dot(vStart, uPlaneNormal)) / denom;
+          if ((lambda >= -LAMBDA_EPSILON) && (lambda <= 1.0 + LAMBDA_EPSILON)) {
+            lambda = Math.max(0, Math.min(1, lambda));
+            vec3.scaleAndAdd(position, vStart, vDir, lambda);
+            vec3.transformMat4(gl_Position, position, uProjectionMatrix);
+            vec3.scale(vChunkPosition, uVertexBasePosition(vidx[0]), 1.0 - lambda);
+            vec3.scaleAndAdd(vChunkPosition, vChunkPosition, uVertexBasePosition(vidx[1]), lambda);
+            console.log(
```
jbms: Perhaps this is just the work of clang-format, but better formatting would be nice.

(This function also hasn't worked for a little while, since it hasn't been updated to reflect changes in other code, but it is still somewhat useful to keep around so that it could be fixed later if necessary for debugging.)

alexbaden: Yeah, it looks like the formatter did that. I believe it has been fixed and is more-or-less in compliance.

alexbaden:
I've cleaned up a lot of the code and refactored the sources processing in both the frontend and backend renderlayer code. I agree that "vectorgraphics" is a more descriptive name for this new functionality. Do you think it should also live under sliceview, with volume? (E.g. sliceview/vectorgraphics)

I need to clean up the Render point match datasource and add those changes, too. That should make the point service useful for Render point matches and provide a nice template for others to follow for reading their own point matches. It is also probably not too hard to add this functionality to the precomputed datasource.

jbms:
Thanks --- yes I think it does make sense to put it in e.g. sliceview/vector_graphics  (sliceview doesn't use an underscore, but for new files I'm trying to do that).

src/neuroglancer/sliceview/vector_graphics/backend.ts
```diff
@@ -35,21 +36,36 @@ export class PointChunk extends SliceViewChunk {
 
     let source = this.source;
 
-    this.systemMemoryBytes = source!.chunkBytes;
-    this.gpuMemoryBytes = source!.chunkBytes;
+    let chunkBytes: number = 0;
+    if (this.vertexPositions) {
+      chunkBytes = chunkBytes + this.vertexPositions!.buffer.byteLength;
```
jbms: I think you want just vertexPositions!.byteLength, not buffer.byteLength, in case it is just a view to a part of a larger buffer.  Also consider chunkBytes += .  Same applies to vertexNormals below.

jbms: Actually, this code for calculating the memory usage should be in downloadSucceeded, not here, since the data won't be available at the time this called.  The code in the existing initializeVolumeChunk method in VolumeChunk that set systemMemoryBytes and gpuMemoryBytes according to source!.spec.chunkBytes should in fact be removed, as it was just left there by mistake.  Originally the chunk manager code expected systemMemoryBytes to be set before a chunk was downloaded in order to ensure there would be sufficient free memory before starting the download, but that is no longer the case, since for variable size chunks it isn't possible to estimate the memory requirement ahead of time.

src/neuroglancer/util/http_request.ts
```diff
@@ -97,6 +99,36 @@ export function sendHttpRequest(xhr: XMLHttpRequest, responseType: XMLHttpReques
...
+    xhr.onloadend = function(this: XMLHttpRequest) {
+      let status = this.status;
+      token.remove(abort);
+      if (status >= 200 && status < 300) {
+        resolve(this.response);
+      } else {
+        reject(HttpError.fromXhr(xhr));
+      }
+    };
+    xhr.send(`${JSON.stringify(payload)}`);
```
jbms: `${  unneccessary

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
```diff
@@ -0,0 +1,236 @@
...
+          gl.ANGLE_instanced_arrays.drawArraysInstancedANGLE(gl.TRIANGLE_STRIP, 0, 4, numInstances);
+
+          gl.ANGLE_instanced_arrays.vertexAttribDivisorANGLE(aVertexFirst, 0);
+          gl.ANGLE_instanced_arrays.vertexAttribDivisorANGLE(aVertexSecond, 0);
+        }
+      }
+    }
+    this.endSlice(shader);
+  }
+}
```
jbms: Add linebreak at end of file

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
```diff
@@ -0,0 +1,236 @@
...
+      if (!visibleChunks) {
+        continue;
+      }
+
+      for (let key of visibleChunks) {
+        let chunk = chunks.get(key);
+        if (chunk && chunk.state === ChunkState.GPU_MEMORY) {
+          let numInstances = chunk.numPoints / 2;  // Two points == One vector
+
+          this.vertexIndexBuffer.bindToVertexAttrib(
```
jbms: You can bind vertexIndexBuffer and normalDirectionBuffer outside the loop

jbms: Actually it can be done at the same time as beginSlice

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
```diff
@@ -0,0 +1,236 @@
...
+import {makeTrackableFragmentMain, makeWatchableShaderError, TrackableFragmentMain} from 'neuroglancer/webgl/dynamic_shader';
+import {ShaderBuilder, ShaderProgram} from 'neuroglancer/webgl/shader';
+
+export const FRAGMENT_MAIN_START =
+    '//NEUROGLANCER_VECTORGRAPHICS_LINE_RENDERLAYER_FRAGMENT_MAIN_START';
+
+const DEFAULT_FRAGMENT_MAIN = `void main() {  
+vec3 color = vec3(0,1,0);
+  
+  float distance = length(vNormal);
```
jbms: While this does provide a very simple and convenient way for the user to specify the color and the amount of feathering, I am concerned that this exposes a lot of implementation details into the user-customizable fragment shader code (which becomes part of the JSON state), and that would make it very difficult to change the implementation in the future without breaking backwards compatibility.

I see two possible ways to resolve this:
1. Abstract all of the logic here into a drawLine function that takes a feather and color parameter.  Then the default fragment shader just becomes:

drawLine(/*color=*/vec3(0,1,0), /*feather=*/0.5);

2. Alternatively, instead of using a text field, you could just expose separate UI elements for specifying the amount of feathering and color.

If in the future there might be more ways to customize the line rendering, e.g. based on arbitrary attributes associated with each line or line vertex, then the first option is probably better.

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
```diff
@@ -0,0 +1,236 @@
...
+  float feather = 0.5;
+  
+  if (distance >= 1.0 - feather) {
+    emitRGBA(vec4(color, (distance - 1.0) / -feather )); 
+  }
+  else if (distance < 1.0 - feather) {
+  	emitRGB(color);
+  }
+  else {
+    emitTransparent();
```
jbms: Is this better than discard; for some reason?

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
```diff
@@ -0,0 +1,236 @@
...
+export function getTrackableFragmentMain(value = DEFAULT_FRAGMENT_MAIN) {
+  return makeTrackableFragmentMain(value);
+}
+
+const tempMat4 = mat4.create();
+
+export class VectorGraphicsLineRenderLayer extends GenericVectorGraphicsRenderLayer {
+  fragmentMain: TrackableFragmentMain;
+  opacity: TrackableAlphaValue;
+  primitiveSize: TrackableFiniteFloat;
```
jbms: I wonder if this should just be lineWidth instead of primitiveSize.  What else do you think it might be used for sizing, and in that case would the same value as used for lineWidth generally be appropriate?

src/neuroglancer/datasource/render/frontend.ts
```diff
@@ -21,17 +21,20 @@
 
 import {ChunkManager} from 'neuroglancer/chunk_manager/frontend';
 import {CompletionResult, registerDataSourceFactory} from 'neuroglancer/datasource/factory';
-import {TileChunkSourceParameters} from 'neuroglancer/datasource/render/base';
-import {DataType, VolumeChunkSpecification, VolumeSourceOptions, VolumeType} from 'neuroglancer/sliceview/base';
-import {defineParameterizedVolumeChunkSource, MultiscaleVolumeChunkSource as GenericMultiscaleVolumeChunkSource, VolumeChunkSource} from 'neuroglancer/sliceview/frontend';
-import {applyCompletionOffset, getPrefixMatchesWithDescriptions, getPrefixMatches} from 'neuroglancer/util/completion';
+import {PointMatchChunkSourceParameters, TileChunkSourceParameters} from 'neuroglancer/datasource/render/base';
+import {VectorGraphicsChunkSpecification, VectorGraphicsSourceOptions} from 'neuroglancer/sliceview/vector_graphics/base';
+import {defineParameterizedVectorGraphicsSource, MultiscaleVectorGraphicsChunkSource as GenericMultiscalePointChunkSource, VectorGraphicsChunkSource} from 'neuroglancer/sliceview/vector_graphics/frontend';
```
jbms: Change old PointChunk naming

src/neuroglancer/datasource/render/frontend.ts
```diff
@@ -187,7 +190,8 @@ export class MultiscaleVolumeChunkSource implements GenericMultiscaleVolumeChunk
       stack: string|undefined, public project: string, public parameters: {[index: string]: any}) {
     let projectInfo = ownerInfo.projects.get(project);
     if (projectInfo === undefined) {
-      throw new Error(`Specificed project ${JSON.stringify(project)} does not exist for specified owner ${JSON.stringify(ownerInfo.owner)}`);
+      throw new Error(
+          `Specificed project ${JSON.stringify(project)} does not exist for specified owner ${JSON.stringify(ownerInfo.owner)}`);
```
jbms: Typo: should be "Specified"

src/neuroglancer/datasource/render/frontend.ts
```diff
@@ -390,8 +392,133 @@ export function volumeCompleter(
...
+      'project': this.stackInfo.project,
+      'stack': this.stack,
+      'encoding': 'points',
+      'matchCollection': this.matchCollection,
+      'zoffset': this.zoffset
+    });
+
+    return [[source]];
+  }
+};
```
jbms: semicolon not needed (I know other neuroglancer code also has redundant semicolons)

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
alexbaden: I'm working on a point render layer as well. I think that the primitiveSize should scale pretty well between points and lines, but haven't had a chance to try it yet. I picked the range values sort of arbitrarily. It is possible with use we will figure out a better range. Or maybe I should switch from a range slider to a text input.

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
alexbaden: I like the idea of exposing separate UI elements. While I suspect both points and vectors (lines with arrows) will make use of color and feather, there might be other properties we wish to introduce. Sharing a fragment shader also becomes a problem across multiple render layers, anyway. I think three boxes with floating point values for RGB color values and a feather box would work. As an aside, perhaps "antialiasing" might be a better descriptor than feather for the feather property?

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
alexbaden: I don't think so. Is it worse? I don't see any reason not to change it, regardless.

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
alexbaden: This is very strange. I added a linebreak, but github doesn't seem to have taken notice... I'll see what I can do about that after the other changes are added to line_renderlayer.

src/neuroglancer/datasource/render/frontend.ts
```diff
@@ -390,8 +392,133 @@ export function volumeCompleter(
...
+
+  dims: vec3;
+
+  constructor(
+      public chunkManager: ChunkManager, public baseUrls: string[], public ownerInfo: OwnerInfo,
+      stack: string|undefined, public project: string, public parameters: {[index: string]: any}) {
+    let projectInfo = ownerInfo.projects.get(project);
+    if (projectInfo === undefined) {
+      throw new Error(
+          `Specificed project ${JSON.stringify(project)} does not exist for specified owner ${JSON.stringify(ownerInfo.owner)}`);
```
jbms: Typo: should be "Specified"

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
jbms: I think in general points need to be drawn somewhat wider than lines to be visible.  As far as text vs range, hard to say.

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
jbms: Actually, can't this case just be eliminated, by ensuring the triangles aren't bigger than needed?

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
jbms: Yeah, just calling it antialiasing sounds better.  Is it generally necessary to tweak that, or does a single value give reasonable results?

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
alexbaden: You're right -- it isn't actually required because of the way the other two cases are setup. Originally the antialiasing code worked a little bit differently (or, perhaps, didn't work but was different...).

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
alexbaden: So far a single value of 0.5 has given good results. I think for now this does not have to be exposed to the user.

src/neuroglancer/sliceview/vector_graphics/vector_graphics_line_renderlayer.ts
alexbaden: Seems like going back to lineWidth gives more flexibility in the future. Done.

src/neuroglancer/vector_graphics_user_layer.ts
```diff
@@ -89,37 +81,27 @@ export class VectorGraphicsUserLayer extends UserLayer {
     let x: any = {'type': 'vectorgraphics'};
     x['source'] = this.vectorGraphicsPath;
     x['opacity'] = this.opacity.toJSON();
-    x['size'] = this.primitiveSize.toJSON();
-    x['shader'] = this.fragmentMain.toJSON();
+    x['size'] = this.lineWidth.toJSON();
```
jbms: Should the state property name be changed to something like "lineWidth"?

src/neuroglancer/vector_graphics_user_layer.ts
```diff
@@ -89,37 +81,27 @@ export class VectorGraphicsUserLayer extends UserLayer {
...
   constructor(public element: HTMLDivElement, public layer: VectorGraphicsUserLayer) {
     super();
     element.classList.add('image-dropdown');
-    let {opacityWidget, primitiveSizeWidget} = this;
+    let {opacityWidget, lineWidthWidget} = this;
     let topRow = document.createElement('div');
     topRow.className = 'image-dropdown-top-row';
     opacityWidget.promptElement.textContent = 'Opacity';
-    primitiveSizeWidget.promptElement.textContent = 'Primitive Size';
+    lineWidthWidget.promptElement.textContent = 'Primitive Size';
```
jbms: Description changed to "Line width" ?

jbms:
Hi Alex,

It looks like this is extremely close to being ready!

alexbaden:
Thanks! I probably need to stop making changes... here are hopefully the last few:

I added a trackable color input to the layer dropdown to allow the user to set RGB color values for the vector graphics layer. This is actually an important feature for us, since we often false color light microscopy data and don't want the vector graphics layers to blend in. There's a generic TrackableVec3 class and a generic widget, as well as a specific widget for RGB color values. I decided to display colors in the integer range [0,255] using the widget but convert to floating point values in the range [0,1] in the TrackableVec3 object because of the floating point rounding errors when converting certain strings -- while we are probably losing a little bit of precision, I think the user experience is a bit cleaner.

I also changed the vector graphics layer type from `vectorgraphics` to an enum. There's some code to convert the enum member to a string, so the `type` is still human readable (e.g. `VECTOR_GRAPHICS_LAYER_TYPE.LINE` becomes `line` in the JSON description for the layer. I'm hoping to add a point render layer fairly soon, and it seems like `vector_graphics_user_layer` could encapsulate all the types of vector graphics layers, similarly to how `segmentation_user_layer` encapsulates meshes, skeletons, volumetric segmentation, etc. I suppose we could have a separate user layer for each vector graphics type, but some of them are going to be quite similar (e.g. lines and vectors). It also should not be too difficult to modify the dropdown widget to hide certain fields based on type -- e.g. `line` and `vector` get `lineWidth` while `point` gets `radius`. What do you think?

alexbaden:
It occurs to me that I might need to adjust the `sliceview/vector_graphics` code to properly set the layer type when loading data (e.g. with the `MultiScaleVectorGraphicsChunkSource`). I'll try and look at that shortly.

PR #32, opened 2017-05-02
Dvid backend updates

westona:
These are some updates to the dvid datasource that have piled up. With our latest data release at emdata.janelia.org, it's a good time to get these changes back to the main repo.

### DIFF
--- a/src/neuroglancer/datasource/dvid/backend.ts
+++ b/src/neuroglancer/datasource/dvid/backend.ts
@@ -15,7 +15,7 @@
  */
 
 import {registerChunkSource} from 'neuroglancer/chunk_manager/backend';
-import {TileChunkSourceParameters, TileEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/dvid/base';
+import {TileChunkSourceParameters, TileEncoding, VolumeChunkEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/dvid/base';
 import {ParameterizedVolumeChunkSource, VolumeChunk} from 'neuroglancer/sliceview/backend';
 import {ChunkDecoder} from 'neuroglancer/sliceview/backend_chunk_decoders';
 import {decodeCompressedSegmentationChunk} from 'neuroglancer/sliceview/backend_chunk_decoders/compressed_segmentation';
@@ -33,15 +33,10 @@ const TILE_CHUNK_DECODERS = new Map<TileEncoding, ChunkDecoder>([
 
 @registerChunkSource(VolumeChunkSourceParameters)
 class VolumeChunkSource extends ParameterizedVolumeChunkSource<VolumeChunkSourceParameters> {
-  compressedSegmentationBlockSize: vec3|undefined;
-
   constructor(rpc: RPC, options: any) {
     super(rpc, options);
 
     this.parameters = options['parameters'];
-    if (this.parameters.volumeType === VolumeType.SEGMENTATION) {
-      this.compressedSegmentationBlockSize = vec3.fromValues(8, 8, 8);
-    }
   }
 
   download(chunk: VolumeChunk, cancellationToken: CancellationToken) {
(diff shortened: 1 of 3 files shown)


westona:
I signed it--hopefully takes this time!

src/neuroglancer/datasource/dvid/backend.ts
```diff
@@ -18,17 +18,32 @@ import {registerChunkSource} from 'neuroglancer/chunk_manager/backend';
...
 import {openShardedHttpRequest, sendHttpRequest} from 'neuroglancer/util/http_request';
+import {RPC} from 'neuroglancer/worker_rpc';
 
 const TILE_CHUNK_DECODERS = new Map<TileEncoding, ChunkDecoder>([
   [TileEncoding.JPEG, decodeJpegChunk],
 ]);
 
 @registerChunkSource(VolumeChunkSourceParameters)
 class VolumeChunkSource extends ParameterizedVolumeChunkSource<VolumeChunkSourceParameters> {
+  compressedSegmentationBlockSize: vec3|undefined;
```
jbms: Instead of adding the volumeType parameter and this extra property, you can reuse the existing compressedSegmentationBlockSize property of VolumeChunkSpecification, available as this.spec, which will let you simplify your code.

src/neuroglancer/datasource/dvid/base.ts
```diff
@@ -21,6 +22,7 @@ export class DVIDSourceParameters {
 }
 
 export class VolumeChunkSourceParameters extends DVIDSourceParameters {
+  volumeType: VolumeType;
```
jbms: I think it would be better to use an explicit encoding enum, e.g.:

export enum VolumeChunkEncoding { JPEG, COMPRESSED_SEGMENTATION }

and then add encoding: VolumeChunkEncoding as a parameter.  That makes it easier to support other encodings in the future.

src/neuroglancer/datasource/dvid/frontend.ts
```diff
@@ -112,6 +110,7 @@ export class VolumeDataInstanceInfo extends DataInstanceInfo {
         'baseUrls': parameters.baseUrls,
         'nodeKey': parameters.nodeKey,
         'dataInstanceKey': dataInstanceKey,
+        'volumeType': this.volumeType,
```
jbms: Determine the encoding here, e.g. based on the volumeType, and specify the encoding.

src/neuroglancer/datasource/dvid/base.ts
jbms: Actually, for now you could even avoid an explicit encoding parameter, and just determine the encoding based on the value of spec.compressedSegmentationBlockSize

src/neuroglancer/datasource/dvid/backend.ts
```diff
@@ -37,13 +52,34 @@ class VolumeChunkSource extends ParameterizedVolumeChunkSource<VolumeChunkSource
...
+          `${chunkPosition[0]}_${chunkPosition[1]}_${chunkPosition[2]}/jpeg`;
+    } else {
+      // volumeType is SEGMENTATION
+      return `/api/node/${params['nodeKey']}/${params['dataInstanceKey']}/raw/0_1_2/` +
+          `${chunkDataSize[0]}_${chunkDataSize[1]}_${chunkDataSize[2]}/` +
+          `${chunkPosition[0]}_${chunkPosition[1]}_${chunkPosition[2]}?compression=googlegzip`;
+    }
+  }
+  getDecoder(params: any) {
+    if (params.volumeType === VolumeType.IMAGE) {
```
jbms: You can determine this based on this.spec.compressedSegmentationBlockSize instead (or an explicit encoding parameter if you want to go that route).  No need to pass params, since they can be accessed as this.parameters.

src/neuroglancer/datasource/dvid/backend.ts
```diff
@@ -37,13 +52,34 @@ class VolumeChunkSource extends ParameterizedVolumeChunkSource<VolumeChunkSource
...
+      // if the volume is an image, get a jpeg
+      path = this.getPath(chunkPosition, chunkDataSize, params);
     }
+    let decoder = this.getDecoder(params);
     return sendHttpRequest(
                openShardedHttpRequest(params.baseUrls, path), 'arraybuffer', cancellationToken)
-        .then(response => decodeRawChunk(chunk, response));
+        .then(response => decoder(chunk, response));
+  }
+  getPath(chunkPosition: Float32Array, chunkDataSize: Float32Array, params: any) {
```
jbms: No need to pass params, since they can be accessed as this.parameters.

src/neuroglancer/datasource/dvid/frontend.ts
```diff
@@ -123,10 +122,12 @@ export class VolumeDataInstanceInfo extends DataInstanceInfo {
                     mat4.create(), vec3.multiply(vec3.create(), lowerVoxelBound, voxelSize)),
                 baseVoxelOffset: lowerVoxelBound,
                 upperVoxelBound: vec3.subtract(vec3.create(), upperVoxelBound, lowerVoxelBound),
-                volumeType: this.volumeType, volumeSourceOptions,
+                volumeType: this.volumeType,
+                volumeSourceOptions,
```
jbms: specify compressedSegmentationBlockSize as either undefined or vec3.fromValues(8, 8, 8) depending on the volume type.

src/neuroglancer/datasource/dvid/frontend.ts
jbms: (That is only needed if you decide to add an explicit encoding parameter)

jbms:
Thanks Alex, this is great.  There are just a few ways I think the code can be simplified, described in my comments on specific code lines.

westona:
Thanks for the review Jeremy! I decided to go with the explicit enum--I saw that that's what we're doing for Tiles, so it makes sense to be consistent.

src/neuroglancer/datasource/dvid/backend.ts
```diff
@@ -15,20 +15,30 @@
...
 import {openShardedHttpRequest, sendHttpRequest} from 'neuroglancer/util/http_request';
+import {RPC} from 'neuroglancer/worker_rpc';
 
 const TILE_CHUNK_DECODERS = new Map<TileEncoding, ChunkDecoder>([
   [TileEncoding.JPEG, decodeJpegChunk],
 ]);
 
 @registerChunkSource(VolumeChunkSourceParameters)
 class VolumeChunkSource extends ParameterizedVolumeChunkSource<VolumeChunkSourceParameters> {
+  constructor(rpc: RPC, options: any) {
```
jbms: You can delete this constructor now, since the superclass constructor now does everything needed (the this.parameters = options['parameters']; is redundant with what is done in the superclass constructor)

PR #33, opened 2017-05-05
Add highlight feature

mtyka:
Adding a generic highlight mechanism. This allows the selection of a set of IDs
separate from their visibility status and separate from the
hover-selected status. This can be used to select a group of IDs to be
merged or otherwise processed, for example.

### DIFF
--- a/src/neuroglancer/segmentation_display_state/frontend.ts
+++ b/src/neuroglancer/segmentation_display_state/frontend.ts
@@ -27,6 +27,7 @@ import {Uint64} from 'neuroglancer/util/uint64';
 import {UseCount} from 'neuroglancer/util/use_count';
 import {SharedObject} from 'neuroglancer/worker_rpc';
 import {NullarySignal} from 'neuroglancer/util/signal';
+import {rgbToHsv, hsvToRgb} from 'neuroglancer/util/colorspace';
 
 export class Uint64MapEntry {
   constructor(public key: Uint64, public value: Uint64) {}
@@ -79,6 +80,7 @@ export class SegmentSelectionState extends RefCounted {
 export interface SegmentationDisplayState extends VisibleSegmentsState {
   segmentSelectionState: SegmentSelectionState;
   segmentColorHash: SegmentColorHash;
+  saturation: TrackableAlphaValue;
 }
 
 export interface SegmentationDisplayStateWithAlpha extends SegmentationDisplayState {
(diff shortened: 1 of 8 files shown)


src/neuroglancer/segmentation_display_state/frontend.ts
```diff
@@ -128,6 +130,25 @@ export function getObjectColor(
   const color = tempColor;
   color[3] = alpha;
   displayState.segmentColorHash.compute(color, objectId);
+
```
jbms: I think you can adjust saturation by just interpolating the color with (1, 1, 1), as is done for the isSelected case.

src/neuroglancer/sliceview/segmentation_renderlayer.ts
```diff
@@ -153,6 +157,17 @@ uint64_t getMappedObjectId() {
     alpha = uNotSelectedAlpha;
   }
   vec3 rgb = segmentColorHash(value);
+`;
+
+    // Override color for all highlighted segments.
+    fragmentMain += `
+  if(${this.hashTableManagerHighlighted.hasFunctionName}(value)) {
```
jbms: To avoid extra texture fetches in the case this functionality isn't being used, the same conditional logic used for the equivalences table could also be used.

src/neuroglancer/sliceview/segmentation_renderlayer.ts
```diff
@@ -153,6 +157,17 @@ uint64_t getMappedObjectId() {
...
+    rgb = vec3(0.2,0.2,2.0);
+    saturation = 1.0;
+  };
```
jbms: Extra semicolon

src/neuroglancer/sliceview/segmentation_renderlayer.ts
```diff
@@ -153,6 +157,17 @@ uint64_t getMappedObjectId() {
...
+`;
```
jbms: Extra semicolon

src/neuroglancer/sliceview/segmentation_renderlayer.ts
```diff
@@ -153,6 +157,17 @@ uint64_t getMappedObjectId() {
     alpha = uNotSelectedAlpha;
   }
   vec3 rgb = segmentColorHash(value);
+`;
```
jbms: Extra semicolon

jbms: Nevermind, I misread it.

jbms:
Thanks for adding this feature.

In general, the way the coloring works seems to be somewhat specific to the particular case of selecting objects to merge --- that is okay, but since the JSON state representation will need to remain backward compatible, I want to be cautious about what is added there (certainly the existing state representation already has some warts).  For your application, is it necessary for the state representation to be backwards compatible over a long time period?  If not, I'd be inclined to prefix the highlighted segments state property with "experimental", and additionally to only enable the key binding if some property is set in the code.

In SegmentationRenderLayer, you need to make sure to set up a signal handler to call this.redrawNeeded.dispatch() if the saturation changes, as is done for selectedAlpha and notSelectedAlpha.

For saturation, do you see that as something users will manually adjust while using the program, or would it generally only be set programatically, e.g. to avoid ambiguity with the color used for indicating highlighted segments?  If the latter, the user interface for toggling it could be removed.

jbms:
Am I correct in assuming these new changes aren't intended to be related to this pull request?  Should we just close this pull request for now and reopen later if the changes are to be merged in?

mtyka:
Yes, let's hold off

jbms:
Already merged.

PR #34, opened 2017-05-09
Added options to hide certain UI elements

timblakely:
Defaults to showing the full display. Also fixes noimplicitany error in layout.ts=>Handler

### DIFF
--- a/src/neuroglancer/viewer.ts
+++ b/src/neuroglancer/viewer.ts
@@ -55,6 +55,21 @@ export function validateLayoutName(obj: any) {
   return layout[0];
 }
 
+export interface UIOptions {
+  showHelpButton?: boolean;
+  showLayerDialog?: boolean;
+  showLayerPanel?: boolean;
+  showLocation?: boolean;
+}
+export interface ViewerOptions extends UIOptions {}
+
+const defaultViewerOptions: ViewerOptions = {
+  showHelpButton : true,
+  showLayerDialog : true,
+  showLayerPanel : true,
+  showLocation : true,
+};
+
 export class Viewer extends RefCounted implements ViewerState {
   navigationState = this.registerDisposer(new NavigationState());
   perspectiveNavigationState = new NavigationState(new Pose(this.navigationState.position), 1);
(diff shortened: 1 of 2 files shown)


src/neuroglancer/viewer.ts
```diff
@@ -31,7 +31,10 @@ import {TrackableBoolean} from 'neuroglancer/trackable_boolean';
 import {TrackableValue} from 'neuroglancer/trackable_value';
 import {RefCounted} from 'neuroglancer/util/disposable';
 import {vec3} from 'neuroglancer/util/geom';
-import {GlobalKeyboardShortcutHandler, KeySequenceMap} from 'neuroglancer/util/keyboard_shortcut_handler';
```
jbms: Can you revert (i.e. via rebasing) unrelated formatting changes?

src/neuroglancer/viewer.ts
```diff
@@ -55,6 +58,21 @@ export function validateLayoutName(obj: any) {
...
 
+export interface UIOptions {
+  showLayerPanel?: boolean;
+  showHelpButton?: boolean;
+  showLocation?: boolean;
+  showLayerDialog?: boolean;
+}
+export interface ViewerOptions extends UIOptions {}
+
+let DefaultViewerOptions: ViewerOptions = {
```
jbms: Make this const, and call it defaultViewerOptions instead.

src/neuroglancer/widget/shader_code_widget.ts
```diff
@@ -16,7 +16,7 @@
 
 import 'codemirror/addon/lint/lint.js';
 
-import CodeMirror from 'codemirror';
```
jbms: Please also revert this unrelated change.  Depending on how the typings are defined, how webpack is configured, and in what mode typescript is run, one or the other import style is needed.  However, I believe the existing style is what works best with the current neuroglancer configuration.

timblakely:
Sorry about the formatting changes! <3 clang-format + formatOnSave... maybe a little bit too much. PTAL.

PR #35, opened 2017-05-16
User event emitter prototype

timblakely:
This is a prototype for an event handling system. It's a bit lower level than `invokeAction` in `LayerManager`, intended to be the place to hook all user-generated events (e.g. clicks, mousemoves, etc). It can also implement custom events like mousedrag, removing the need for `startRelativeMouseDrag`. 

Events are hooked by using the `.on()` syntax, and each handler is strongly typed. To disable a handler, use the `Disposer` functions returned from `.on()`.  For example, to hook mouse movements but disable them on a mouse down, you'd do the following:

```
let disposers = this.eventEmitter.on({
  mousemove: (e: MouseEvent) => console.log('Mouse moved:', e),
  mousedown: () => disposers['mousemove'](),
});
```

One major advantage of this is that it allows external apps (e.g. Wintermute) to hook into user events that occur on layers. It currently does not pass any additional state/context along with the event, but I'd imagine that would be more appropriate somewhere within layerManager as this is solely for hooking user actions.

Todos:

- Implement event bubbling (might only be necessary for custom events, since DOM events inherently bubble)
- Integration with layerManager's action system
- Keyboard support 
- Providing additional context (e.g. which mesh/ID was clicked on, what layer, etc)

### DIFF
--- a/src/neuroglancer/ui/user_events.ts
+++ b/src/neuroglancer/ui/user_events.ts
@@ -1,0 +1,85 @@
+import {RefCounted} from 'neuroglancer/util/disposable';
+
+interface TypedEventListener<T> extends EventListener {
+  (e: T): void;
+}
+
+type ListenerConfig<T> = TypedEventListener<T> | [TypedEventListener<T>, boolean];
+
+// Used to perform type narrowing on a ListenerConfig.
+function getListenerConfig<T>(listenerConfig: ListenerConfig<T>): [TypedEventListener<T>, boolean|undefined] {
+    if (Array.isArray(listenerConfig)) {
+      return listenerConfig;
+    } else {
+      return [listenerConfig, false];
+    }
+}
+
+interface EventConfig {
+  mousemove?: ListenerConfig<MouseEvent>;
+  mousedown?: ListenerConfig<MouseEvent>;
+  mouseleave?: ListenerConfig<MouseEvent>;
+  dblclick?: ListenerConfig<Event>;
+}
+
+function keys<T>(obj: T): [keyof T] {
+  return Object.keys(obj) as any;
+}
+
+type ListenerType = EventConfig[keyof EventConfig];
+type DisposerRef = () => void;
+type Disposers = {
+  [K in keyof EventConfig]: DisposerRef;
+}
+
+export class UserEventEmitter extends RefCounted {
+
+  private activeListeners = new Map<DisposerRef, ListenerType>();
+
+  constructor(private element: HTMLElement) {
+    super();
+  }
+
+  on(eventConfig: EventConfig) {
+    let disposers: Disposers = {};
+    for (let key of keys(eventConfig)) {
+      let {element} = this;
+      let listenerConfig = eventConfig[key];
+      if (listenerConfig === undefined) {
+        continue;
+      }
+      let [listener, useCapture] = getListenerConfig(listenerConfig);
+      element.addEventListener(key, listener, useCapture)
+      let disposer = () => {
+        element.removeEventListener(key, listener);
+        this.removeDisposer(disposer);
+      }
+      this.activeListeners.set(disposer, listener);
+      disposers[key] = disposer;
+    }
+    return disposers;
+  }
+
+  off(disposers: DisposerRef[]|DisposerRef) {
+    if (!Array.isArray(disposers)) {
+      disposers = [disposers];
+    }
+
+    for (let disposer of disposers) {
+      disposer();
+    }
+  }
+
+  disposed() {
+    for (let disposer of this.activeListeners.keys()) {
+      disposer();
+    }
+    super.dispose();
+  }
+
+  private removeDisposer(disposer: DisposerRef) {
+    if(this.activeListeners.has(disposer)) {
+      this.activeListeners.delete(disposer);
+    }
+  }
+}
--- a/src/neuroglancer/ui/user_events.ts
+++ b/src/neuroglancer/ui/user_events.ts
@@ -4,23 +4,65 @@ interface TypedEventListener<T> extends EventListener {
   (e: T): void;
 }
 
-type ListenerConfig<T> = TypedEventListener<T> | [TypedEventListener<T>, boolean];
+type ListenerConfig<T> = TypedEventListener<T>|[TypedEventListener<T>, boolean];
 
 // Used to perform type narrowing on a ListenerConfig.
-function getListenerConfig<T>(listenerConfig: ListenerConfig<T>): [TypedEventListener<T>, boolean|undefined] {
-    if (Array.isArray(listenerConfig)) {
-      return listenerConfig;
-    } else {
-      return [listenerConfig, false];
-    }
+function getListenerConfig<T>(listenerConfig: ListenerConfig<T>):
+    [TypedEventListener<T>, boolean | undefined] {
+  if (Array.isArray(listenerConfig)) {
+    return listenerConfig;
+  } else {
+    return [listenerConfig, false];
+  }
 }
 
-interface EventConfig {
+// Flatten and type narrow disposers.
+function getDisposerList<T>(inputDisposers: Disposers | DisposerRef[] | DisposerRef):
+    DisposerRef[] {
+  if (Array.isArray(inputDisposers)) {
+    return inputDisposers;
+  };
+  if (typeof(inputDisposers) === 'function') {
+    return [inputDisposers];
+  }
+  let list = [];
+  for (let key of keys(inputDisposers)) {
+    list.push(inputDisposers[key]);
+  }
+  return list;
+}
+
+export interface MouseDragEvent extends MouseEvent {
+  start: DragPosition;
+  current: DragPosition;
+  delta: DragPosition;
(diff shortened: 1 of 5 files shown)


src/neuroglancer/ui/user_events.ts
```diff
@@ -0,0 +1,220 @@
+import {RefCounted} from 'neuroglancer/util/disposable';
+
+interface TypedEventListener<T> extends EventListener {
+  (e: T): void;
```
timblakely: I'm thinking about allowing additional context to be generated here. Instead of returning the raw event (MouseEvent, WheelEvent, etc), it would return something like this:

```
interface NeuromancerEvent<EventType, ContextType> {
  rawEvent: EventType;
  context: ContextType;
}

interface TypedEventListener<E, C> extends EventListener {
 (e: NeuromancerEvent<E, C>): void;
}

type ListenerConfig<E, C> = TypedEventListener<E, C>|[TypedEventListener<E, C>, boolean];
```

Which then could be used like so:

```
interface MouseContext {
  layerId: string;
  active: boolean; // <-- e.g. from the mouse 
  // Other things that would be useful, e.g. fields from Viewer.mouseState
}

export interface DomEvents {
  mousemove?: ListenerConfig<MouseEvent, MouseContext>;
}
```

jbms:
Hi Tim,

Certainly there is much room for improvement in Neuroglancer's existing event handling.  I do like the more uniform handling of mouse dragging.  One concern though is that it adds a lot of machinery for keeping track of which events have handlers, but is still a low level mechanism quite close to the DOM event mechanism.  Would just registering events on the element property of RenderedDataPanel perhaps solve your immediate needs?

As far as providing additional context to the handler, it seems like the code that binds the handler would in general have access to that state directly, such as the MouseSelectionState object.  Do you think this would be problematic?

A higher level mechanism, in which there is an added level of indirection of specifying handlers by name, as is used for key handling, would have certain advantages (like making it easier to change the bindings of existing actions, and supporting a simple self-documentation feature of listing the mouse binding/action associations) but would also add some constraints.  I'm not sure whether a higher level interface would be useful/work for your purposes, but I'd be interested in discussing.

timblakely:
You're right in that this is quite low level at the moment; it's only intending to be a centralized mechanism for enabling/disabling listeners. The UI handlers are spread out throughout the wider codebase, making it basically impossible to disable any of the baked-in interactions in Neuroglancer. AFAICT, there isn't any way to selectively disable any of Neuroglancer's event handling at the moment, nor is there a way to hook something like "mouseup", short of attaching listeners to the DOM (whose bubbling could be stopped by one of the Neuroglancer event handlers).

The main idea of this is to be able to a) allow hooking of events from outside Neuroglancer (e.g. a dependent project)  b) selectively enable/disable Neuroglancer event binding. Centralizing user-event hooking makes adding/removing event listeners quite easy when contexts change and/or elements are added or removed from the app. Neuroglancer doesn't do much of this at the moment since it's UI is basically static, but dependent projects will almost certainly do more. The future higher level "actions" could build on top of this, selectively adding DOM listeners when the action is enabled via `.on()` and removing them via `.off()`

jbms:
The existing registerEventListener method of RefCounted was intended to make it easy to remove events when not needed --- if a group of event listeners is needed for a particular action, they can all be registered with the same RefCounted object, and then removed as a group by calling dispose().  Currently the registerEventListener method does 3 things, though --- it adds the event listener, constructs a nullary function that removes it, and registers that removal function as a disposer.  Perhaps it would be more convenient if there were an interface for just adding the event listener and returning the removal function without registering it on any particular RefCounted object?

timblakely:
> Perhaps it would be more convenient if there were an interface for just adding the event listener and returning the removal function without registering it on any particular RefCounted object?

Yup, my thoughts exactly :) And that's what the UserEventEmitter does, in a nutshell. It just takes most of that logic and centralizes it in a class. UEE is itself refcounted, which handles disposal of attached eventListeners, much like the existing implementation it replaces.

Here's where I'm coming from: As an "external user" I'd *love* to have an interface into Neuroglancer like like:

```
class AnExternalApp {
  private neuroglancer = new Neuroglancer();  // née Viewer

  constructor() {
    let {neuroglancer} = this;

    // Handle click events.
    neuroglancer.on({
      click: e => {
        /* ... */
        e.preventDefault();  // Prevents Neuroglancer's default action
      },
    });

    // Add a bunch of event hooks at once.
    let manyHooks = neuroglancer.on({
      mousemove: e => console.log('mouse moved'),
      mousedrag: e => console.log('mouse dragged'),
      prefetchdone: e => console.log('prefetching done'),
    });

    // Handle object selection
    neuroglancer.on({objectselected: e => { /* ... */ }});

    // Remove the mousemove listener
    neuroglancer.on({
      layerremoved: e => neuroglancer.off(manyHooks['mousemove']),
    });
    // Or alternatively: neuroglancer.on({layerremoved: e => hooks['mousemove']()});

    // Remove all hooks added by one call.
    neuroglancer.off(manyHooks);
  }
}
```
When using Neuroglancer as an external library, it's much more pleasant to use an API like `neuroglancer.on({posechanged: e => console.log('new pose's position:', e.position)` versus reaching deep down into Neuroglancer:
```
viewer.perspectiveNavigationState.pose.changed.add(() => {
    let position = viewer.perspectiveNavigationState.pose.position;
    console.log('new pose's position:', position);
});
```
The UEE approach obviously doesn't get us there yet, but it's a foundation to build a the API on. And I figured if it works for external users why not test out and leverage a similar mechanism internally? :)

jbms:
As far as separating the logic of adding the event and creating the removal function from the registering of the removal function as a disposer, I was imagining something much more lightweight:

function addDisposableEventListener(target: EventTarget, eventType: string, listener: any, any?: any) {
  target.addEventListener(eventType, listener, arg);
  return () => target.removeEventListener(eventType, listener, arg);
}

For handling mouse (or keyboard) events on perspectiveview and sliceview, I agree that there needs to be a more convenient way to do that than having to bind events on the individual PerspectivePanel or SliceViewPanel, since as the layout changes the panels also change, and it would be annoying to have to handle them individually if the desired behavior should be uniform over all panels of a given type or both sliceview and perspectiveview panels.

If the high level mouse handling were in place, that would be pretty easy to achieve, I think: the viewer would have a mouse event map for sliceview and for perspectiveview, as well as a joint event map used for both, that maps mouse events to action names.  Any sliceviewpanels or perspectivepanels created by that viewer would use these event maps.  An external library could both modify these event maps, as well as the mapping from action names to handler functions, in order to define high level mouse and keyboard behavior.

Just using the raw DOM events is tricky though --- I think preventing the neuroglancer default behavior can't be done the way you suggest --- stopImmediatePropagation() would come closest, but still would be tricky to make work because the neuroglancer default event handlers would normally have been added first.

As far as handling the perspectiveNavigation pose (and similar things) changing, I'm not sure I agree with you there.  The perspectiveNavigation object identity doesn't change when the state changes, so there is no risk in using that object's signals directly.  Just accessing the object directly seems more modular to me.  One thing consolidating a bunch of events in one place does help with is that it makes it easier to find them, but that could also be addressed with better documentation.

For your example, if you shorten your example to (note that viewer.perspectiveNavigationState.pose.position is the same as viewer.navigationState.pose.position):
const {position} = viewer.navigationState;
position.changed.add(() => {
    console.log('new pose's position:', position);
});

it starts to look pretty clean.  I think it would also be reasonable to add a `position` property to `viewer` to simplify this further.

timblakely:
> If the high level mouse handling were in place, that would be pretty easy to achieve, I think: the viewer would have a mouse event map for sliceview and for perspectiveview, as well as a joint event map used for both, that maps mouse events to action names. Any sliceviewpanels or perspectivepanels created by that viewer would use these event maps. An external library could both modify these event maps, as well as the mapping from action names to handler functions, in order to define high level mouse and keyboard behavior.

Yup, this is along the lines of what I was looking for: a stable JS API into Neuroglancer. As long as these actions could be either modified or disabled entirely this should work.

> Just using the raw DOM events is tricky though --- I think preventing the neuroglancer default behavior can't be done the way you suggest --- stopImmediatePropagation() would come closest, but still would be tricky to make work because the neuroglancer default event handlers would normally have been added first.

In its current form you're correct: it wouldn't work. That said, there is existing prior art in various JS library APIs out there that achieve this by enforcing that any user-generated hooks are called first, allowing the propagation of the event to stop before it arrives at in the default Neuroglancer handlers.

Regarding the pose example, either way what it comes down to is the desire to create an API into Neuroglancer that can be used by external projects. Your example works in it's current form, but will break any dependent project if any of the following happen:

1. navigationState is renamed
2. position is renamed
3. The state hierarchy is modified
4. Navigation is refactored in any way

However the API into Neuroglancer is implemented - actions, function calls, accessing public member variables, whatever - it must remain constant or risk breaking any project that depends on it. If we're confident that any point that external projects can touch within Neuroglancer won't change (e.g. navigationState, layerManager, Sliceview), then reaching into Neuroglancer like this is fine. If any of those pieces could change, that requires an additional layer on top so that the external API remains stable during any internal refactors.

PR #37, opened 2017-05-21
Feat/viewer prefetch

jbms:


timblakely:
Finally getting around to this. Can you sync this branch? There have been a bunch of formatting changes since then that are wrecking havoc with my merge :(

jbms:
I believe this branch is already rebased on top of the master branch.  I did reformat the master branch, which unfortunately makes merging more difficult, but I'm not sure there is any way I can help.

timblakely:
Ok, np. I've just been in merge hell today (see other PR from pre-formatting) :)  Is there a way that we can add a check to enforce that PRs/commits are all formatted before merging? Looks like HEAD fails `gulp check-format` ATM.

jbms:
Unfortunately the formatting depends on the specific version of clang-format used.  If you npm update clang-format, it should be updated to clang-format 1.0.52 which should pass.

jbms:
Travis-CI can do presubmit checks like that --- I just haven't gotten around to setting it up.  However, it might take a bit of work to have it check only the formatting of the parts that have changed.  Without that, there will be problems any time clang format is updated.

timblakely:
Good to know, I'll update `clang-format` now.

If it gets set up as a presubmit after the entire codebase passes `gulp check-format`, can't you just ensure that `check-format` passes after merging in the pull request since it covers a) the new code that's committed and b) the existing codebase that should pass check-format anyway?

src/neuroglancer/viewer.ts
```diff
@@ -83,30 +104,46 @@ export class Viewer extends RefCounted implements ViewerState {
...
-      this.layerManager, this.chunkManager, this.worker, this.layerSelectedValues,
-      this.navigationState.voxelSize);
+  layerSpecification: LayerListSpecification;
   layoutName = new TrackableValue<string>(LAYOUTS[0][0], validateLayoutName);
 
   state = new CompoundTrackable();
 
   private options: ViewerOptions;
 
-  constructor(public display: DisplayContext, options: ViewerOptions = {}) {
```
timblakely: It appears you need to add the showLayerDialog=false flag to options to suppress the initial overlay. It probably makes sense to change the check at 186 to include a check for dataContext and/or WatchableVisibilityPriority as well.

timblakely:
That showLayerDialog change would be nice to have., but LGTM otherwise.  Feel free to merge.

PR #38, opened 2017-05-30
Feature: ClipBounds

timblakely:
Add a shared_trackable_value class that allows two-way communication of values from frontend<->backend. Example use case is for applying (optional) ClipBounds to meshes.

### DIFF
--- a/src/neuroglancer/datasource/brainmaps/backend.ts
+++ b/src/neuroglancer/datasource/brainmaps/backend.ts
@@ -20,6 +20,7 @@ import {registerChunkSource} from 'neuroglancer/chunk_manager/backend';
 import {ChangeStackAwarePayload, HttpCall, makeRequest, MeshFragmentPayload, SkeletonPayload, SubvolumePayload} from 'neuroglancer/datasource/brainmaps/api';
 import {ChangeSpec, MeshSourceParameters, SkeletonSourceParameters, VolumeChunkEncoding, VolumeSourceParameters} from 'neuroglancer/datasource/brainmaps/base';
 import {decodeJsonManifestChunk, decodeTriangleVertexPositionsAndIndices, FragmentChunk, ManifestChunk, ParameterizedMeshSource} from 'neuroglancer/mesh/backend';
+import {Bounds} from 'neuroglancer/segmentation_display_state/base';
 import {decodeSkeletonVertexPositionsAndIndices, ParameterizedSkeletonSource, SkeletonChunk} from 'neuroglancer/skeleton/backend';
 import {decodeCompressedSegmentationChunk} from 'neuroglancer/sliceview/backend_chunk_decoders/compressed_segmentation';
 import {decodeJpegChunk} from 'neuroglancer/sliceview/backend_chunk_decoders/jpeg';
@@ -27,8 +28,9 @@ import {decodeRawChunk} from 'neuroglancer/sliceview/backend_chunk_decoders/raw'
 import {ParameterizedVolumeChunkSource, VolumeChunk} from 'neuroglancer/sliceview/volume/backend';
 import {CancellationToken} from 'neuroglancer/util/cancellation';
 import {Endianness} from 'neuroglancer/util/endian';
-import {vec3Key} from 'neuroglancer/util/geom';
+import {vec3Key, vec3, decodeMorton} from 'neuroglancer/util/geom';
 import {verifyObject, verifyObjectProperty, verifyStringArray} from 'neuroglancer/util/json';
+import {Uint64} from 'neuroglancer/util/uint64';
 
 const CHUNK_DECODERS = new Map([
   [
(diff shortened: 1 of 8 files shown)


jbms:
Thanks, Tim!

The way the commits have been created/rebased is a bit strange --- for instance the top commit has me listed as the author.  Also, this excludes the top of the most recent commit in master, which modifies tslint.json.

src/neuroglancer/datasource/brainmaps/backend.ts
```diff
@@ -207,6 +211,100 @@ export class MeshSource extends ParameterizedMeshSource<MeshSourceParameters> {
     return makeRequest(parameters['instance'], httpCall, cancellationToken)
         .then(response => decodeFragmentChunk(chunk, response));
   }
+
+  filterFragments(fragmentIds: string[]|null, clipBounds: Bounds) {
```
jbms: Instead of needing a separate filterFragments method, the fragments could just be filtered when the manifest is downloaded.

src/neuroglancer/datasource/brainmaps/backend.ts
```diff
@@ -207,6 +211,100 @@ export class MeshSource extends ParameterizedMeshSource<MeshSourceParameters> {
...
+      let fragmentBounds = this.getFragmentBounds(fragmentId, vec3.clone([fragmentSize, fragmentSize, fragmentSize]));  
+      if (this.boundsIntersect(fragmentBounds, clipBounds)) {
+        filteredFragments.push(fragmentId);
+      }      
+    }
+    
+    return filteredFragments;
+  }
+
+  private numBitsInPositiveValue(value: number) {
```
jbms: This can be a free function rather than a method since it doesn't depend on `this`.

src/neuroglancer/datasource/brainmaps/backend.ts
```diff
@@ -207,6 +211,100 @@ export class MeshSource extends ParameterizedMeshSource<MeshSourceParameters> {
...
+
+  private getFragmentCorner(fragmentId: string, fragmentSize: vec3) {
+    let id = new Uint64();
+    if (!id.tryParseString(fragmentId, 16)) {
+      throw new Error(`Couldn't parse fragmentId ${fragmentId} as hex-encoded Uint64`);
+    }
+    if (id.high) {
+      throw new Error(`Fragment ids > 2^32 not supported yet`);
+    }
+    const maxCoordBits = fragmentSize.map(size => this.numBitsInPositiveValue(size));
```
jbms: I think the objective of this code is to decode the z-index.  Should the max bits be based on the size of the volume (which is not known here) rather than the size of the fragment?

I think there are ways to do this with just a small number of bit operations, e.g. as shown here: https://fgiesen.wordpress.com/2009/12/13/decoding-morton-codes/
That way you can avoid the looping.

src/neuroglancer/mesh/backend.ts
```diff
@@ -243,8 +244,8 @@ export abstract class MeshSource extends ChunkSource {
     fragmentSource.meshSource = this;
   }
 
-  getChunk(objectId: Uint64) {
-    const key = getObjectKey(objectId);
+  getChunk(objectId: Uint64, clipBounds?: Bounds) {
```
jbms: Storing the clipBounds in the ManifestChunk will allow the download function to have access to them.

timblakely:
Heh, yeah, hence my headaches this morning. I think I've finally got it sorted... maybe... I hope? Working on the comments now.

src/neuroglancer/datasource/brainmaps/backend.ts
timblakely: Done. Implemented Morton compaction via magic bits (and added some support code to Uint64 as well).

src/neuroglancer/datasource/brainmaps/backend.ts
timblakely: Done.

src/neuroglancer/mesh/backend.ts
timblakely: Done.

src/neuroglancer/util/geom.ts
```diff
@@ -148,3 +149,35 @@ export function translationRotationScaleZReflectionToMat4(
...
+  let x = input;
+  x = x.and(magicBits[0]);
+  for (let i = 1; i < magicBits.length; ++i) {
+    x = x.xor(x.rshift(Math.pow(2,i))).and(magicBits[i]);
+  }
+  return x;
+}
+
+export function decodeMorton(input: Uint64) {
+  if (input.high) {
```
jbms: What is the reason for this restriction?  Seems like the morton decoding is designed to handle 64-bit input.

src/neuroglancer/util/geom.ts
```diff
@@ -148,3 +149,35 @@ export function translationRotationScaleZReflectionToMat4(
...
+    throw new Error('Fragment ids >= 2^32 not supported yet');
+  }
+  const x = compactMorton(input.rshift(0));
+  const y = compactMorton(input.rshift(1));
+  const z = compactMorton(input.rshift(2));
+  return vec3.clone([x.low, y.low, z.low]);
+}
```
jbms: Add newline at end of file

src/neuroglancer/util/geom.ts
```diff
@@ -148,3 +149,35 @@ export function translationRotationScaleZReflectionToMat4(
...
+const magicBits = [
+  new Uint64(0x49249249, 0x12492492), // AND step
+  new Uint64(0xc30c30c3, 0x30c30c30), // XOR steps
+  new Uint64(0x0f00f00f, 0xf00f00f0),
+  new Uint64(0xff0000ff, 0x00ff0000),
+  new Uint64(0x0000ffff, 0x00ff0000),
+  new Uint64(0x001fffff, 0x00000000),
+];
+
+export function compactMorton(input: Uint64) {
```
jbms: In general I think avoiding creating new objects is helpful, though I'm not sure exactly how it affects performance, and this isn't necessarily extremely performance sensitive, so I'm not too worried about it in this case.

src/neuroglancer/shared_trackable_value.ts
```diff
@@ -0,0 +1,55 @@
+/**
```
jbms: This is very nice.  However, almost exactly the same code is now in src/neuroglancer/shared_watchable_value.ts, which was added as part of the prefetching support.

src/neuroglancer/segmentation_display_state/base.ts
```diff
@@ -17,18 +17,41 @@
...
+export interface Bounds {
+  center: vec3;
+  size: vec3;
+};
+
+export class ClipBounds extends SharedTrackableValue<Bounds> {
+  toJSON(): {}|null {
+    let bounds = this.value;
+    if (bounds === undefined) { 
+      return null;
```
jbms: Returning undefined instead of null may be more convenient.  Because `JSON.stringify({clipBounds: undefined}) === '{}'`, you can simplify the code that calls this method.

src/neuroglancer/segmentation_display_state/base.ts
```diff
@@ -17,18 +17,41 @@
 import {SharedDisjointUint64Sets} from 'neuroglancer/shared_disjoint_sets';
 import {Uint64Set} from 'neuroglancer/uint64_set';
 import {Uint64} from 'neuroglancer/util/uint64';
+import {vec3} from 'neuroglancer/util/geom';
+import {SharedTrackableValue} from 'neuroglancer/shared_trackable_value';
+
+export interface Bounds {
+  center: vec3;
+  size: vec3;
+};
```
jbms: No semicolon needed

src/neuroglancer/segmentation_display_state/base.ts
```diff
@@ -17,18 +17,41 @@
...
+    if (bounds === undefined) { 
+      return null;
+    }
+    let result = {
+      center: [...bounds.center.values()],
+      size: [...bounds.size.values()],
+    };
+    return result;
+  }
+}
```
jbms: I think it would make sense to move the restore logic into a restoreState method here rather than in segmentation_user_layer.ts

src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -159,6 +163,17 @@ export class SegmentationUserLayer extends UserLayer {
         });
       }
     });
+
+    verifyObjectProperty(x, 'clipBounds', y => {
+      if (y === undefined) {
+        return;
+      }
+      let bounds = {
+        center: vec3.clone(y['center']),
```
jbms: Perhaps use verifyObjectProperty(y, 'center', ...), and parseFiniteVec, etc. to more thoroughly check format

src/neuroglancer/util/geom.ts
timblakely: Yeah, the Morton decoding likely isn't a performance bottleneck. I just grouped 'em in a global const out of habit, though they aren't exported so they shouldn't pollute the global namespace. That, and you can loop through 'em for more compact code :)

src/neuroglancer/util/geom.ts
timblakely: Ack, sorry, that's a leftover from when I was only doing the 32 bit version. Will remove.

src/neuroglancer/shared_trackable_value.ts
timblakely: SGTM. Would you prefer dropping this class entirely in favor of using SWV then?

timblakely:
Done. PTAL

src/neuroglancer/segmentation_user_layer.ts
timblakely: Done.

src/neuroglancer/segmentation_display_state/base.ts
timblakely: Done.

src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -183,6 +201,13 @@ export class SegmentationUserLayer extends UserLayer {
     if (segmentEquivalences.size > 0) {
       x['equivalences'] = segmentEquivalences.toJSON();
     }
+    let {clipBounds} = this.displayState;
+    if (clipBounds.value) {
+      x['clipBounds'] = JSON.stringify({
```
jbms: These toJSON() methods are actually supposed to return a JSON value as a javascript value, not the JSON string encoding.

timblakely: Done.

PR #41, opened 2017-06-02
Feature: Layer and volumeLayer registration is trackable

timblakely:
Allow external/dependent projects to listen for layer types to register themselves. This fixes a race condition where a Viewer would be created and its state set to show some layers (e.g. 'image' and 'mesh') _before_ the layers had time to register themselves.

### DIFF
--- a/src/neuroglancer/layer_specification.ts
+++ b/src/neuroglancer/layer_specification.ts
@@ -26,6 +26,7 @@ import {vec3} from 'neuroglancer/util/geom';
 import {verifyObject, verifyObjectProperty, verifyOptionalString} from 'neuroglancer/util/json';
 import {NullarySignal, Signal} from 'neuroglancer/util/signal';
 import {Trackable} from 'neuroglancer/util/trackable';
+import {TrackableMap} from 'neuroglancer/trackable_map';
 
 export function getVolumeWithStatusMessage(
     chunkManager: ChunkManager, x: string,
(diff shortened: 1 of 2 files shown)


jbms:
TrackableMap looks like a potentially useful addition.  However, I think your particular issue can be solved more easily by delaying the running of your setup code, e.g. whatever creates Viewer instances.  For example, in src/main.ts used for the demo viewer, the setup code is run as an event handler for DOMContentLoaded.

timblakely:
I had tried the delay approach, but unfortunately there were a few issues:

1) we don't have much control over when/how soon the angular code runs
2) I tried setting arbitrary delays of up to 1500ms and was still occasionally seeing it
3) I'm using code splitting in my webpack config to keep recompile times down, which means that some code is asynchronously loaded after the DOMContentLoaded

There's no functional change to Neuroglancer itself other than exposing these events as NullarySignals, but if you prefer I can keep the changes in the annotation-specific branch of Neuroglancer.

jbms:
Is all of the neuroglancer code not contained in a single bundle?  In the default neuroglancer webpack configuration, incremental rebuild via the webpack-dev-server seems to be pretty fast, though maybe that doesn't work as well with angular2.  Can you just have the initialization code (that creates the Viewer instances) run after both the DOMContentLoaded event and the neuroglancer bundle has loaded?  It just seems like you will run into more unexpected behavior if you are trying to use neuroglancer as a library in a partially initialized/partially loaded state.

PR #44, opened 2017-07-14
Increase visibility of some private members

pacher:
Hello.
First, thank you very much for your work. It's truly awesome and helped us really a lot.

In our project we are extending some of neuroglancer classes to implement custom functionality specific to our needs. But we need access to some members of base classes which are declared `private`. It is possible with typescript, so it does not block us, but produces unnecessary clatter in diffs and might be harder to maintain.
Is there any strong reason to keep them private? Otherwise we suggest to promote at least the ones we use to `protected` to make our work easier.

Thank you in advance
Best regards, Pavel


pacher:
I signed it!

PR #48, opened 2017-08-03
Ensure that only a single web worker is created.

timblakely:
Needed to ensure that when a viewer is created/destroyed, it does not create multiple workers that all listen to and respond to each message in parallel.

### DIFF
--- a/src/neuroglancer/viewer.ts
+++ b/src/neuroglancer/viewer.ts
@@ -57,13 +57,14 @@ export function validateLayoutName(obj: any) {
   return layout[0];
 }
 
+const WORKER_RPC = new RPC(new Worker('chunk_worker.bundle.js'));
+
 export class DataManagementContext {
-  chunkQueueManager =
-      new ChunkQueueManager(new RPC(new Worker('chunk_worker.bundle.js')), this.gl, {
-        gpuMemory: new AvailableCapacity(1e6, 1e9),
-        systemMemory: new AvailableCapacity(1e7, 2e9),
-        download: new AvailableCapacity(32, Number.POSITIVE_INFINITY)
-      });
+  chunkQueueManager = new ChunkQueueManager(WORKER_RPC, this.gl, {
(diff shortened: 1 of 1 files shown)


jbms:
Can you explain more in what case this problem can occur?  The idea of the existing code is that there is one DataManagementContext per worker, so if you want to ensure there is a single worker, just reuse the DataManagementContext.

timblakely:
If I have ViewerA  embedded in the page and remove it, then create a new ViewerB, WorkerA still exists and both WorkerA and WorkerB listen/respond to ViewerB's messages.

jbms:
Indeed, the worker is currently never terminated.  That should probably be fixed.  However, I don't see how the old worker would receive messages from the new viewer.

Is it feasible to create just a single DataManagementContext and pass it to both Viewer instances, e.g.:

new Viewer(display, {dataContext})

?

timblakely:
The datacontext is explicitly set to a new DataManagmentContext regardless of the options passed (that should probably be fixed too :). It also looks like the DMC needs the GL context, which might not exist after the Viewer's canvas has been removed from the DOM, which means when creating a new DisplayContext you also need to create a new DMC, which in turn creates new ChunkQueueManagers and ChunkManagers (dependent on the GL instance).

jbms:
The:

      dataContext = new DataManagementContext(display.gl),

line in the Viewer constructor only creates a new DataManagementContext if one isn't specified in the options.

I can see there is a clear need to be able to dispose of a DataManagementContext, though.  I'm working on a fix now.

timblakely:
Huh, yeah, sorry, haven't seen that style of default object assignment much before. Thanks for looking into it!

jbms:
Take a look at 48f8cd2a9ad3f16805d2d5cd5112cdae4c1f024b.  Now calling the dispose() method on the Viewer terminates the worker and removes all of the added DOM elements.

timblakely:
Yup, that seems like it'd work. Closing this.

Issue #49, opened 2017-09-07
Segmentation Downsampling


william-silversmith:
Hi Jeremy!

While working on Seung Lab's pipeline processing pipeline, I found that it's possible to use numpy to 2x2x1 downsample segmentation correctly for at least one mip level at reasonable performance. Would you be interested in a contribution? Two benefits are 1) Recursively downsampled segmentation mip levels won't visually march diagonally as they load. 2) MIP 1 is 100% accurate as in the most frequent pixel in a 2x2 grid is always chosen, so e.g. agglomeration algorithms can depend on it.

I have a 2x2x2 implementation as well that works in theory but that I haven't tested extensively.

I've written it up here: https://medium.com/towards-data-science/countless-high-performance-2x-downsampling-of-labeled-images-using-python-and-numpy-e70ad3275589

Thanks for the many enjoyable hours of neuroglancing. ^_^.

Will

jbms:
Hi,

This would certainly be a welcome contribution.  If I understand correctly, you're proposing to add this new most-frequent-label downsampling method in addition to the averaging and striding methods already in downsample.py, and to modify volume.py to use it when appropriate?

william-silversmith:
Yep! I'll put something together within a week or so. Glad I can contribute something back. :)

PR #50, opened 2017-09-16
Add 2x2 Downsample by Mode

william-silversmith:
See issue #49 

Adds 2x2x1 factor downsampling using numpy. You can read about the algorithm here: https://medium.com/@willsilversmith/countless-high-performance-2x-downsampling-of-labeled-images-using-python-and-numpy-e70ad3275589


william-silversmith:
I signed it!

william-silversmith:
One issue I just realized, Neuroglancer uses ZYX coordinates and this is written with XYZ in mind. I'll correct this.

PR #52, opened 2017-09-28
Fix sliceview debug mode

alexbaden:
Just a syntax error in one of the variable names. I think everything else is identical to the vertex shader and working.


jbms:
Thanks for this fix.

PR #59, opened 2017-10-26
Render Datasource Improvements

alexbaden:
Fix a rounding bug that occurs when setting clip bounds for projects with multiple levels.
Support channels.
Add support for the following query params:
- `minIntensity`: resample from this value to `maxIntensity`
- `maxIntensity`: resample from `minIntensity` 
- `filter`: local contrast normalization
- `maxTilesToRender`: limit the number of rendered tile specs for each bounding box


jbms:
This looks good --- one thing you may want to verify though is whether the max voxel bounds returned by render are inclusive or exclusive.  The neuroglancer upper bounds are exclusive, but I think the render bounds may be inclusive.

PR #60, opened 2017-10-27
Feat(datasource/precomputed): Support skeletons

alexbaden:
Adds support for the skeleton layer in the precomputed datasource. Need to add `getSkeletonSource()` or similar to the `MultiscaleVolumeChunkSource` to add skeletons from a field in the `info` file.

### DIFF
--- a/src/neuroglancer/datasource/precomputed/backend.ts
+++ b/src/neuroglancer/datasource/precomputed/backend.ts
@@ -15,8 +15,9 @@
  */
 
 import {WithParameters} from 'neuroglancer/chunk_manager/backend';
-import {MeshSourceParameters, VolumeChunkEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/precomputed/base';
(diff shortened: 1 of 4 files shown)


jbms:
This is great --- one concern though is that the seung lab has also independently created a similar change, but using the skeleton format of the Python backend.

alexbaden:
Is that format the same, just with 32-bit headers instead of 64-bit headers?

jbms:
The first four bytes is the number of vertices, next four bytes is the number of edges.  Then there is the vertex positions.  Then there are the additional vertex attributes, if any.  The number and type of the vertex attributes is specified as part of the skeleton source parameters, so there is no metadata in the skeleton chunk itself indicating anything about the vertex attributes.  Then there is the edge index data.  So ignoring the possibility of vertex attributes, it only differs from the 64-bit format in that bytes 4-8 specify the number of edges, rather than being 0.

william-silversmith:
Just to add my two cents here, I've been integrating this format into CloudVolume, the precomputed client, and Igneous, a kubernetes based pipeline:

https://github.com/seung-lab/cloud-volume/wiki/Advanced-Topic:-Skeletons

jbms:
Thanks for your work on getting this started.  There is now support for a precomputed skeleton format on the master branch.

PR #63, opened 2017-11-09
feat(viewer): make default viewer layout configurable

alexbaden:
In ndviz I set the default layout to `xy`. Mostly it's for legacy reasons, but I find it useful for data that is anistropic or not properly downsampled -- improves load times a bit. This change makes the default layer layout configurable so I can override it in the ndviz viewer constructor.


alexbaden:
On second thought, it looks like I can access the `layout` member directly from an extended viewer class. Nevermind!

PR #66, opened 2017-11-28
feat(sliceview): Support per-layer blend functions

alexbaden:
This commit adds a new `blend` property to each layer, allowing the user to change the blend function used on a per-layer basis. Currently, only `additive` blending is supported, but others modes could be added in the `trackable_blend` file. Note that `additive` blending is always enabled, whereas the `default` blend mode is only enabled after the first layer has been added. This mirrors existing behavior.

It might make sense to add `none` as an option, for layers that have blending totally disabled. It would seem like the default should continue to be `gl.blendFunc(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA);` for backwards compatibility.

### DIFF
--- a/src/neuroglancer/trackable_blend.ts
+++ b/src/neuroglancer/trackable_blend.ts
@@ -1,0 +1,45 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {TrackableValue} from 'neuroglancer/trackable_value';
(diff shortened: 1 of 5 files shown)


src/neuroglancer/sliceview/volume/renderlayer.ts
```diff
@@ -433,6 +433,14 @@ export class RenderLayer extends GenericSliceViewRenderLayer {
     return this.gl.memoize.get(key, () => this.buildShader());
   }
 
+  setGLBlendMode(gl: WebGLRenderingContext, renderLayerNum: number) {
```
jbms: If you define this in sliceview/renderlayer.ts, you don't have to duplicate it for both vector_graphics and volume.

src/neuroglancer/trackable_blend.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (verifyEnumString(obj, BLEND_MODES)) {
+    return verifyString(obj);
+  } else {
+    throw new Error();
+  }
+}
+
+export function trackableBlendModeValue(initialValue = 'default') {
+  return new TrackableValue<string>(initialValue, blendModeValidator);
+}
```
jbms: Add linebreak at end of file.

jbms:
Overall this looks good!  It makes sense to preserve the existing special behavior for the first layer until we figure out something better, but it would be nice to do something different.  Also, for annotations like skeletons and vector graphics it probably makes sense in most cases to always draw them last, since you wouldn't want them covered up by an image.

alexbaden:
Should be updated to address your comments.

PR #68, opened 2017-11-30
Sort values during UInt64Set serialization.

timblakely:
UInt64Sets were serialized and sorted according to their hashed value, not their value. This resulted in indeterminate order when serializing state.


jbms:
This is a good idea.  It might be better to sort in numerical order rather than string order, but that can be done later.

PR #69, opened 2017-12-08
Quickfix of example project

Tevemadar:
Webpack of example project seemed to be broken (appaerntly for almost a
year?)
- reference to non-existing variable in webpack config, fixed
- reference to non-exisiting function in webpack config (fixed below)
- reintroduced missing function in webpack helper
- undone oversimplification (anonymous function) so reintroduced
function can work

Disclaimer: blind fixes. no interpretation of changes was done, just
cross-checking with 'config' in project root


jbms:
Thanks for identifying this and creating an initial fix.  I made a similar fix in 80c4854f3935fc18f275da27a2afd705dba6e1df that avoids some of the code duplication.

PR #71, opened 2017-12-18
fix(python): Python3 compatibility for ViewerState

alexbaden:
When getting layers directly in python3 programs, the existence of the `basestring` type causes an error. Replacing with six seems to resolve the problem, but a quick test of existing programs in python2 would be appreciated. My python3 tests are more exhaustive.


jbms:
Thanks for this fix!

Issue #72, opened 2018-01-18
small volumes with render source break


fcollman:
We tried to visualize a small stack that was 512x512x94 in render and ran into an error 
```
renderlayer.ts:53 Uncaught (in promise) TypeError: Cannot read property '0' of undefined
    at new RenderLayer (renderlayer.ts:53)
    at new RenderLayer (renderlayer.ts:400)
    at new ImageRenderLayer (image_renderlayer.ts:52)
    at new ImageRenderLayer (image_renderlayer.ts:43)
    at ImageUserLayer.Object.then.volume (image_user_layer.ts:87)
    at <anonymous>
```
Here was the render stack metadata for the stack we were visualizing..

this stopped breaking when we increased it to 1K x 1K
```
{
  "stackId" : {
    "owner" : "6_ribbon_experiments",
    "project" : "M335503_Ai139_smallvol",
    "stack" : "STEP_1_GFP"
  },
  "state" : "COMPLETE",
  "lastModifiedTimestamp" : "2018-01-18T22:53:04.794Z",
  "currentVersionNumber" : 11,
  "currentVersion" : {
    "createTimestamp" : "2022-04-18T22:52:57.000Z",
    "cycleNumber" : 100,
    "cycleStepNumber" : 1,
    "stackResolutionX" : 1.0,
    "stackResolutionY" : 1.0,
    "stackResolutionZ" : 1.0
  },
  "stats" : {
    "stackBounds" : {
      "minX" : 0.0,
      "minY" : 0.0,
      "minZ" : 0.0,
      "maxX" : 511.0,
      "maxY" : 511.0,
      "maxZ" : 94.0
    },
    "sectionCount" : 95,
    "nonIntegralSectionCount" : 0,
    "tileCount" : 95,
    "transformCount" : 1,
    "minTileWidth" : 511,
    "maxTileWidth" : 511,
    "minTileHeight" : 511,
    "maxTileHeight" : 511
  }
}
```

jbms:
The problem is in datasource/render/frontend.ts --- specifically, computeStackHierarchy decides to use 0 levels if the volume is smaller than the tile size.  It should instead always use at least one level.

PR #74, opened 2018-01-20
fix(datasource/render): Fix issue computing stack hierarchy for const…

alexbaden:
…rained stacks

When the stack bounds are smaller than the render tilesize, computeStackHierarchy would return a stack hierarchy of size 0. Fixed to return hierarchy of size 1 if the stack bounds are smaller than the tile size. Fix google/neuroglancer#72


alexbaden:
Waiting for confirmation from @fcollman, but this should resolve his issue.

fcollman:
I can confirm this is correct for the example i gave alex that broke the data source.

PR #77, opened 2018-02-01
Add offsets to label volume in Python example

jni:
Without this, the example looks weird, with one volume offset from the other.

Let me know if the missing offsets were intentional...

### DIFF
--- a/python/examples/example.py
+++ b/python/examples/example.py
@@ -34,6 +34,7 @@ void main() {
     s.layers.append(
         name='b', layer=neuroglancer.LocalVolume(
             data=b,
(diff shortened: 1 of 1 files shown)


jbms:
For this example, the different offsets were actually intentional, just to demonstrate the use of offsets.  Certainly a comment to clarify that would have made sense, though.

jni:
In addition to a comment, possibly a different-sized volume would also be useful, so that one is not tempted by their OCD to align them. ;) Would a PR doing this be welcome?

jbms:
Sure, that would certainly be welcome.  Thanks!

PR #78, opened 2018-02-22
feat(annotation): Adjustable colors and sizes

nkemnitz:
Lab asked for more control over color and size of all points, and being able to switch between the current orthogonal/fixed size view and a perspective projection for points.
Thought that might be a useful feature for other teams as well.

Visible to user:
* Adjustable color and size for points (only by modifying URL)
* Adjustable default color and size for points (right click on point annotation layer)
* Optional perspective scaling for 2D and/or 3D view. (right click on point annotation layer) Point sizes in this case are interpreted as physical units.

Invisible to user:
* Added `RgbToHex` and `HexToRgb` function in `utils/colorspace.ts`
* fixed unintended behavior (I think?) in `trackable_vec3.ts` -- wanted to verify with @alexbaden 

All changes should be backward compatible. A working demo can be found here: https://goo.gl/WSCPV7

### DIFF
--- a/src/neuroglancer/annotation/frontend.ts
+++ b/src/neuroglancer/annotation/frontend.ts
@@ -14,15 +14,20 @@
  * limitations under the License.
  */
 
+import {AnnotationPointColorList, DEFAULT_COLOR} from 'neuroglancer/annotation/point_color_list';
 import {AnnotationPointList} from 'neuroglancer/annotation/point_list';
+import {AnnotationPointSizeList, DEFAULT_SIZE} from 'neuroglancer/annotation/point_size_list';
 import {ChunkManager} from 'neuroglancer/chunk_manager/frontend';
 import {MouseSelectionState, RenderLayer} from 'neuroglancer/layer';
 import {VoxelSize} from 'neuroglancer/navigation_state';
 import {PerspectiveViewRenderContext, PerspectiveViewRenderLayer} from 'neuroglancer/perspective_view/render_layer';
 import {SliceViewPanelRenderContext, SliceViewPanelRenderLayer} from 'neuroglancer/sliceview/panel';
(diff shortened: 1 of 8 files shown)


jbms:
Thanks for this patch.

I just recently merged a major overhaul of the annotation functionality in Neuroglancer, that replaces the rather primitive point annotation functionality that existed before, and adds support for lines, bounding boxes, and ellipsoids, in addition to points.

See for example this link:

https://goo.gl/VYZ4ZE

There is partial overlap with the functionality in this patch.  For example, in the recently merged annotation support, colors can be set on a per-layer basis, but currently aren't supported for individual annotations.  (Several different colors can be used by using several different layers.)  Points are still displayed at fixed size in pixels on screen, but ellipsoid annotations have a physical size and are rendered accordingly.

It would be helpful to understand your annotation use cases a bit more, to see how we might integrate your changes with the newly merged annotation support.

sdorkenw:
I really like the additions that you made @jbms! 

@nkemnitz asked me to go into some more details how the additional features from his patch are useful to us. I believe there are three main ones:

- node specific colors: This is very helpful to display probabilities of a classifier (say along a skeleton) because you can show different shades between two or three colors of {R, G, B} representing different classes.
- node specific sizes: This allows us to encode an additional feature within the node (eg. synapse size, confidence of some classifier).
- perspective for node sizes: This is useful when displaying many nodes (say all synapses onto a cell). In a zoomed out view one cannot tell the exact location (eg. spine head, shaft) of the synapses anymore. However, choosing the node size small enough to avoid this problem leads to too tiny nodes when zoomed in. Additionally, it also looks nicer for figures. I thought that having this as an option that the user may activate / deactivate is a good compromise.

PR #88, opened 2018-05-20
Bugfix: Make computeVertexNormals normalize the normal vectors correc…

machacekch:
…tly: Float32Array.subarray must be called with [begin, end) indices.

Visually, this fix will make neurons appear slightly brighter overall and remove artifacts where certain vertices were much too dark before.

Optimization: Removed vertexFaceCount: With normal vector normalization working correctly, dividing by the number of adjacent triangles has no effect and can be removed.


jbms:
Thanks for catching this bug!

PR #92, opened 2018-06-24
fix(karma): "extract-text-webpack-plugin" loader is used without the corresponding plugin

nkemnitz:
Something introduced in 9c8a392391452b55b99ce597a7e5b94bef5620a3 causes `npm run test` to fail, complaining about missing plugin `extract-text-webpack-plugin`. Not sure what, or why it worked before...

### DIFF
--- a/config/karma.conf.js
+++ b/config/karma.conf.js
@@ -19,6 +19,7 @@
 let webpackHelpers = require('./webpack_helpers');
 const path = require('path');
 const webpack = require('webpack');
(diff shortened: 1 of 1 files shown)


jbms:
Thanks.  This problem is due to one of the tests transitively depending on a module that requires a css file (a dependency from layer.ts to a css file was introduced).  Previously, I had avoided having any such dependencies, but it is difficult to maintain that separation and it isn't really necessary.

PR #93, opened 2018-06-24
fix(util/uint64): string to uint64 conversion with low32 overflow

nkemnitz:
`tryParseString` fails when the `low` part overflows.


jbms:
Thanks for catching this!

PR #96, opened 2018-07-11
updates to DVID driver

stephenplaza:
All changes were made within the dvid datasource.

* Adds supported for faster, pre-computed JPEG grayscale blocks
* Adds support for new multi-scale segmentation datatype
* Adds support for showing skeletons stored in DVID
* Fixes a bug in the offset calculation
* Removes support for dvid image tiles since its use in DVID is mostly deprecated and a non-ideal fit for neuroglancer

tslint and clang-format were run on the files.

### DIFF
--- a/src/neuroglancer/datasource/dvid/frontend.ts
+++ b/src/neuroglancer/datasource/dvid/frontend.ts
@@ -21,7 +21,8 @@
 
 import {ChunkManager, WithParameters} from 'neuroglancer/chunk_manager/frontend';
 import {CompletionResult, DataSource} from 'neuroglancer/datasource';
-import {DVIDSourceParameters, TileChunkSourceParameters, TileEncoding, VolumeChunkEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/dvid/base';
+import {DVIDSourceParameters, SkeletonSourceParameters, TileChunkSourceParameters, TileEncoding, VolumeChunkEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/dvid/base';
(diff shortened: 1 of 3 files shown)


src/neuroglancer/datasource/dvid/frontend.ts
```diff
@@ -91,34 +111,50 @@ export class VolumeDataInstanceInfo extends DataInstanceInfo {
...
   }
 
   getSources(
       chunkManager: ChunkManager, parameters: DVIDSourceParameters,
       volumeSourceOptions: VolumeSourceOptions) {
     let {encoding} = this;
     let sources: VolumeChunkSource[][] = [];
+
+    // must be 64 block size to work with neuroglancer properly
+    let blocksize = 64;
```
jbms: Can you explain what this blocksize alignment solves?

stephenplaza: Before this modification, there were some circumstances where the lower bound was not coming back as multiple of the chunk size (set to 64).  I think this was occurring because in DVID, the dataset doesn't have to start at 0,0,0 (and can even be negative) which seems to introduce a small off-by-one issue.

jbms: In principle neuroglancer supports arbitrary bounds non-chunk-aligned (or even non-voxel-aligned via the separate "clip" bounds), even negative ones, but perhaps there is some bug somewhere.  In any case I've merged this and we can address this issue later perhaps.

stephenplaza: Thanks Jeremy!  I will do a little sleuthing around to see if there is a bug somewhere deeper in the code.

PR #100, opened 2018-09-03
Moveable annotations

fcollman:
This PR allows the user to move annotations once they have been placed in both 2d slice views and 3d perspective views, presently via a alt+right mouse click drag.

### DIFF
--- a/src/neuroglancer/annotation/bounding_box.ts
+++ b/src/neuroglancer/annotation/bounding_box.ts
@@ -354,7 +354,7 @@ emitAnnotation(vec4(vColor.rgb, uFillOpacity));
     });
   }
 }
-function getEdgeCorners(corners: Float32Array, edgeIndex: number){
+function getEdgeCorners(corners: Float32Array, edgeIndex: number) {
   const i = edgeIndex * 7;
   const cA = vec3.create(), cB = vec3.create();
   for (let j = 0; j < 3; ++j) {
@@ -365,7 +365,7 @@ function getEdgeCorners(corners: Float32Array, edgeIndex: number){
     cB[j] = (1 - mb) * a + mb * b;
   }
 
-  return {cornerA: cA, cornerB: cB}
+  return {cornerA: cA, cornerB: cB};
 }
 function snapPositionToEdge(
     position: vec3, objectToData: mat4, corners: Float32Array, edgeIndex: number) {
@@ -418,7 +418,7 @@ registerAnnotationTypeRenderHandler(AnnotationType.AXIS_ALIGNED_BOUNDING_BOX, {
   getRepresentativePoint: (objectToData, ann, partIndex) => {
     let repPoint= vec3.create();
     // if the full object is selected pick the first corner as representative
-    if (partIndex==FULL_OBJECT_PICK_OFFSET){
+    if (partIndex===FULL_OBJECT_PICK_OFFSET) {
       vec3.transformMat4(repPoint, ann.pointA, objectToData);
     }
     else if (partIndex >= CORNERS_PICK_OFFSET && partIndex < EDGES_PICK_OFFSET) {
@@ -430,24 +430,19 @@ registerAnnotationTypeRenderHandler(AnnotationType.AXIS_ALIGNED_BOUNDING_BOX, {
       vec3.transformMat4(repPoint, ann.pointA, objectToData);
       // snapPositionToCorner(repPoint, objectToData, corners, 5);
     }
-    else{ // for now faces will move the whole object so pick the first corner
+    else { // for now faces will move the whole object so pick the first corner
       vec3.transformMat4(repPoint, ann.pointA, objectToData);
     }
-    return repPoint
-  }, 
+    return repPoint;
+  },
+
   updateViaRepresentativePoint: (oldAnnotation: AxisAlignedBoundingBox, position: vec3, dataToObject: mat4, partIndex: number) => {
     let newPt = vec3.transformMat4(vec3.create(), position, dataToObject);
-    let baseBox = <AxisAlignedBoundingBox>{
-      id: '',
-      type: AnnotationType.AXIS_ALIGNED_BOUNDING_BOX,
-      description: oldAnnotation.description,
-      segments: oldAnnotation.segments,
-      pointA: newPt,
-      pointB: newPt
-    }
+    let baseBox = {...oldAnnotation};
+    baseBox.id = '';
     // if the full object is selected pick the first corner as representative
     let delta = vec3.sub(vec3.create(), oldAnnotation.pointB, oldAnnotation.pointA);
-    if (partIndex==FULL_OBJECT_PICK_OFFSET){
+    if (partIndex===FULL_OBJECT_PICK_OFFSET) {
       baseBox.pointA = newPt;
       baseBox.pointB = vec3.add(vec3.create(), newPt, delta);
     }
@@ -459,7 +454,7 @@ registerAnnotationTypeRenderHandler(AnnotationType.AXIS_ALIGNED_BOUNDING_BOX, {
       baseBox.pointA = newPt;
       baseBox.pointB = vec3.add(vec3.create(), newPt, delta);
     }
-    else{ // for now faces will move the whole object so pick the first corner
+    else { // for now faces will move the whole object so pick the first corner
       baseBox.pointA = newPt;
       baseBox.pointB = vec3.add(vec3.create(), newPt, delta);
     }
--- a/src/neuroglancer/annotation/bounding_box.ts
+++ b/src/neuroglancer/annotation/bounding_box.ts
@@ -354,7 +354,7 @@ emitAnnotation(vec4(vColor.rgb, uFillOpacity));
     });
   }
 }
-function getEdgeCorners(objectToData: mat4, corners: Float32Array, edgeIndex: number){
+function getEdgeCorners(corners: Float32Array, edgeIndex: number){
   const i = edgeIndex * 7;
   const cA = vec3.create(), cB = vec3.create();
   for (let j = 0; j < 3; ++j) {
@@ -364,13 +364,15 @@ function getEdgeCorners(objectToData: mat4, corners: Float32Array, edgeIndex: nu
     cA[j] = (1 - ma) * a + ma * b;
     cB[j] = (1 - mb) * a + mb * b;
   }
-  vec3.transformMat4(cA, cA, objectToData);
-  vec3.transformMat4(cB, cB, objectToData);
+
   return {cornerA: cA, cornerB: cB}
 }
 function snapPositionToEdge(
     position: vec3, objectToData: mat4, corners: Float32Array, edgeIndex: number) {
-  let edgeCorners = getEdgeCorners(objectToData, corners, edgeIndex);
+  let edgeCorners = getEdgeCorners(corners, edgeIndex);
+  vec3.transformMat4(edgeCorners.cornerA, edgeCorners.cornerA, objectToData);
+  vec3.transformMat4(edgeCorners.cornerB, edgeCorners.cornerB, objectToData);
+
   projectPointToLineSegment(position, edgeCorners.cornerA, edgeCorners.cornerB, position);
 }
 
@@ -413,23 +415,28 @@ registerAnnotationTypeRenderHandler(AnnotationType.AXIS_ALIGNED_BOUNDING_BOX, {
       // vec3.transformMat4(position, annotation.point, objectToData);
     }
   },
-  getRepresentativePoint: (objectToData, data, offset, partIndex) => {
+  getRepresentativePoint: (objectToData, data, offset, partIndex, ann) => {
     let repPoint= vec3.create();
     const corners = new Float32Array(data, offset, 6);
+    console.log(partIndex)
     // if the full object is selected pick the first corner as representative
     if (partIndex==FULL_OBJECT_PICK_OFFSET){
-      snapPositionToCorner(repPoint, objectToData, corners, 0);
+      vec3.transformMat4(repPoint, ann.pointA, objectToData);
     }
     else if (partIndex >= CORNERS_PICK_OFFSET && partIndex < EDGES_PICK_OFFSET) {
       // picked a corner
       snapPositionToCorner(repPoint, objectToData, corners, partIndex);
     } else if (partIndex >= EDGES_PICK_OFFSET && partIndex < FACES_PICK_OFFSET) {
       // for edges we will pick cornerA to move
-      let edgeCorners = getEdgeCorners(objectToData, corners, partIndex - EDGES_PICK_OFFSET); 
-      vec3.transformMat4(repPoint, edgeCorners.cornerA, objectToData);
+      // let edgeCorners = getEdgeCorners(corners, partIndex - EDGES_PICK_OFFSET);
+      // }
+      //
+      // FIXME: can't figure out how to resize based upon edge grabbed
+      vec3.transformMat4(repPoint, ann.pointA, objectToData);
+      // snapPositionToCorner(repPoint, objectToData, corners, 5);
     }
     else{ // for now faces will move the whole object so pick the first corner
-      snapPositionToCorner(repPoint, objectToData, corners, 0);
+      vec3.transformMat4(repPoint, ann.pointA, objectToData);
     }
     return repPoint
   }, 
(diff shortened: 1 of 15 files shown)


nkemnitz:
Playing around with the controls in Neuroglancer, it would feel more natural to me to use Ctrl+LMB again to "pick up" an existing annotation... The RMB I would mostly (try to) use to delete a point, or abort a modification. Not sure... why did you end up with Alt+RMB?

fcollman:
ctrl+LMB will add another annotation I think on mousedown, there are potentially use cases where people want to add annotations while on top of other annotations.  I'm not sure alt+RMB is the best choice, i'm definitely open to alternatives.. maybe alt+LMB is better (alt for alter in this case?)

src/neuroglancer/perspective_view/panel.ts
```diff
@@ -225,6 +226,39 @@ export class PerspectivePanel extends RenderedDataPanel {
...
+          let point = handler.getClosestPoint(ann, voxelSize, mouseState.position)
+
+          if (mouseState.updateUnconditionally()) {
+            startRelativeMouseDrag(e.detail, (_event, deltaX, deltaY) => {
+              const temp = tempVec3;
+              const {projectionMat} = this;
+              const {width, height} = this;
+      
+              voxelSize.spatialFromVoxel(point,point)
+              vec3.transformMat4(temp, point, projectionMat);
```
jbms: The annotations are intended to be immutable, so instead of modifying point here, instead create a new annotation with the point modified, and call annotationLayer.source.update(annotationRef, newAnnotation);

Finally, once the drag ends, call annotationLayer.source.commit(annotationRef);  and annotationRef.dispose(); to avoid leaking a reference.

Unfortunately, startRelativeMouseDrag doesn't currently provide a way to specifying a callback to be invoked when the drag ends (needed into order to call commit), but that would be easy to add as an optional additional argument.

For the in-memory annotations, just updating in place and calling source.changed.dispatch works, as you've noticed, but for annotations stored on a server, the update/commit calls are needed.

src/neuroglancer/perspective_view/panel.ts
```diff
@@ -225,6 +226,39 @@ export class PerspectivePanel extends RenderedDataPanel {
...
+
+      let voxelSize = this.viewer.navigationState.voxelSize
+      if (typeof(annotationLayer) != 'undefined'){
+        if (typeof(selectedAnnotationId) != 'undefined'){
+          
+          let annotationRef = annotationLayer.source.getReference(selectedAnnotationId)!;
+          let ann = <Annotation>annotationRef.value
+          
+          const handler = getAnnotationTypeHandler(ann.type)
+          let point = handler.getClosestPoint(ann, voxelSize, mouseState.position)
```
jbms: For bounding boxes, it would be great if it were possible to drag any of the 8 corners.  However, admittedly that would add some complexity.

src/neuroglancer/sliceview/panel.ts
```diff
@@ -140,6 +143,35 @@ export class SliceViewPanel extends RenderedDataPanel {
       }
     });
 
+    registerActionListener(element, 'translate-annoation-via-mouse-drag', (e: ActionEvent<MouseEvent>) => {
+      const {mouseState} = this.viewer;
+      const selectedAnnotationId = mouseState.pickedAnnotationId;
+      const annotationLayer = mouseState.pickedAnnotationLayer;
+
+      let voxelSize = this.viewer.navigationState.voxelSize
```
jbms: Please add a semicolon to the end of all lines.

src/neuroglancer/ui/default_input_event_bindings.ts
```diff
@@ -87,6 +87,7 @@ export function getDefaultPerspectivePanelBindings() {
         {
           'at:mousedown0': {action: 'rotate-via-mouse-drag', stopPropagation: true},
           'at:shift+mousedown0': {action: 'translate-via-mouse-drag', stopPropagation: true},
+          'at:alt+mousedown2': {action: 'translate-annoation-via-mouse-drag', stopPropagation: true}
```
jbms: Since it is the same for both perspective and slice view, you can instead specify this binding in the getDefaultRenderedDataPanelBindings() function.

src/neuroglancer/perspective_view/panel.ts
fcollman: Thanks for the feedback.  Seems all straightforward andI have most done already, except I'm a bit confused about how to move forward with the pickedOffset.  

We'd have to translate each of these pickedOffsets to a desired behavior for each annotation.. i.e. clicking on the line should translate both endpoints for the Line annotation, but clicking on a endpoint should only move one, so maybe we'd have to have an interface for each annotation type which would give a point for an annotation and its pickedOffset, so the handler put into the Drag callback would move that point, and then another interface which would update the annotation's internal representation given a new point and the same pickedOffset?  

something like this (for sliceView?) (with getRepresentativePoint and updateViaRepresentativePoint to be implemented by each AnnotationType).
``` typescript
          let annotationRef = annotationLayer.source.getReference(selectedAnnotationId)!;
          let ann = <Annotation>annotationRef.value;
          let newAnnotation = {...ann};
          annotationLayer.source.update(annotationRef, newAnnotation);
          const handler = getAnnotationTypeRenderHandler(newAnnotation.type);
          let repPoint = handler.getRepresentativePoint(ann, mouseState.pickedOffset);
          if (mouseState.updateUnconditionally()) {
            startRelativeMouseDrag(e.detail, (_event, deltaX, deltaY) => {
              
              vec3.transformMat4(repPoint, repPoint, this.sliceView.dataToViewport);
              vec3.set(repPoint, repPoint[0]-deltaX, repPoint[1]-deltaY, point[2]);
              vec3.transformMat4(repPoint, repPoint, this.sliceView.viewportToData);
              handler.updateViaRepresentativePoint(ann, mouseState.pickedOffset, repPoint)
              annotationLayer.source.changed.dispatch();
            },
            (_event) => {
              annotationRef.value = newAnnotation;
              annotationLayer.source.commit(annotationRef);
              annotationRef.dispose();
            });
         }
```

Another issue I encountered when exploring this... 
Seems like for a Line it returns 0  for the line, 1 for pointA and 2 for pointB, but for BoundingBox it only returns 0 (the FULL_OBJECT_PICK_OFFSET) no matter where I click on the box in the sliceView (though I get different indices for each line of the box in perspectiveView).  I don't think the corner's are selectable in any of the views.. probably because they don't have shaders... but that's maybe not a problem. Do you know why and how I might fix that?

fcollman: oh and this assumes of course that I have the new optional finishMoveHandler option on startRelativeMouseDrag

```typescript
export function startRelativeMouseDrag(initialEvent: MouseEvent,
                                       handler: RelativeDragHandler,
                                       finishDragHandler?: FinishDragHandler ) {
  let {document} = initialEvent.view;
  let prevScreenX = initialEvent.screenX, prevScreenY = initialEvent.screenY;
  let mouseMoveHandler = (e: MouseEvent) => {
    let deltaX = prevScreenX - e.screenX;
    let deltaY = prevScreenY - e.screenY;
    prevScreenX = e.screenX;
    prevScreenY = e.screenY;
    handler(e, deltaX, deltaY);
  };
  let button = initialEvent.button;
  let mouseUpHandler = (e: MouseEvent) => {
    if (e.button === button) {
      document.removeEventListener('mousemove', mouseMoveHandler, true);
      document.removeEventListener('mouseup', mouseUpHandler, false);
      if (typeof(finishDragHandler) != 'undefined'){
        finishDragHandler(e);
      }
    }
  };
  document.addEventListener('mousemove', mouseMoveHandler, true);
  document.addEventListener('mouseup', mouseUpHandler, false);
}
```

fcollman:
okay.. i have made another attempt.  There are a view problems.  i couldn't sort out how to make the edge moving for the bounding box work... I think I have all the infrastructure in place to make the right choices, but moving between the buffer representation and the pointA/pointB representation is tripping me up.  For now i've just made it only translate the boxes in slice and perspective view.  

Another issue the remains that existed in the original version is that the mouseDrag isn't tightly locked to where the cursor is located, so when people try to drag a point to a new location the point doesn't end up where the cursor is.  It isn't terrible because you can see where the point will go, but it would be better if this would be fixed.  It is also present in the panning functionality right now, but there I think its less of a problem as people aren't usually doing precise panning, and if they want to do that they can simply right click, which works perfectly.  I tried to accumulate the pixel deltas in a variable, rather than processing them incrementally in hopes that would solve some rounding/conversion accumulation bug, but that wasn't it.

jbms:
Thanks for these changes.  I think it is reasonable to just translate bounding boxes for now.

Regarding the mouseDrag not being tightly locked to where the cursor is, I was not aware of that issue with panning.  I'm not able to reproduce it on my machines, which makes it trickier for me to debug.  Which operating system and browser do you see this on?  Can you see if it is a fixed multiplier?  I suspect it is due to some sort of scale mismatch.

Some things you might try investigating:
window.devicePixelRatio

The moue dragging is based on the screenX and screenY fields of the MouseEvent, while the "right click" positioning is based on the offsetX and offsetY fields.  You could check to see if those fields in fact have different scaling.

fcollman:
I switched it to the double click method you suggested,  I think its much easier to use and more intuitive.  I stoppedPropogation() of the event if there is an annotation selected, assuming that people want to interact with the annotation if they have hovered over it.

jbms:
Thanks for your information about the devicePixelRatio.  I see that changing the normal browser zoom (not neuroglancer's zoom) also affects devicePixelRatio, and currently leads to incorrect panning behavior with Neuroglancer.  It appears that screenX and screenY behave somewhat inconsistently in the zoom vs HiDPI display scenarios, and also are inconsistent across browsers to some extent.  Replacing the uses of screenX and screenY with clientX and clientY in startMouseRelativeDrag at least makes panning work correctly in the presence of browser zoom.  Can you confirm if that also fixes the issue you are seeing?

fcollman:
Yes, I can confirm that on my macbook that clientX does fix the issue, and
works with zooming (assuming you enter the page with that zoom.. it seems
to still fail if you change zoom while on the page).

src/neuroglancer/util/mouse_drag.ts
```diff
@@ -15,21 +15,25 @@
  */
 
 export type RelativeDragHandler = (event: MouseEvent, deltaX: number, deltaY: number) => void;
-export function startRelativeMouseDrag(initialEvent: MouseEvent, handler: RelativeDragHandler) {
+export type FinishDragHandler = (event: MouseEvent) => void;
```
jbms: We can use RelativeDragHandler as the type for the finishDragHandler as well, and supply the offsets as well (even if they usually won't be used).

src/neuroglancer/annotation/type_handler.ts
```diff
@@ -163,6 +163,8 @@ interface AnnotationTypeRenderHandler<T extends Annotation> {
   };
   sliceViewRenderHelper: {new(gl: GL): AnnotationRenderHelper;};
   pickIdsPerInstance: number;
+  getRepresentativePoint: (objectToData: mat4, data: ArrayBuffer, offset: number, partIndex: number, annotation: T) => vec3;
```
jbms: Do we need the data and offset parameters?  All of the information should be present in the annotation object.

src/neuroglancer/perspective_view/panel.ts
```diff
@@ -229,6 +231,56 @@ export class PerspectivePanel extends RenderedDataPanel {
       });
     });
 
+    registerActionListener(element, 'select', (e: ActionEvent<MouseEvent>) => {
```
jbms: Can we avoid the code duplication here, but factoring out common logic?

I'm actually not that keen on overriding the meaning of the select command here.  Additionally, startRelativeMouseDrag was really intended to be used only on mousedown events.  It happens to work, but it leaves room for weird behavior, like both moving an annotation and placing another annotation at the same time.

I think it would be better to revert to the original behavior of using alt+mousedown0 for now.  I think it would be better that alt+mousedown0 starts moving, and alt+mousedown0 again stops moving, and the two clicks need not even be in the same panel.  However, implementing that would make this change more complicated, so it could wait until later.

fcollman: I think I could use a little more guidance to address this.  I've switched to the alt+mouse0down functionality, so you have are mouseDragging again, rather than picking up and putting down. 

However, my attempts of reducing code duplication have all been rather ugly.  There are certaintly many duplicate lines of code, but with subtle different lines interspersed, and I'm having trouble finding the right abstraction to remove the duplicate part replication without having a lot of ugly options.

jbms:
Regarding refactoring to avoid code duplication between the two move-annotation action listeners, it looks to me like the only difference is in how newRepPt is computed.  You could, therefore, define an abstract method in RenderedPanel:

  abstract translateDataPointByViewportPixels(out: vec3, orig: vec3, deltaX: number, deltaY: number): void;

and then define move-annotation within rendered_data_panel.ts and define that method for both slice view and perspective view.

src/neuroglancer/annotation/bounding_box.ts
```diff
@@ -410,4 +415,54 @@ registerAnnotationTypeRenderHandler(AnnotationType.AXIS_ALIGNED_BOUNDING_BOX, {
...
+      vec3.transformMat4(repPoint, ann.pointA, objectToData);
+      // snapPositionToCorner(repPoint, objectToData, corners, 5);
+    }
+    else{ // for now faces will move the whole object so pick the first corner
+      vec3.transformMat4(repPoint, ann.pointA, objectToData);
+    }
+    return repPoint
+  }, 
+  updateViaRepresentativePoint: (oldAnnotation: AxisAlignedBoundingBox, position: vec3, dataToObject: mat4, partIndex: number) => {
+    let newPt = vec3.transformMat4(vec3.create(), position, dataToObject);
```
jbms: You can instead use the object spread notation let baseBox = <AxisAlignedBoundingBox>{...oldAnnotation};

Same thing applies to other annotation types.

jbms: In addition to being shorter, this also ensures that we copy all of the fields.  Is there a reason why you are setting the id to the empty string?

src/neuroglancer/util/mouse_drag.ts
```diff
@@ -15,21 +15,25 @@
...
+    prevScreenY = e.clientY;
     handler(e, deltaX, deltaY);
   };
   let button = initialEvent.button;
   let mouseUpHandler = (e: MouseEvent) => {
     if (e.button === button) {
       document.removeEventListener('mousemove', mouseMoveHandler, true);
       document.removeEventListener('mouseup', mouseUpHandler, false);
+
+      if (typeof(finishDragHandler) != 'undefined') finishDragHandler(e, prevScreenX - e.clientX, prevScreenY - e.clientY);
```
jbms: You can check for undefined here and elsewhere with the shorter syntax finishDragHandler !== undefined.  Note the triple equals.

jbms:
Thanks for all of your work on this, it is almost ready to merge.

In addition to the few comments I made regarding specific changed lines, can you run clang-format on the changed parts, and fix tslint errors (mainly lack of semicolons)?

src/neuroglancer/annotation/bounding_box.ts
fcollman: I thought you had said that the annotations should be atomic, so I thought the new annotation should get a new ID, and I thought that ID got assigned when i added the new annotation to the layer.  Is my thinking off?

jbms: The id should stay the same.  One of the purposes of the id is to allow an annotation to be uniquely identified even as it is modified.

fcollman:
thank you for your patience as I work through my typescript and neuroglancer on-boarding process.   I have changed the style, but if i don't give it a blank ID and force the add function to construct a new ID the code doesn't work.  I suspect this has to do with some collision/timing of the disposing of the old annotation relative to the creation of the new one, but i'm still investigating.  Also, i'm getting some irreproducible behavior with whether the new points end up being updated on the python server side, which makes me think that i'm also not doing the add/dispose/etc/ in the right way with the annotation references.

fcollman:
that latest commit I think fixes the problem of the inconsistent saving of the state.  It seems terribly inefficent to do a full update of state on every deltaX, deltaY move of the cursor though.  And I still haven't gotten it to work with keeping the ID the same.

fcollman:
This latest commit solves the problem through using update.  Annotation's are still atomic, but the AnnotationRef is now updated, rather than the old one removed and the new one added.  Is this ok?

PR #16, opened 2017-02-10
refractor: Using WebGL 2 and GLSL 300.

tartavull:
### WebGL2
* Added WebGL2RenderingContext typescript files
* Switched extensions (draw_buffers & instanced arrays) with now standard functions

### GLSL 300
* `attribute` --> `in`
* `varying` --> `out` in vertex shader, `in` in fragment shader
* `texture2D` --> `texture`
* added basic function to add fragment shader outputs (no validation checks!)
* replaced `gl_FragColor` with `v4f_fragColor`
* replaced `gl_FragData[X]` with `v4f_fragDataX`

Author:
https://github.com/nkemnitz

PR #101, opened 2018-09-03
feat(webgl): Switch to WebGL2 and GLSL 300 ES

nkemnitz:
Supersedes https://github.com/google/neuroglancer/pull/16

What this PR does:
* Switching over from WebGLRenderingContext to WebGL**2**RenderingContext
* Upgrade shader code to version GLSL 300 ES

What this PR **not** does:
* Use any of the new cool features of WebGL/GLSL other than required for maintaining current Neuroglancer behavior and passing the tests.
* try to be backward compatible: No IE/Edge support [in the foreseeable future](https://caniuse.com/#search=WebGL)

WebGL changes in more detail:
* Using the DefinitelyTyped definitions for WebGL2
* `WEBGL_draw_buffers`, `OES_texture_float`, `OES_element_index_uint` are standard in WebGL2 - no need to check
* WebGL2 offers many new combinations of `internalFormat`, `format` and `type` for textures, but doesn't support *unsized* internal formats when using `GL_FLOAT` types --> had to separate `internalFormat` and `format`.

GLSL changes in more detail:
* `#version 300 es` string at beginning of shader
* `attribute`  in VS --> `in`
* `varying` in VS --> `out`
* `varying` in FS --> `in`
* No predefined `gl_FragColor` or `gl_FragData[x]` outputs anymore; user has to specify outputs using new function `addBufferOutput(typename, name, location)`. I used `v4f_fragColor`/`v4f_fragData[x]` to replicate old behavior.
* `texture2D` --> `texture`
* Overloading built-in functions does not seem to work anymore; had to rename `equal(uint32_t a, uint32_t b)` to `equals`.

One thing:
` neuroglancer/webgl/constants` still only contains the WebGL (1.0) constants - not sure if I should add the 2.0 ones or if I can remove it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ or within Neuroglancer, press `h` or click on the button labeled `?` in the uppe
 
 - Neuroglancer doesn't appear to load properly.
 
-  Neuroglancer requires WebGL (1.0) and the `WEBGL_draw_buffers`, `OES_texture_float`, and `OES_element_index_uint` extensions.
+  Neuroglancer requires WebGL (2.0) and the `EXT_color_buffer_float` extension.
   
   To troubleshoot, check the developer console, which is accessed by the keyboard shortcut `control-shift-i` in Firefox and Chrome.  If there is a message regarding failure to initialize WebGL, you can take the following steps:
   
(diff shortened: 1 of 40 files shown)


jbms:
This is great!  Thanks for revising this, and making sure all the tests pass!

The original reason for the constants was to allow their use from functions that did not otherwise require an instance of WebGLRenderingContext.  However, I see from your changes that they can in fact be accessed directly on the WebGLRenderingContext global; I wasn't aware that was supported.  Therefore I agree it makes sense to eliminate constants.ts.

PR #106, opened 2018-09-19
adds dvid simple key/value mesh support

stephenplaza:
This commit adds support for meshes with the DVID backend using the pre-computed mesh format specified by neuroglancer.  The current implementation assumes a single chunk representation for the mesh as the intended use case, for now, is for quick (low-res if needed) browsing.  I otherwise tried to closely conform to the interface used in the precomputed instance in case we want to support mesh chunking in the future.

tslint and clang-format were run on the files.

### DIFF
--- a/src/neuroglancer/datasource/dvid/backend.ts
+++ b/src/neuroglancer/datasource/dvid/backend.ts
@@ -15,13 +15,15 @@
  */
 
 import {WithParameters} from 'neuroglancer/chunk_manager/backend';
-import {SkeletonSourceParameters, VolumeChunkEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/dvid/base';
+import {MeshSourceParameters, SkeletonSourceParameters, VolumeChunkEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/dvid/base';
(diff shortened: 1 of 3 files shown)


src/neuroglancer/datasource/dvid/backend.ts
```diff
@@ -42,6 +44,38 @@ import {registerSharedObject} from 'neuroglancer/worker_rpc';
...
+(WithParameters(MeshSource, MeshSourceParameters)) {
+  download(chunk: ManifestChunk) {
+    // DVID does not currently store meshes chunked, the main
+    // use-case is for low-resolution 3D views.
+    // for now, fragmentId is the body id
+    return new Promise(function(resolve) {
+             resolve();
+           })
+        .then(() => {
+          chunk.fragmentIds = [`${chunk.objectId}`];
```
jbms: You can simplify this code by just setting chunk.fragmentIds = ... and returning Promise.resolve(undefined);  no need for new Promise.

stephenplaza: I was wondering the most idiomatic way to do this.  Thanks for the feedback.  I will post a change shortly.

PR #107, opened 2018-09-20
[feature] Json link shortener

fcollman:
This adds a link shortener functionality to neuroglancer by implementing two things.  First, if the query_parameter json_url is passed to the neuroglancer page, it will go attempt to get a json from the specified URL, and use that to set the state, and remove that query parameter from the url.

Second, a button is added to the upper right that when pressed, posts the present viewer state to a url specified by a configurable jsonStateServer parameter in the json state.  The resulting response from that url should contain a URL where the json can be later retrieved.  For now there is a dialog that prompts the user to fill in the name of such a service, of which https://api.myjson.com/ is a free and available prototype, but others could be implemented.

### DIFF
--- a/src/main_python.ts
+++ b/src/main_python.ts
@@ -119,7 +119,7 @@ window.addEventListener('DOMContentLoaded', () => {
   configState.add('screenshot', screenshotHandler.requestState);
 
   let sharedState: Trackable|undefined = viewer.state;
-
+  viewer.loadFromJsonUrl();
   if (window.location.hash) {
     const hashBinding = viewer.registerDisposer(new UrlHashBinding(viewer.state));
     hashBinding.updateFromUrlHash();
@@ -177,6 +177,4 @@ window.addEventListener('DOMContentLoaded', () => {
 
   bindDefaultCopyHandler(viewer);
   bindDefaultPasteHandler(viewer);
-
-
 });
--- a/src/main_python.ts
+++ b/src/main_python.ts
@@ -177,4 +177,6 @@ window.addEventListener('DOMContentLoaded', () => {
 
   bindDefaultCopyHandler(viewer);
   bindDefaultPasteHandler(viewer);
+
+
 });
--- a/src/neuroglancer/ui/default_viewer_setup.ts
+++ b/src/neuroglancer/ui/default_viewer_setup.ts
@@ -38,6 +38,8 @@ export function setupDefaultViewer() {
     hashBinding.parseError;
   }));
   hashBinding.updateFromUrlHash();
+  
+  viewer.loadFromJsonUrl();
 
   bindDefaultCopyHandler(viewer);
   bindDefaultPasteHandler(viewer);
--- a/src/neuroglancer/ui/url_hash_binding.ts
+++ b/src/neuroglancer/ui/url_hash_binding.ts
@@ -33,6 +33,10 @@ function encodeFragment(fragment: string) {
   });
 }
 
+export function removeParameterFromUrl(url: string, parameter: string) {
+  return url.replace(new RegExp('[?&]' + parameter + '=[^&#]*(#.*)?$'), '$1')
+      .replace(new RegExp('([?&])' + parameter + '=[^&]*&'), '$1');
+}
 /**
  * An instance of this class manages a binding between a Trackable value and the URL hash state.
  * The binding is initialized in the constructor, and is removed when dispose is called.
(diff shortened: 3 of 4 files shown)


falkben:
Confirmed!

jbms:
One way this could be simplified is to have it work more like a normal URL shortener --- you would simply post the full URL to the shortener, which would generate a short URL that would be returned and automatically copied to the clipboard.  The short URL would then simply redirect, like a normal URL shortener, back to neuroglancer.  This would eliminate the need for any logic within neuroglancer itself for unshortening.  The main downside is that it would necessarily redirect to a specific neuroglancer client, rather than being usable with any neuroglancer client.  You would have to manually change the host afterwards.  Is that likely to be an issue for your use case?

This same logic would actually work perfectly well as a URL shortener browser extension that wouldn't depend on neuroglancer at all.  However, I'm not sure if there are any that can handle extremely long URLs, and I suppose many people may have no need for a URL shortener outside of their use of neuroglancer.

Another way this could be simplified is to specify the URL shortener service as a build-time parameter, rather than allowing it to be specified as part of the state.  That would simplify the code quite a bit, and might also make it simpler to integrate with a URL shortener that requires user authentication for creating new links.  I'm also a bit concerned that allowing it to be specified as part of the state may be confusing, since it would be hard to know which URL shortener service will be used.

I'm also not sure how feasible it is to rely on an anonymous URL shortener service for this purpose.  The length of the URLs can be quite long, such that the total amount of data that the service has to store could be substantial.  Additionally, as a user, you would like to have confidence that the links will remain valid essentially forever.

jbms:
I think I agree with you regarding not relying on an external URL shortener, or even URLs at all.

One thing I've been looking into is using the Google Drive API as the storage mechanism.  That has the advantage of being able to make use of the free 15GB of storage provided to every Google account, while still having the storage "owned" by a particular account and therefore avoids the problem of storing an arbitrarily large amount of neuroglancer state data for anonymous users forever.  It also provides some nice functionality like thumbnail images (which can contain screenshots captured at the same time as the view state).  I've been experimenting with a prototype of this, that mostly works, but there are some issues regarding sharing, in that there doesn't seem to be a way to provide a way to share the drive file with another user and provide them with a URL that directly opens neuroglancer, especially if they haven't already "installed" neuroglancer as a Google Drive app.

jovo:
@jbms hey, any update on this, we are writing a manuscript on our cloud stuff now, and currently mentioning this functionality as existing in our fork (which is semi-up-to-date).  So, if you are still planning on merging it in, and waiting on something from us, maybe now is a good time to finalize this?

jbms:
Part of this (the easy part) is already merged in, in that you can just put a URL to the JSON state directly in the Neuroglancer URL, for example:  https://hemibrain-dot-neuroglancer-demo.appspot.com/#!gs://neuroglancer-janelia-flyem-hemibrain/v1.0/neuroglancer_demo_states/base.json

I didn't continue work on the Google Drive integration because of the limitations I ran into.

As far as this pull request, I do agree that the basic approach, a server for storing the JSON state with specific integration into Neuroglancer, rather than relying on a generic URL shortener, makes sense.

I'm still a bit concerned about relying on a JSON "POST" API where there is no authentication, presumably allows fairly large data sizes (e.g. tens of MB), and intends to retain the data indefinitely.  However, I can also imagine the chance of abuse may be low and there would likely be no problems.

jovo:
I would be pretty surprised if people abused it, but maybe I am naive.  @fcollman @falkben what do you guys think?

falkben:
I'll just preface by just saying that since I'm no longer working on this I probably shouldn't be a point person for this anymore and my opinion shouldn't carry any weight.

I wasn't aware that neuroglancer had the ability to load from a url now, so that's very nice. I do think that having an interface for users to store state is still valuable. Is that present or something that could be added outside of this PR?

I do share a concern about security here. I cannot vouch for the free service we defaulted to https://api.myjson.com/. Neurodata has their own json store service I set up which is open to outside use. So perhaps could switch to that for a default. But I'm not sure that solves the problem.

There's a minor issue of people abusing it but I think a bigger issue is possibly just trusting any url shortener since this does get executed by the browser and we didn't add any protections around that. We also store things no encryption which should be something to tell users about as well.

Perhaps removing a default and letting the user pick would be adequate? Or a warning?

fcollman:
There is link shortening now available as a build time option so this is out of date.

PR #109, opened 2018-10-02
feat(annotations): Delete annotation per mouse click

nkemnitz:
People in our lab mentioned that the old way of removing an annotation (`Ctrl+RMB`) now only selects the annotation, and an additional click on the paper bin symbol in the annotation panel is required to delete the annotation.
This PR adds `Alt+RMB` to restore the old functionality for quickly removing single annotations. `Alt+RMB` was chosen because the other combinations are currently taken... Not sure if we can make this prettier / more symmetric / intuitiv.

`Ctrl+LMB`: Add Annotation
`Ctrl+RMB`: Select Annotation
`Alt+LMB`: Move Annotation
`Alt+RMB`: Delete Annotation **[NEW]**

Closes https://github.com/seung-lab/neuroglancer/issues/218


nkemnitz:
Oh, hm... looks like `Alt+RMB` is already taken by some setups to force show the context menu. Have to investigate a bit more.

PR #112, opened 2018-10-19
feat: add configurable cursor size

william-silversmith:
The current neuroglancer cursor is only sensitive around its central pixel. However, when interacting with thin objects, a few more pixels can help. This PR works by increasing the receptive field of the cursor from 1px to 23px in diameter.

This is done by creating a spiral search pattern out from the central pixel towards the extremities. Thus, the central pixel will always be selected first, but anything up to radius 11px away will be selected if it is within range and nothing else is in the way. 

We have tested this and it appears to work without imposing significant CPU costs.


jbms:
This is a great idea, thanks!  I didn't merge your exact implementation, but I implemented something very similar based on it in 4ac770f2d3e651c05a71979000e4d02f19359835.

PR #113, opened 2018-11-27
fix(brainmaps): Optimize listing volumes

larrylindsey:


jbms:
Thanks, this is great!

PR #114, opened 2018-11-29
Feat(datasource/precomputed): Support skeletons

falkben:
This adds support for precomputed skeletons compatible with the neuroglancer python format.  

This is really only a small continuation of the original PR that @alexbaden made here: https://github.com/google/neuroglancer/pull/60

There is currently a constraint in that within the precomputed `info` file you can only have either a mesh or a skeleton folder, not both.  I think this is as intended, as otherwise the segments IDs would be linked and that doesn't make sense.  To get meshes AND skeletons in the same view we load two separate info files as separate layers.


alexbaden:
Thanks for picking this up. I am not as familiar with the base codebase as I used to be, but this looks good to me. If there’s anything else needed from me please let me know!

jbms:
Thanks for this contribution, and I apologize for the long delay.  This functionality is now available on the master branch, with the only difference that an info JSON file is required to specify the metadata for the skeleton data.

PR #115, opened 2018-12-04
add two more storage backends

jingpengw:


jingpengw:
I signed it!

jbms:
Thanks, these are good links to include, but please add them under the "Related Projects" section rather than the "Supported data sources" section.  Alternatively, or in addition, you could link them under the documentation for the "Precomputed data source".

jingpengw:
just moved to "Related Projects" section.

PR #119, opened 2019-01-18
removes ndstore & openconnectome

falkben:
This removes ndstore and openconnectome datasources.  These two services have been offline and will not be reinstated.


src/neuroglancer/datasource/python/backend.ts
```diff
@@ -21,7 +21,6 @@ import {decodeSkeletonVertexPositionsAndIndices, SkeletonChunk, SkeletonSource}
 import {VertexAttributeInfo} from 'neuroglancer/skeleton/base';
 import {ChunkDecoder} from 'neuroglancer/sliceview/backend_chunk_decoders';
 import {decodeJpegChunk} from 'neuroglancer/sliceview/backend_chunk_decoders/jpeg';
-import {decodeNdstoreNpzChunk} from 'neuroglancer/sliceview/backend_chunk_decoders/ndstoreNpz';
```
jbms: This format support shouldn't be removed here.

src/neuroglancer/sliceview/backend_chunk_decoders/bossNpz.ts
```diff
@@ -15,7 +15,7 @@
  */
```
jbms: There isn't really a need to rename this file.

jbms:
Hi,

I apologize for the delay.  Removing these old sources makes sense.  I've just added a few comments regarding the ndstoreNpz.ts file.

src/neuroglancer/sliceview/backend_chunk_decoders/bossNpz.ts
falkben: the spec is slightly different between ndstore and boss with the file duplication occurring on the initial BOSS feature add.  sSems easier at this point to just delete the ndstore file?

jbms: Oh, good point.  I didn't realize that.  The Python backend supports multiple channels and therefore needs the ndstore version, though, and they are trivial enough that there may not be a point in trying to combine them.  Perhaps just leave both versions.

PR #120, opened 2019-01-18
Allow user to limit data downloads to specific MIP levels

manuel-castro:
This PR adds two dropdown menus into the Rendering tab for a selected image/segmentation layer; these dropdowns allow to user to limit the data that will be downloaded and displayed for the selected layer by specifying a min and a max voxel size/MIP level.

### DIFF
--- a/src/neuroglancer/image_user_layer.ts
+++ b/src/neuroglancer/image_user_layer.ts
@@ -67,7 +67,9 @@ export class ImageUserLayer extends Base {
           fragmentMain: this.fragmentMain,
           shaderError: this.shaderError,
           transform: this.transform,
+          mipLevelConstraints: this.mipLevelConstraints
         });
+        this.setupVoxelSelectionWidget(renderLayer);
         this.addRenderLayer(renderLayer);
         this.shaderError.changed.dispatch();
         this.isReady = true;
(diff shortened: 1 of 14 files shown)


jbms:
What I'm trying to get at is whether there is a way to make neuroglancer just do the right thing without having to configure anything.

manuel-castro:
The PR is mainly aimed at improving Neuroglancer sliceview performance when one is only interested in a specific resolution. Neuroglancer downloads chunks from all MIP levels that are visible, which is wasteful in this case, particularly on slower connections.

jbms:
In your typical usage of this, is it the low-resolution data that you want to skip or the higher-resolution data that you want to skip?

The reason I'm a bit hesitant is that it seems that you would have to constantly change these parameters if you were to change the zoom level, so I'm wondering if there is a way to specify this in a zoom-level-independent fashions, or just do something automatic.

Some example ideas:
 - Currently the sliceview continues to higher-resolution data if it improves the voxel size in a given dimension until the voxel size in that dimension is <= 1 screen pixel.  The 1 screen pixel limit could be changed to be a parameter, that could be increased.

 - Currently sliceview after determining the highest resolution data to use, also uses all lower-resolution data.  Instead, there could be some limit such that if the voxel size is > some number of screen pixels, it is excluded.

manuel-castro:
Usually you'd want to skip the lower resolution data, often desiring only to load/see the highest appreciable resolution. I see your concern about changing the zoom level, but that case is also handled indirectly: when the resolutions within the specified range are not appreciable/visible, it falls back to the highest appreciable resolution.

PR #122, opened 2019-02-02
Feature: Computational Datasource

larrylindsey:


src/neuroglancer/layer_dialog.ts
```diff
@@ -214,7 +214,9 @@ export class LayerDialog extends Overlay {
     this.setInfo('Validating volume source...');
     const token = this.volumeCancellationSource = new CancellationTokenSource();
     this.manager.dataSourceProvider
-        .getVolume(this.manager.chunkManager, url, /*options=*/undefined, token)
+        .getVolume(
+            this.manager.chunkManager, url,
+            /*options=*/{dataSourceProvider: this.manager.dataSourceProvider}, token)
```
jbms: I think it would make sense to move this logic of passing through the dataSourceProvider to the getVolume, etc. functions in neuroglancer/datasource/index.ts that way it doesn't have to be duplicated.

src/neuroglancer/user_layer_with_volume_source.ts
```diff
@@ -40,9 +40,14 @@ function helper<TBase extends BaseConstructor>(Base: TBase) {
...
+        const volumeOptions = this.volumeOptions ?
+            this.volumeOptions :
+            {dataSourceProvider: this.manager.dataSourceProvider};
+        if (!volumeOptions.dataSourceProvider) {
+          volumeOptions.dataSourceProvider = this.manager.dataSourceProvider;
+        }
         const multiscaleSource = this.multiscaleSource = getVolumeWithStatusMessage(
-            this.manager.dataSourceProvider, this.manager.chunkManager, volumePath,
-            this.volumeOptions);
+            this.manager.dataSourceProvider, this.manager.chunkManager, volumePath, volumeOptions);
```
jbms: Same comment about moving logic to datasource/index.ts

src/neuroglancer/datasource/computed/frontend.ts
```diff
@@ -0,0 +1,309 @@
...
+  // When we can't find data on the frontend, it usually means that we're in a
+  // race condition: we started listening just after the origin chunk's data
+  // was sent to the frontend, but this request arrived before the data was
+  // fully transferred. This is a fairly rare occurrence, but it does happen.
+  // Here, we wait for progressively longer periods of time, giving up after
+  // about 1.25s.
+  //
+  // This is simpler than a listener-callback structure, but is not
+  // deterministic.
+
```
jbms: Messages are always in order, but the issue this is trying to solve is caused by the Chunk.update logic in chunk_manager/frontend.ts, which defers handling of chunk update requests from the backend in order to avoid harming interactivity of the UI thread.  It would make sense to integrate this RETURN_FRONTEND_REQUEST handling with the same Chunk.update queue so that you don't have to retry.

src/neuroglancer/chunk_manager/base.ts
```diff
@@ -35,9 +34,23 @@ export enum ChunkState {
   // Download failed.
   FAILED = 6,
 
-  EXPIRED = 7
+  EXPIRED = 7,
+
+  COMPUTING = 8
 }
 
+export const CHUNK_STATE_NAMES = [
```
jbms: The TypeScript compiler actually already provides this: ChunkState[ChunkState.COMPUTING] === 'COMPUTING'

src/neuroglancer/chunk_manager/frontend.ts
```diff
@@ -156,6 +158,26 @@ export class ChunkQueueManager extends SharedObject {
...
+      const key = update['id'];
+
+      const chunk = source.chunks.get(key);
+      const data = chunk ? (<any>chunk)['data'] : undefined;
+      let message: string|undefined = undefined;
+      if (!data) {
+        if (chunk) {
+          message = 'Chunk has no data';
+        } else {
+          message = `No chunk found at ${key} for source ${source.constructor.name}`;
```
jbms: Can this error ever occur?

larrylindsey: In theory, yes, both of these could occur, but they would most likely be due to a programming error. The case of an existing chunk with no data would most likely happen if someone were to try to somehow implement a computation on non-volume chunks.

src/neuroglancer/chunk_manager/backend.ts
```diff
@@ -649,15 +760,15 @@ export class ChunkQueueManager extends SharedObjectCounterpart {
       // Note: After calling this, chunk may no longer be valid.
       this.updateChunkState(chunk, ChunkState.QUEUED);
     };
-    let promotionCandidates = this.queuedPromotionQueue.candidates();
+    let downloadPromotionCandidates = this.queuedDownloadPromotionQueue.candidates();
     let downloadEvictionCandidates = this.downloadEvictionQueue.candidates();
     let systemMemoryEvictionCandidates = this.systemMemoryEvictionQueue.candidates();
     let downloadCapacity = this.downloadCapacity;
     let systemMemoryCapacity = this.systemMemoryCapacity;
     while (true) {
```
jbms: It would be nice to factor out this logic so that it doesn't need to be repeated for the download and compute queues.

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,508 @@
...
+  private computationParams_?: ComputationParameters;
+
+  private cancellationToken_?: CancellationToken;
+  private resolve_?: () => void;
+  private reject_?: (reason: any) => void;
+  private success_ = false;
+  private failure_?: any;
+  private computing_ = false;
+
+  private outputSource_?: ComputedVolumeChunkSource;
```
jbms: No need for this separate member variable.  You can just put:

source: ComputedVolumeChunkSource;

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,508 @@
...
+    for (const outputChunk of this.activeComputations_.values()) {
+      for (const gridPosition of outputChunk.getOverlappingOriginGridPositions()) {
+        const gridString = gridPosition.toLocaleString();
+        if (!priorityMap.has(gridString)) {
+          priorityMap.set(
+              gridString,
+              {priority: outputChunk.priority, tier: outputChunk.priorityTier, pos: gridPosition});
+        } else {
+          const currentPriority = priorityMap.get(gridString)!;
+          currentPriority.priority = Math.max(currentPriority.priority, outputChunk.priority);
```
jbms: This logic isn't quite right, because you can't max/min tier and priority independently.  However, requestChunk already does this min/max logic so you can just use that directly rather than building priorityMap

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,508 @@
...
+    this.registerDisposer(this.chunkManager.recomputeChunkPrioritiesLate.add(() => {
+      this.updateChunkPriorities();
+    }));
+  }
+
+  updateChunkPriorities() {
+    const priorityMap = new Map<String, {priority: number, tier: number, pos: vec3}>();
+
+    for (const outputChunk of this.activeComputations_.values()) {
+      for (const gridPosition of outputChunk.getOverlappingOriginGridPositions()) {
```
jbms: If outputChunk priorityTier is RECENT, you can bail out immediately here.  However, I think if the cancellation logic is correct that should never happen.

jbms: I guess the cancellation logic could involve deferring some callbacks, in which case updateChunkPriorities might get called while cancellation is still in progress, so probably best not to rely on that.

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,508 @@
...
+
+  private cancellationToken_?: CancellationToken;
+  private resolve_?: () => void;
+  private reject_?: (reason: any) => void;
+  private success_ = false;
+  private failure_?: any;
+  private computing_ = false;
+
+  private outputSource_?: ComputedVolumeChunkSource;
+  private originGridPositions_ = new Map<String, vec3>();
```
jbms: It would be good to document the key/value meaning.

jbms: More generally it would be quite helpful to document the meaning/invariants of these members --- I know most of the code in Neuroglancer currently doesn't set a great standard in that regard, but I think it would be particularly helpful here.

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,508 @@
...
+      }
+    }
+  }
+
+  unregisterChunk(chunk: ComputedVolumeChunk) {
+    const key = chunk.key!;
+    this.activeComputations_.delete(key);
+  }
+
+  requestChunkData(
```
jbms: You could simplify this logic by using registerPromiseRPC / promiseInvoke (see e.g. datasource/nifti/backend.ts and datasource/nifti/frontend.ts) --- it would eliminate the need for you to separately manage frontendChunkRequests, cancellation, etc.

src/neuroglancer/datasource/computed/frontend.ts
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  getMeshSource() {
+    return null;
+  }
+}
+
+
+export class ComputedDataSource extends DataSource {
+  parseConfig(config: string, {} /*dataSourceProvider: DataSourceProvider*/) {
```
jbms: Why have an extra unused argument?

src/neuroglancer/datasource/computed/frontend.ts
```diff
@@ -0,0 +1,258 @@
...
+    // config is expected to be like computeName(computeParams)source(url)
```
jbms: Clarify accepted syntax, perhaps with an example.

It also seems likely that the parameters will be sufficiently complicated that no one would type them in directly, in which case you could simplify things by just using JSON encoding of the whole thing.  The code in datasource/index.ts doesn't place any restrictions on the contents of the datasource "url" after the protocol:// part.

src/neuroglancer/datasource/computed/frontend.ts
```diff
@@ -0,0 +1,258 @@
...
+      cancellationToken: CancellationToken) {
+    if (!options.dataSourceProvider) {
+      return Promise.reject('Need a DataSourceProvider');
+    }
+
+    const dataSourceProvider = options.dataSourceProvider!;
+    const [sourceNames, sourceConfigs] = this.parseConfig(config, dataSourceProvider);
+
+    if (sourceNames.length < 2) {
+      return Promise.reject();
```
jbms: It would be good to reject with an informative error message using: reject(new Error("Message here")).

jbms: (That applies to elsewhere as well.)

src/neuroglancer/datasource/computed/frontend.ts
```diff
@@ -0,0 +1,258 @@
...
+   * @returns a Promise containing parameters passed to the backend
+   *   computation.
+   */
+  abstract initialize(
+      config: string, volumes: VolumeChunkSource[][],
+      params: ComputedVolumeDataSourceParameters): Promise<T>;
+}
+
+export interface VolumeComputationFrontendProvider<T extends ComputationParameters> {
+  makeComputation(): VolumeComputationFrontend<T>;
```
jbms: I think it might make sense to have this function do the work of initialize, returning a Promise<VolumeComputationFrontend> (you could give VolumeComputationFrontend a constructor that takes the parameters) --- that would be a bit more consistent with how DataSourceProvider works.

src/neuroglancer/datasource/computed/frontend.ts
larrylindsey: That makes sense to me, and yes, computational datasources are likely to have fairly complicated configurations. What would you think about modifications to the UI to make it easier to manage those? (Not in this PR, though)

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: Done

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: Fixed.

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: I ended up bailing on RECENT tier chunks. Lmk if this looks right?

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: Comments added generally. If there are any areas of the code that need more explanation, I'm happy to add it.

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: As I mentioned in a side conversation, this turned out to be much more difficult to implement than I'd hoped, but I might give it another go.

src/neuroglancer/datasource/computed/frontend.ts
larrylindsey: Removed.

src/neuroglancer/datasource/computed/frontend.ts
larrylindsey: Done.

src/neuroglancer/datasource/computed/frontend.ts
larrylindsey: Done.

jbms:
One issue that I realized is that the original chunks required for a given computed chunk are assigned VISIBLE priority which means they get moved to the frontend and then copied to the GPU.  This is an unfortunate consequence of the fact that priority tier is also used to indicate whether a chunk should get copied to the GPU.

I think the solution would be to introduce an additional per-chunk flags bitvector that indicates whether the chunk is needed in the UI thread/GPU (per one flag for UI thread, one flag for GPU that implies UI thread).

src/neuroglancer/datasource/computed/frontend.ts
jbms: Yeah, agreed that UI modifications would make sense in a future PR.

src/neuroglancer/chunk_manager/backend.ts
```diff
@@ -861,3 +986,8 @@ registerRPC(CHUNK_SOURCE_INVALIDATE_RPC_ID, function(x) {
   const source = <ChunkSource>this.get(x['id']);
   source.chunkManager.queueManager.invalidateSourceCache(source);
 });
+
```
jbms: I think this is no longer needed, but you could re-purpose CHUNK_SOURCE_FETCH_RPC_ID for the promise rpcs.

jbms: Nevermind, I thought you were still using promiseInvoke

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,617 @@
...
+
+  /**
+   * Sets up computation parameters, computes overlapping origin chunks and
+   * initializes the input buffer.
+   * @param computationParams computation parameters
+   * @param cancellationToken cancellation token
+   */
+  initializeComputation(
+      computationParams: ComputationParameters, cancellationToken: CancellationToken) {
+    if (!this.source) {
```
jbms: I think if you just create the promise here, you could avoid having the separate success_, failure_ variables and the logic for handling them, and just always directly call this.reject_ or this.resolve_.  Promise already ignores calls to reject/resolve after one has already been called so you don't have to deal with that.

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,617 @@
...
+
+          const inputSpec = this.computationParams_!.inputSpec;
+          const destination = getArrayView(this.inputBuffer_!, inputSpec.dataType);
+          const numChannels = originSource.spec.numChannels;
+          const rawSource = this.maybeDecodeBuffer_(
+              source, inputSpec.dataType, originChunk.chunkDataSize!, numChannels);
+          copyBufferOverlap(
+              chunkCorner, chunkSize, rawSource, this.inputLower_!, inputSpec.size, destination,
+              inputSpec.dataType);
+          setTimeout(() => this.checkDone_(), 0);
```
jbms: What's the reason for the setTimeout?

larrylindsey: checkDone calls the computation when the buffer is full, which can take an arbitrary amount of time. The setTimeout call is to prevent it blocking the queue handling.

src/neuroglancer/chunk_manager/backend.ts
larrylindsey: Ah, yeah. I have an incoming update to this PR...

src/neuroglancer/datasource/computed/frontend.ts
```diff
@@ -0,0 +1,249 @@
...
+export class ComputedDataSource extends DataSource {
+  getOriginVolumes(
+      dataSourceProvider: DataSourceProvider, originUrl: string, chunkManager: ChunkManager,
+      cancellationToken: CancellationToken): Promise<ComputedVolumeSpecs> {
+    const slashPosition = originUrl.indexOf('://');
+    const config = originUrl.substring(slashPosition + 3);
+    return new Promise((resolve, reject) => {
+             const dataSource = dataSourceProvider.getDataSource(originUrl)[0];
+             if (!dataSource || !dataSource.getVolume) {
+               reject(`Unable to fetch data source for URL ${originUrl}`);
```
jbms: reject should always be called with new Error("...") since that is what downstream code is expecting  (same comment applies elsewhere)

larrylindsey: Done.

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: Done.

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: Done.

src/neuroglancer/datasource/computed/frontend.ts
```diff
@@ -0,0 +1,252 @@
...
+
+
+export class ComputedDataSource extends DataSource {
+  getOriginVolumes(
+      dataSourceProvider: DataSourceProvider, originUrl: string, chunkManager: ChunkManager,
+      cancellationToken: CancellationToken): Promise<ComputedVolumeSpecs> {
+    const slashPosition = originUrl.indexOf('://');
+    const config = originUrl.substring(slashPosition + 3);
+    return new Promise((resolve, reject) => {
+             const dataSource = dataSourceProvider.getDataSource(originUrl)[0];
```
jbms: You can just call dataSourceProvider.getVolume

src/neuroglancer/datasource/computed/frontend.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    const computationName = verifyString(configObj['computation']);
+    const computationProvider = ComputedDataSource.computationMap.get(computationName);
+
+    const originUrl = verifyString(configObj['origin']);
+
+    if (!computationProvider) {
+      return Promise.reject(new Error(`Unable to find computation ${computationName}`));
+    }
+
```
jbms: Add memoization based on the config, using chunkManager.memoize.getUncounted.  Otherwise you'll lose the cached data too easily, as layers can be reloaded from their JSON representation frequently (e.g. when using the Python API).  Refer to e.g. datasource/brainmaps/frontend.ts:getVolume

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,594 @@
...
+    const inputCenter = vec3.divide(vec3.create(), outputCenter, scaleFactor);
+    const inputSize = inputSpec.size;
+    this.inputLower_ =
+        vec3.subtract(vec3.create(), inputCenter, vec3.divide(vec3.create(), inputSize, twos));
+    const bufferLength = prod(inputSize) * inputSpec.numChannels;
+    const originDataType = inputSpec.dataType;
+
+    // Signal that we're now taking up memory. This value will be overwritten
+    // post-computation by a call to decodeRawChunk.
+    this.systemMemoryBytes = bufferLength * DATA_TYPE_BYTES[originDataType];
```
jbms: It is nice that this code is able to account for the memory used.  However, I think this will lead to systemMemoryCapacity accounting being incorrect, because the current code doesn't expect the chunk.systemMemoryBytes value to change while in a queue state where chunk.systemMemoryBytes is included in systemMemoryCapacity.  Previously, the systemMemoryBytes would only change while in DOWNLOADING state, which was not included in systemMemoryCapacity.  The same issue would also apply when calling decodeRawChunk later.  To fix this, we probably need a separate variable, like newPriorityTier, newPriority, etc., and then only actually update systemMemoryBytes when the queue state changes.  Then you could also set systemMemoryBytes in the chunk constructor rather than in initializeComputation since the transition to COMPUTING state happens before initializeComputaiton is called.

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,594 @@
...
+    }
+    if (!this.chunkDataSize) {
+      throw new Error('initializeComputation must be called after computeChunkBounds.');
+    }
+    this.computationParams_ = computationParams;
+    this.cancellationToken_ = cancellationToken;
+    this.computing_ = false;
+
+    this.cancellationToken_.add(() => {
+      if (this.reject_) {
```
jbms: It looks like this.reject_ is never reset, so I think all of these checks for this.reject_ can be removed.

src/neuroglancer/datasource/computed/backend.ts
```diff
@@ -0,0 +1,594 @@
...
+   * Idempotently performs the computation, if the input buffer is ready. This
+   * function should be called after a timeout in most cases, because it may
+   * take a long time to return.
+   */
+  private checkDone_() {
+    if (this.computing_) {
+      return;
+    }
+    if (this.originGridPositions_.size === 0) {
+      this.computing_ = true;
```
jbms: I think this.cleanup() could actually be called here, since the cleanup it does is only related to copying of origin chunks, not the computation.

src/neuroglancer/datasource/computed/backend.ts
jbms: This should also reset this.inputBuffer_ so that if the compute is cancelled, but the chunk remains in QUEUED state because it is still needed, the memory is not held.

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: The problem I see with this approach is that the chunk has basically no context in the constructor. I think I'd prefer to add a function call on the chunk object that returns an estimate of the memory it would occupy, perhaps even at a given state, replacing the systemMemoryBytes variable. Or, I could see implementing systemMemoryBytes as a getter in this one specific case.

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: Done.

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: Done.

src/neuroglancer/datasource/computed/backend.ts
larrylindsey: Done.

src/neuroglancer/datasource/computed/frontend.ts
larrylindsey: Done.

src/neuroglancer/datasource/computed/frontend.ts
larrylindsey: Done.

src/neuroglancer/datasource/computed/backend.ts
jbms: Good point, the constructor is too soon.  But you could override initializeVolumeChunk.  After you call super.initializeVolumeChunk, I think all relevant parameters should be present.

jbms:
Everything looks great, and this also fixes a few hacks with how the queuing worked.

Issue #125, opened 2019-03-12
do not work with tornado>=5.1.1


jingpengw:
Thanks for this great package. 

The default installation from setup.py will install tornado 5.1.1, and is not working properly now. It works with tornado 5.0.

This is the error message I got:
```
In [1]: import neuroglancer as ng                                                                           
---------------------------------------------------------------------------                                 
ImportError                               Traceback (most recent call last)                                 
<ipython-input-1-4c1b9d133e96> in <module>()                                                                
----> 1 import neuroglancer as ng                                                                           
                                                                                                            
/usr/local/lib/python3.5/dist-packages/neuroglancer/__init__.py in <module>()                               
     14                                                                                                     
     15 from __future__ import absolute_import                                                              
---> 16 from .server import set_static_content_source, set_server_bind_address, is_server_running, stop     
     17 from .static import dist_dev_static_content_source                                                  
     18 from .viewer import Viewer, UnsynchronizedViewer                                                    
                                                                                                            
/usr/local/lib/python3.5/dist-packages/neuroglancer/server.py in <module>()                                 
     28 import tornado.web                                                                                  
     29                                                                                                     
---> 30 import sockjs.tornado                                                                               
     31                                                                                                     
     32 from . import local_volume, static                                                                  
                                                                                                            
/usr/local/lib/python3.5/dist-packages/sockjs/tornado/__init__.py in <module>()                             
      1 # -*- coding: utf-8 -*-                                                                             
      2                                                                                                     
----> 3 from .router import SockJSRouter                                                                    
      4 from .conn import SockJSConnection                                                                  
                                                                                                            
/usr/local/lib/python3.5/dist-packages/sockjs/tornado/router.py in <module>()                               
      9 from tornado import ioloop, version_info                                                            
     10                                                                                                     
---> 11 from sockjs.tornado import transports, session, sessioncontainer, static, stats, proto              
     12                                                                                                     
     13                                                                                                     
                                                                                                            
/usr/local/lib/python3.5/dist-packages/sockjs/tornado/transports/__init__.py in <module>()                  
      1 # -*- coding: utf-8 -*-                                                                             
      2                                                                                                     
----> 3 import sockjs.tornado.transports.pollingbase                                                        
      4                                                                                                     
      5 from .xhr import XhrPollingTransport, XhrSendHandler                                                
                                                                                                            
/usr/local/lib/python3.5/dist-packages/sockjs/tornado/transports/pollingbase.py in <module>()               
[...]

PR #126, opened 2019-03-13
fix(python): Constrain tornado<6 due to sockjs-tornado incompatibility

jbms:
This works around issue #125.


chrisroat:
Does https://github.com/mrjoes/sockjs-tornado/pull/118 allow this change to be undone?

PR #128, opened 2019-03-26
Bug fixes to facilitate eventual npm module

neomorphic:
In order to eventually publish neuroglancer as an npm module, it would need to be transpiled to javascript before upload. During my experimentation with converting the code I noticed a few bugs with some of the import statements that will cause this process to fail. 

It also looks like the key_name_logger code can no longer be built, so I removed that in an effort to clean up. Apologies if that is not required.


jbms:
Thanks!

PR #132, opened 2019-04-08
feat(annotations): UI interaction changes.

larrylindsey:
A handful of changes that are meant to make annotating more intuitive:
* When an annotation tool is active, the cursor changes to a crosshair
  PNG.
* Mouse button 1 (middle) may be used for dragging as well as button 0.
* Annotations are added via unmodified mouse button 0 clicks, instead of
  ctrl+click. Translate/rotate is disabled for this button when a tool
  is active.
* The button that activates an annotation tool is darkend when clicked.
* The Esc key will clear an active annotation tool.


jbms:
One concern I have regarding the mouse bindings change is that some machines (like my laptop) don't have a good way to generate mouse1 events, so we would effectively lose the ability to navigate at all while navigating.

I'm not sure exactly how it works out in common use, but the model I had in mind for annotations is that you would commonly keep the annotation tool active for a long time, but use it relatively rarely mixed in with a lot of other navigation.  E.g. you might keep the point annotation tool active and use it to flag certain things.  For painting it could be somewhat different, though I think even in that case you would have more navigation than painting.

larrylindsey:
In my experience, the most common use case is to actively set an annotation tool when its in heavy use, but escape out of it when a user is panning around. This could be made easier by assigning a key combo to the tool buttons. The paradigm of using the middle button to navigate and the left button for tool use is a common one, especially in image manipulation programs (gimp, photoshop, etc) and CAD software, so I think its kind of important that we align to it, especially as voxel painting moves along.

You make a good point, though, that not every laptop computer has a middle button. What would you thing of binding navigation to ctrl+mouse0?

jbms:
I think it is not great to switch around the bindings when an annotation tool is selected.  In practice I think people will still use mouse0 drag for normal panning when not annotating, since it is used in google maps and other 2-d viewers of the sort, and holding the middle mouse button or control is more cumbersome to the point that people won't do it if it isn't necessary.  Therefore, in practice that will mean that when an annotation tool is selected, the familiar mouse0 drag binding now does something else (and in particular something that may make a persistent change to a changestack), and a different binding has to be used for navigation.

It is true that middle mouse drag is commonly used for panning in Gimp and cad software, but I think it is less common for mouse0 drag to also pan when no other tool is selected --- I think the "default" action is normally a relatively benign "select" action or something similar.   In the cad software (typically 3d views, I think) where this binding switching happens depending on whether a tool is selected, I've found it annoying and frequently leads to doing the wrong action -- I think seasoned users probably just avoid relying on the mouse0 drag as a pan action altogether.  With CAD software and probably more serious image editing use cases, a mouse is pretty much a requirement.

What about reversing things from how you've proposed, and making both mouse1 and control+mouse0 trigger an annotation?  This is backwards from gimp but avoids the switching of bindings problem.

larrylindsey:
Well, I didn't mention the precedence in EM annotation software (I'd hoped the image manipulation paradigm would be convincing enough) - this is the same way that all of the annotation software in the field that I can think of already works - CATMAID, VAST, Armitage, TrakEM2, and Reconstruct all work this way. Often panning, zooming, etc, is performed by a separate tool, in the same class as annotating or painting. Most people using Neuroglancer come from at least one of the first three, and I think that it is very important to maintain that paradigm. Serious annotators will probably choose to use a 3+ usb mouse even when using a laptop, just as they would for other image manipulation software.

That said, there is room for differing expectations and opinions. I think there's an argument for making the key and mouse bindings user-configurable.

jbms:
I think aside from the precise bindings that we choose, there are some higher level issues:

1. Currently there is just a single tool-dependent "annotation" action (bound to control+mousedown0).  However, programs like gimp, cad software, catmaid all support multiple tool-dependent actions, via leftlick, control+leftclick, shift+leftclick, etc.  Currently we have a "move annotation" binding which is not tool-specific, but probably should be changed to be once there are tools not related to spatial annotations.  "delete annotation" would be another useful action to support via a tool-specific binding.  That means we need to find room for probably at least 3 tool-specific bindings, likely more (e.g. voxel painting probably needs some sort of mouse wheel binding for adjusting brush size).

2. Pretty much the full set of navigation bindings need to be available during tool use; in particular, you may need to place one end of a line annotation, then navigate to a different location and/or zoom level, maybe even a different panel, and then place the other end.  Whether to make toggling selected segments a tool-independent action or itself a segmentation-related tool is a bit unclear.  For voxel painting you might not want to toggle segments, but when placing spatial annotations toggling segments is more likely to be useful.

3. Given that the navigation bindings need to remain available at all times, I think it is undesirable for the navigation bindings to change depending on whether there is an active tool, since that seems like a likely source of user interaction error.  I think it is better to just settle on one set of bindings that work for navigation.

larrylindsey:
I think your points 1 and 2 are good ones. To step back, I'd say that it makes sense to implement a UI manager that maps (event, context) => (action), instead of the current (event) => (action), or even the (event) => (action + context) currently in this PR.

As to point 3, as I said before, most existing programs change the left-click bindings already, and user confusion hasn't been an issue. (And I say "most," but really I'm hedging - as far as I'm aware the precedent is universal, but its possible that there are some that I'm unfamiliar with that behave differently).

But, overall, I think what it boils down to is how you view Neuroglancer: if you see it as primarily a viewer with some annotation tools on the side, then I can see how it might feel weird to use the same action for navigation and annotation. On the other hand, if you're approaching it as an annotation workhorse to replace something like Armitage or VAST, then ctrl+click annotation is going to seem kind of awkward and clunky.

Ultimately, though, what really matters here is what an "average" or "normal" user will find to be intuitive, and neither of us really falls into that category. Which, this brings me back to the first thing I said. Building a context-aware event->action map will allow us to make this configurable so that a) we can do user tests and b) allow user-defined maps to customize Neuroglancer to a given use case.

jbms:
Regarding point 3 (changing of bindings depending on selected tool), I agree that it is universal for the bindings to be tool-dependent --- in that when the "rectangle select" tool is enabled in gimp, left click starts a rectangle selection, while when the "pen" tool is enabled, left click starts drawing.  The behavior which is less universal and more problematic, in my opinion, is when a command doesn't go from being unavailable to being available with left click, but rather goes from being available with left click to available with a different binding.  E.g. in gimp I don't think you can ever use leftclick+drag for panning, and it certainly isn't the default.  I have seen some cad software do that when in 3-d view, where if you select the "orbit" tool leftclick+drag pans, but and there I have found it somewhat error prone --- in practice I just learn to never rely on the leftclick+drag for navigation, and always just use the "remapped" pan binding (typically middle mouse drag, or sometimes middle+left drag).

jbms:
Regarding customizing the bindings, that is already possible through the python integration.  Outside the python integration it is not possible, mainly because it is not clear how to do it in a reasonable way, given that there isn't just a single neuroglancer host.

larrylindsey:
I think it should be possible to build a tab widget, similar to how annotations work, that would allow for customization by user input. I could also imagine a drop-down with a handful of presets. The user map could be stored in some browser-side store. What do you think?

I think our disagreement over the aforementioned point 3 is a good illustration of why custom UI bindings are important.

jbms:
I agree custom bindings would be nice to support in general, but there are some challenges: there are multiple urls from which neuroglancer is accessed --- e.g. the public neuroglancer-demo.appspot.com, the internal google instance, development instances.  Using browser-local storage would mean that those instances would all use separate bindings, which would be kind of annoying.  Also, there would be the question of how to handle changes made to the "default" neuroglancer bindings after the user has created custom bindings, changes to command names, etc.  I'm not sure it is worth the trouble at the moment.

larrylindsey:
It would be kind of annoying, but ultimately I don't think it would be that big a deal - most of people this would affect probably don't use multiple instances of neuroglancer, and even so it's easy to imagine semi-manual ways to sync the map between them. There may even be a way to avoid this problem, but I haven't done deep enough research to tell.

There are also several solutions to the problem of changing the default bindings. For instance, one could imagine storing only the settings that were changed by the user, and falling through to the default map for the nonexistent keys.

I share your uncertainty, at least for now. This is at least my third time implementing voxel painting tools, and while I may not have become very efficient at it, I have learned from experience that the UI can get complicated quickly. This isn't my first priority for now, but once I've made some headway on rendering and persistence, this will become fairly important. In terms of wall-time, I think we're looking at 1-2 months.

PR #134, opened 2019-04-16
modify single annotation instead of recreating the entire list

akhileshh:
Currently all the annotations are removed from the DOM and recreated everytime there is an add/update/delete on a single annotation. This affects performance as the list grows.
This PR aims to avoid that by targeting specific annotations when the list is modified.
There maybe better way to solve this but I think this solution avoids any changes to the core framework.

  * Create separate functions and signals for add/update/delete annotation and trigger signals accordingly
  * Create interface for common signals between AnnotationSource and MultiscaleAnnotationSource

### DIFF
--- a/src/neuroglancer/annotation/frontend_source.ts
+++ b/src/neuroglancer/annotation/frontend_source.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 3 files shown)


akhileshh:
I signed it!

jbms:
Thanks!

akhileshh:
Sure! The other way I was thinking is to add support to selectively dispatch functions instead of all, that would get rid of the interface with only signals. Maybe use a map in place of the array of functions.

I can work on it if it seems useful elsewhere.

PR #135, opened 2019-04-21
fix(skeleton): fix regression in client-side equivalences rendering

nkemnitz:
Same as in 97c522bf1271f24ba29baef34ca4d5035eb5dd98


jbms:
Thanks!

PR #141, opened 2019-06-30
Support for a LocalVolume with multiple resolutions

perlman:
This is an extension of LocalVolume to support multiple resolution local data sources. This was a quick fix to let us access large local multi-resolution data (e.g. hdf5 or zarr).

LocalMultiscaleVolume works by sending requests to the underlying LocalVolume. Metadata come from a reference layer (chosen by minimum voxel size).

Suggestions of an alternate/better implementation are welcome. 

(Implemented with @funkey and @sheridana.)


jingpengw:
thanks a lot for this PR! I was looking for it!

Issue #147, opened 2019-07-18
change defaultViewerOptions w/ webpack


falkben:
I am attempting to override `defaultViewerOptions` in [`viewer.js`](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/viewer.ts#L147-L150) in deployment using a webpack command.  Going off the suggestion in this issue [comment](https://github.com/google/neuroglancer/pull/107#issuecomment-426507492), I'm trying:  

`npm run build -- --define viewer.defaultViewerOptions="{showLayerDialog: false, resetStateWhenEmpty: true}"`

where the only difference is showLayerDialog set to `false` but this webpack command seems to *have no effect.*

I've tried also changing `defaultViewerOptions` in viewer.ts to be a `var` instead of a `const`, thinking that maybe webpack couldn't override the const.  I am hoping this is just something simple I'm missing about webpack -- thanks for your help.

Note, this is really simple, but if I can get this working, I'm hoping to return to the JSON url [PR](https://github.com/google/neuroglancer/pull/107) and convert it to a build option using a similar webpack override.

jbms:
I fully support making this use case work.  Note, though, that Webpack define is not module (or possibly even scope) aware, as far as I know.  It basically just replaces all instances of the key, as they appear exactly in the code, with the replacement code on the right.  I'm also not sure what it does to an existing definition within the code.

So you should be able to do what you want, but you may need to do it slightly differently or make some other changes to how those default options are defined in the code.

xgui3783:
I have made a somewhat example with this commit: https://github.com/xgui3783/neuroglancer/commit/9453eb1d640b88f723be2f24428239e086edd6ab
, which I believe is in the spirit of the referenced comment. and with the commit, the below script successfully runs neuroglancer with `showLayerDialog` as false 

```
npm run dev-server -- --define OVERWRITE_DEFAULT_VIEWER_OPTIONS="'showLayerDialog: false,resetStateWhenEmpty: true'"
```
I have also unsuccessfully tried to replace raw string (assignment of `defaultViewerOptions`). For that, you may need https://www.npmjs.com/package/string-replace-loader or https://www.npmjs.com/package/string-replace-webpack-plugin

falkben:
Cool, that worked! Thanks. Seems like a more natural place to put the declare would be just at the top of `viewer.ts`?

falkben:
See PR https://github.com/google/neuroglancer/pull/148

PR #148, opened 2019-07-19
default viewer build option

falkben:
This adds a build option for `defaultViewerOptions`

To overwrite the options on build, run the following command (this one disables the layer dialog):

`npm run build-min -- --define OVERWRITE_DEFAULT_VIEWER_OPTIONS="'showLayerDialog: false, resetStateWhenEmpty: true'"`

Thanks @xgui3783 for your [example](https://github.com/xgui3783/neuroglancer/commit/9453eb1d640b88f723be2f24428239e086edd6ab), which I modified slightly to make the declaration inside `viewer.ts` as well as handle builds which don't specify the default viewer options.

The syntax I'm using to check if it's undefined is a bit ugly -- happy to change to something else if you prefer.

Related to issue https://github.com/google/neuroglancer/issues/147

### DIFF
--- a/src/neuroglancer/viewer.ts
+++ b/src/neuroglancer/viewer.ts
@@ -57,6 +57,8 @@ import {TrackableScaleBarOptions} from 'neuroglancer/widget/scale_bar';
 import {makeTextIconButton} from 'neuroglancer/widget/text_icon_button';
 import {RPC} from 'neuroglancer/worker_rpc';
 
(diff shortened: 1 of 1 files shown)


jbms:
This is great, thanks.  My only suggestion is to rename to NEUROGLANCER_OVERRIDE_DEFAULT_VIEWER_OPTIONS

jbms:
Thanks!

PR #149, opened 2019-07-23
Feature: Free Rotation and Translation

ogewan:
Mousedrag and Shift + Mousedrag on perspective view only now implement pointer lock, pointer disappears and movement is unimpeded. Cursor restored upon mouseup


ogewan:
Woops this isn't ready for the google branch

PR #150, opened 2019-07-29
fix(python): work with jupyter-server-proxy

perlman:
A very small tweak to the token parsing in the python client to preserve the path.  The current regex expects the path to be in the form `^/v/<token>/`, which will not work with [Jupyter Server Proxy](https://github.com/jupyterhub/jupyter-server-proxy/).

With Jupyter Server Proxy server extension enabled in jupyter, the normal neuroglancer URL (e.g. `http://127.0.0.1:51119/v/2df69e9395e0adb8b0d14b7dbec99751220625aa/`) can be accessed as  `http://localhost:8888/proxy/51119/v/2df69e9395e0adb8b0d14b7dbec99751220625aa/`.

My goal here is have neuroglancer notebooks working with [binder](https://mybinder.org/), which does not allow remote access to additional ports except through a proxy.

### DIFF
--- a/src/neuroglancer/python_integration/server_connection.ts
+++ b/src/neuroglancer/python_integration/server_connection.ts
@@ -21,11 +21,11 @@ import SockJS from 'sockjs-client';
 import {StatusMessage} from 'neuroglancer/status';
 
 function getServerConnectionURL() {
-  const match = window.location.pathname.match(/^\/v\/([^\/]+)/);
+  const match = window.location.pathname.match(/^(.*)\/v\/([^\/]+)/);
   if (match === null) {
     throw new Error('Failed to determine token from URL.');
   }
-  return `${window.location.origin}/socket/${match[1]}`;
+  return `${window.location.origin}${match[1]}/socket/${match[2]}`;
 }
 
 const defaultReconnectionDelay = 1000;


jbms:
Thanks!

chrisroat:
@perlman This is a very nice addition.  I am running a jupyter hub with a jupyter server proxy running.  I can create a neuroglancer viewer, and translate the url to get to the running neuroglancer viewer. (woot!) 

However, the source is a local numpy volume and it doesn't seem to be getting served properly. 

My neuroglancer url:
```
[url]/user/chrisroat/proxy/37279/v/key1
```

The source volume:
```
python://volume/key1.key2
```

Neuroglancer translates this and tries to fetch the following, which is a 404:
<pre>
[url]/<b>hub</b>/neuroglancer/info/key1.key2
</pre>

I think it should instead be:
<pre>
[url]/<b>user/chrisroat/proxy/37279</b>/neuroglancer/info/key1.key2
</pre>

The webapp's fetch thinks it's root is `[url]/hub` instead of the proxy.  Any thoughts on how this is happening?

perlman:
@chrisroat It's been a while since I've tried to do this. I'll take a look later today and get back to you.

perlman:
Unsurprisingly, I run into the exact same problem using my old notebooks with a current neuroglancer.

As you noted, the path is being preserved for static content & websocket, but not for the dynamic volumes.

It looks like [this line](https://github.com/google/neuroglancer/blob/fb35aea0b44cf421f17871c32e1cdbc2dab24c35/src/neuroglancer/datasource/python/frontend.ts#L343) and [this line](https://github.com/google/neuroglancer/blob/9d2d5d10bd763842010bcb34dbda9e2a40697b31/src/neuroglancer/datasource/python/backend.ts#L44) are to blame, both using root-relative URLs.

See  #256 for a quick hacky solution.

PR #155, opened 2019-08-06
feat: support explicit segment colors

hubbardp:
In our experience at Janelia, it is helpful for proofreading tools to have explicit control over the colors of bodies in the segmentation.  This code adds that capability to Neuroglancer.  The display state of a layer now can include a map like the following, making one body purple and another green:

"segmentColors": { "193125660": "#ff00ff", "206326381": “#00ff00” }

The implementation uses a new Uint64Map class that is quite similar to the Uint64Set class.  I added some unit tests for both of those classes.

For slice rendering, there is a new SegmentStatedColorShaderManager class that uses a HashMapShaderManager and adds shader code to find a segment’s explicit color if present.  For mesh rendering, the getObjectColor function now checks for the segment’s explicit color.

### DIFF
--- a/src/neuroglancer/segmentation_user_layer.ts
+++ b/src/neuroglancer/segmentation_user_layer.ts
@@ -33,9 +33,10 @@ import {trackableAlphaValue} from 'neuroglancer/trackable_alpha';
 import {ElementVisibilityFromTrackableBoolean, TrackableBoolean, TrackableBooleanCheckbox} from 'neuroglancer/trackable_boolean';
 import {ComputedWatchableValue} from 'neuroglancer/trackable_value';
 import {Uint64Set} from 'neuroglancer/uint64_set';
+import {Uint64Map} from 'neuroglancer/uint64_map';
 import {UserLayerWithVolumeSourceMixin} from 'neuroglancer/user_layer_with_volume_source';
 import {Borrowed} from 'neuroglancer/util/disposable';
-import {parseArray, verifyObjectProperty, verifyOptionalString} from 'neuroglancer/util/json';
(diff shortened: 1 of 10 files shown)


hubbardp:
@googlebot I signed it!

src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -157,6 +161,21 @@ export class SegmentationUserLayer extends Base {
       this.specificationChanged.dispatch();
     });
 
+    verifyObjectProperty(specification, SEGMENT_STATED_COLORS_JSON_KEY, y => {
+      if (y !== undefined) {
+        let {segmentEquivalences} = this.displayState;
+        // Parse the color as a 32-bit integer here, because the Uint64.parseString()
+        // function does not seem to handle strings of the form '0xRRGGBB'.
+        // And support a CSS color "#RRGGBB" by turning into "0xRRGGBB".
+        let result = verifyObjectAsMap(y, x => parseInt(String(x).replace(/^#/, '0x'), 0));
```
jbms: You can use parseRGBColorSpecification or parseRGBAColorSpecification in util/color.ts to handle arbitrary CSS color specifications.  Handling alpha might be a bit more work but would probably be quite useful -- but I think that would be reasonable to put off if you don't want to do it now.

jbms:
Overall this looks great!  The only change I'd like to see is with the color parsing.

src/neuroglancer/segmentation_user_layer.ts
hubbardp: Thanks for the suggestion.  I just pushed an update using parseRGBColorSpecification.  I think I will put off looking into alpha handling.

jbms:
Thanks!

PR #159, opened 2019-08-13
fix: allow copy pasting layer location in Firefox

william-silversmith:
In Firefox, "disabled" disables any interaction with the
layer entry field whereas in Chrome it allows you to select
the text in the box -- a common workflow when using CloudVolume
to grab the layer path.

A standard property "readOnly" is available that works cross-browser.


jbms:
Thanks!

Issue #151, opened 2019-07-30
Control-click on the [+] button does not open annotation layer on Mac OS


ceesem:
Apparently control-clicking on the [+] button in the layer name list is supposed to open an annotation layer on, e.g., Linux. However, on Mac OS nothing happens when the user does this, presumably due to the OS reading control-click as a right mouse button. This is particularly tricky, since there is no other way to generate an annotation layer without going into the JSON. The data views have no such issues and control-click can produce annotations just fine, while right click moves the view around as expected. If doing the same thing for the [+] button is possible, making the same thing work on OS X with a command-click would be natural.

As an aside, this is really useful and not particularly discoverable (especially as a Mac user...). It might would be good to add it to the documentation under help button and perhaps also give 'create annotation layer' a keyboard shortcuts (perhaps control-n?).

PR #160, opened 2019-08-14
Mac annotation support

ogewan:
Fix for #151 
Control + click on Mac works as intended. This makes right click create annotations as well. Alternatively, ⌘+Click can also be used to create annotation layers.


ogewan:
@googlebot I fixed it.

jbms:
Thanks!

PR #161, opened 2019-09-10
copy and import button in state editor

ogewan:
It just works.


jbms:
I agree this is useful functionality.  Unfortunately clipboard usage from web pages is a bit tricky due to security measures that browsers have implemented.  In both Chrome and Firefox, "Copy state" works.  However, "Import state" does not work in Firefox, and in Chrome it only works if the user grants clipboard permissions to the site.

To handle "Paste" without needing additional permissions, instead of using a separate button you can wait for the user to do a paste (e.g. via the normal keyboard shortcut implemented by the browser) and then handle that event.  See for example src/neuroglancer/ui/default_clipboard_handling.ts  This approach does have the unfortunate downside that it can be tricky to make it obvious to the user what the effect of pasting will be.

PR #164, opened 2019-09-20
Fix: Update annotation position on childUpdated

ogewan:
A change to how annotationListElement are signaled did not update the position child element.


jbms:
Thanks for reporting this issue and creating this pull request.  In the referenced commit, I decided to fix this in a slightly different way that simplifies the update logic.

ogewan:
https://github.com/google/neuroglancer/commit/a05650cd604887dd6ba8a23855a7afe99a89b7d0#r35414033
The position values are not updating correctly. I tried debugging production :eyes: , but my breakpoints were skipped, suggesting the update is never called. Although it it might be an issue with debugging the production code, I'll look into it locally later on.

PR #167, opened 2019-10-14
fix(annotation/frontend_source): update offsets for new annotation only

tingzhao:
This PR for fixing a bug in the updateAnnotation function in annotation/frontend_source.ts. 

In the original code, offsets of types are updated no matter if the passed annotation is new. The problem can be triggered by painting a multi-point annotation. For example, when painting line segment, moving pointB following the mouse cursor before putting it down can keep adding offsets to type AXIS_ALIGNED_BOUNDING_BOX and ELLIPSOID, resulting in wrong offsets for these two types. In this fix I moved offset update inside the annotation existing condition check so that the offsets are updated only when the annotation hasn't been recorded.


jbms:
Thanks for catching this bug!

PR #171, opened 2019-10-25
feat(datasource/dvid): add point annotation functions for DVID

tingzhao:
This PR adds basic features for fetching, displaying and editing DVID bookmark annotations. It only involves changes in datasource/DVID, including:

1. Frontend DVIDAnnotationSource class for handling parameter parsing and point annotation creation. It will ask the user to specify in a user name to fetch annotations with a certain user tag. We do have more changes to handle annotations without tags by using multiscale chunks, but in this PR we decided to include tag-based annotation only.

2. Backend annotation fetching and parsing functions and DVIDAnnotationSource class designed for our own bookmark JSON format.

3. Helper functions in datasource/dvid/api.ts and datasource/dvid/utils.ts, including those for handling special properties in DVID bookmarks, which are instantiated the DVIDPointAnnotation class. The DVIDPointAnnotation class extends from Point to have additional properties. The helper functions will parse these additional properties and render annotations accordingly with our own shaders.

According to our testing, these changes do not affect any other functions in Neuroglancer if no DVID annotation layer is specified.

### DIFF
--- a/src/neuroglancer/datasource/dvid/api.ts
+++ b/src/neuroglancer/datasource/dvid/api.ts
@@ -1,0 +1,47 @@
+import {CancellationToken, uncancelableToken} from 'neuroglancer/util/cancellation';
+import {responseJson, cancellableFetchOk} from 'neuroglancer/util/http_request';
+
+export interface HttpCall {
+  method: 'GET' | 'POST' | 'DELETE';
+  path: string;
+  payload?: string;
+}
+
+export class DVIDInstance {
+  constructor(public baseUrl: string, public nodeKey: string) {}
+
+  getNodeApiUrl(): string {
+    return `${this.baseUrl}/api/node/${this.nodeKey}`;
+  }
+}
+
+function responseText(response: Response): Promise<any> {
+  return response.text();
+}
+
+export function makeRequest(
+  instance: DVIDInstance,
+  httpCall: HttpCall & { responseType: 'arraybuffer' },
+  cancellationToken?: CancellationToken): Promise<ArrayBuffer>;
+
+export function makeRequest(
+  instance: DVIDInstance,
+  httpCall: HttpCall & { responseType: 'json' }, cancellationToken?: CancellationToken): Promise<any>;
+
+export function makeRequest(
+  instance: DVIDInstance,
+  httpCall: HttpCall & { responseType: '' }, cancellationToken?: CancellationToken): Promise<any>;
+
+export function makeRequest(
+  instance: DVIDInstance,
+  httpCall: HttpCall & { responseType: XMLHttpRequestResponseType },
+  cancellationToken: CancellationToken = uncancelableToken): any {
+    let requestInfo = `${instance.getNodeApiUrl()}${httpCall.path}`;
+    let init = { method: httpCall.method, body: httpCall.payload };
+
+    if (httpCall.responseType === '') {
+      return cancellableFetchOk(requestInfo, init, responseText, cancellationToken);
+    } else {
+      return cancellableFetchOk(requestInfo, init, responseJson, cancellationToken);
+    }
+}
--- a/src/neuroglancer/datasource/dvid/backend.ts
+++ b/src/neuroglancer/datasource/dvid/backend.ts
@@ -15,7 +15,7 @@
  */
 
 import {WithParameters} from 'neuroglancer/chunk_manager/backend';
-import {MeshSourceParameters, SkeletonSourceParameters, VolumeChunkEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/dvid/base';
+import {AnnotationSourceParameters, MeshSourceParameters, SkeletonSourceParameters, VolumeChunkEncoding, VolumeChunkSourceParameters} from 'neuroglancer/datasource/dvid/base';
 import {assignMeshFragmentData, decodeTriangleVertexPositionsAndIndices, FragmentChunk, ManifestChunk, MeshSource} from 'neuroglancer/mesh/backend';
 import {SkeletonChunk, SkeletonSource} from 'neuroglancer/skeleton/backend';
 import {decodeSwcSkeletonChunk} from 'neuroglancer/skeleton/decode_swc_skeleton';
@@ -25,7 +25,16 @@ import {VolumeChunk, VolumeChunkSource} from 'neuroglancer/sliceview/volume/back
 import {CancellationToken} from 'neuroglancer/util/cancellation';
 import {Endianness} from 'neuroglancer/util/endian';
 import {cancellableFetchOk, responseArrayBuffer} from 'neuroglancer/util/http_request';
-import {registerSharedObject} from 'neuroglancer/worker_rpc';
+import {registerSharedObject, SharedObject, RPC} from 'neuroglancer/worker_rpc';
+import {vec3} from 'neuroglancer/util/geom';
+import {Uint64} from 'neuroglancer/util/uint64';
+import {DVIDInstance, makeRequest} from 'neuroglancer/datasource/dvid/api';
+import {DVIDPointAnnotation, updateAnnotationTypeHandler} from 'neuroglancer/datasource/dvid/utils';
+import {Annotation, AnnotationId, AnnotationSerializer, AnnotationType} from 'neuroglancer/annotation';
+import {AnnotationGeometryChunk, AnnotationGeometryData, AnnotationMetadataChunk, AnnotationSource, AnnotationSubsetGeometryChunk} from 'neuroglancer/annotation/backend';
+import {verifyObject, verifyObjectProperty, verifyOptionalString, parseIntVec} from 'neuroglancer/util/json';
+import {ChunkSourceParametersConstructor} from 'neuroglancer/chunk_manager/base';
+
 
 @registerSharedObject() export class DVIDSkeletonSource extends
 (WithParameters(SkeletonSource, SkeletonSourceParameters)) {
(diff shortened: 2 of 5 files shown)


jbms:
For the user-based tag, would it make sense to put that in the datasource URL so that you don't have to prompt for it?  It seems like having to enter it manually each time the browser refreshes may be undesirable.  Alternatively, it could be cookie based, but I'm not sure how convenient that would be.

Overriding the POINT rendering is kind of problematic since that does not work for the case where there are a mixture of DVID and non-DVID annotation layers.  The proper solution would probably be for an annotation source to support arbitrary per-annotation properties, with various supported types, and then allow defining a custom shader that sets the color based on those properties.  That would be a larger effort though.  Perhaps alternatively you could create a new annotation type?

It looks like currently every point annotation can be in one of 4 classes, meaning "default", "checked", "split", or "merge", and the goal is to display them with different colors/border colors?

tingzhao:
The prompt box is a kind of temporary solution for authentication to be supported by DVID. We did have a version that stores the user name in cookies, but that requires some additional package (es-cookie) that supports cookies. We discarded this solution since we don't want to make any changes outside of dvid data source. Due to some future plan of DVID authentication,  if I understand correctly, @stephenplaza is against the idea of including the user name in the URL, which in fact was in the original implementation.

For our annotations, we have classes for "default", "split" and "merge". "checked" is an additional tag. We want to display them in distinguishable appearances, ideally with different shapes. But we only found ways to change colors without modifying code outside of dvid. We have designed the serialization and rendering overriding to be compatible with the default handlers. Therefore A non-DVID layer should also be displayed correctly. The additional attribute in DVID annotation serialization will be set to 0 for a non-DVID layer, making rendering behave the same as the default rendering does. But I agree that this creates some kind of coupling between these two parts, i.e. the compatibility might be broken if the default rendering is changed in the future.

Adding a new annotation type could be a better solution. Do you have any more suggestions on going this direction so that our implementation will not break any design principles in Neuroglancer ?

jbms:
What other shapes would you like to use?  Currently the point annotations are drawn as squares (two triangles), and then the fragment shader just discards pixels outside the boundary of the circle, and chooses the center vs. edge color and handles the antialiasing.  So any other simple geometric shape wouldn't be too hard to draw instead using the same method.  How to fit that into the UI and the existing annotation model, though, is another matter.

In general because Neuroglancer does not yet support additional per-annotation attributes, we have used a workaround of having a separate layer per "class" --- e.g. one annotation layer for each of "default", "split", "merged".  That works okay if you have a small number of classes --- as then you can set a separate color for each layer, and rely on the existing layer visibility toggles to choose what to display, and your DVIDAnnotationSource could have an additional parameter that excludes all but a particular class.  That would also allow you to add annotations to a particular class --- as currently there would be no way to do that.  However, that solution may not be ideal for your use case, and with the additional "checked" bit it might be awkward to have 6 layers.

tingzhao:
Yes, I understand the fragment shader part and actually played around with it in one of our hack branches. But I haven't found a way to write a new shader for DVID annotations without having to change the original code.

Using layers sounds like a good idea. But the number of classes might be more than what we have now in the future. BTW, if I use multiple layers for the same data, will the data be downloaded multiple times, or the chunks will be shared across the layers? What if I want to refresh annotations to get changes made by other programs? Currently I have to delete the old layer and create a new one. There must be a better way for that, but I haven't figured out how.

jbms:
Regarding receiving updated data: this is kind of supported through Python
datasources, where it is possible from Python to invalidate the cache and
force all of the chunks to be reloaded.  In the ideal case, DVID itself
would support some protocol by which the client can subscribe to be
notified when something changes.  That might take some work to integrate,
though, since none of the existing data sources support such a thing.

Alternatively, if you want to control it externally (e.g. from the react
integration), you could just change the datasource URL (e.g. by adding some
parameter that is ignored) to force the data to be reloaded.  This would
have the downside that the existing data would not be displayed while it is
reloading.  There is also an existing "invalidateCache" mechanism (see
chunk_manager/frontend.ts) used by the Python datasource, which currently
has the same limitation that existing data is dropped from the cache before
the new data is downloaded, but it could be improved so that it doesn't do
that.

jbms:
As far as sharing the data for multiple layers --- the chunk data cache is
not directly tied to layers, so you can certainly share data.  However, for
the separate layers approach you would need the backend.ts code to perform
the filtering, i.e. creating annotation chunks that contain just the
annotations in a given class.  To download all classes as a single request,
and then split that into separate per-class chunks, you can use
GenericSharedDataSource, e.g. as in datasource/precomputed/backend.ts or
datasource/nifti/backend.ts.  That allows you to cache the result of
performing some request and use it to obtain the data for multiple other
chunks, in a way that integrates with the Neuroglancer caching system.

Regarding adding a new annotation type: taking another look at the existing
code I see that does involve some complications.  Probably the simplest and
most general thing would be to add support to the POINT annotation for
multiple markers and per-point colors, rather than just circles.  It sounds
like that would both solve your use case and would also be useful generally.

tingzhao:
To support per-point colors, do I just add a color property to the Point class like
`export interface Point extends AnnotationBase {
  point: vec3;
  type: AnnotationType.POINT;
  color: vec3;
}`?
This seems conflict with the global color control widget.

jbms:
You could represent the color with 4 bytes, the first 3 indicating the color as RGB, the remaining one set to 0 to indicate that the global color should be used, or otherwise set to 1.

jbms:
To follow up on this, in order to support this and other use cases, I'm working on adding support for allowing an annotation source to specify an arbitrary set of additional properties that will be associated with each annotation, and allow custom shader code to determine such things as the color, size, etc. based on those properties.  Additionally, rather than a single set of associated segment ids, annotation sources  will be permitted to have multiple types of "relationships" (e.g. pre-synaptic, post-synaptic), and a set of segment ids per relationship.

I think that would avoid the need for any custom rendering code for DVID.

tingzhao:
That's great! I'm looking forward to it!

jbms:
It took a bit longer than expected, but I just pushed out the updated annotation support in Neuroglancer, including support for additional user-defined properties, custom shaders, much better spatial indexing support, and support for a precomputed annotation format.

I believe the custom property and custom shader support should address the custom rendering issues you raised in this pull request.

The UI support for user-defined properties is still a bit limited at the moment, though, and does not provide a way to modify them.

Let me know if you have any questions or issues.

tingzhao:
Thank you! I'll have a try.

tingzhao:
@jbms Do you have any instructions about defining a custom shader for the new properties field in AnnotationBase, or any example in the current code? (Sorry to bother you about that. It is hard for me to find time to figure it out by myself as we're rushing to the data release.)

jbms:
There is documentation here:
https://github.com/google/neuroglancer/blob/master/src/neuroglancer/annotation/rendering.md

Suppose there is a float32 property score:

```glsl
#uicontrol vec3 preColor color(default="red")
#uicontrol vec3 postColor color(default="blue")
#uicontrol float scorethr slider(min=0, max=1000)

void main() {
  setColor(defaultColor());
  setEndpointMarkerColor(
    vec4(preColor, 0.5),
    vec4(postColor, 0.5));
  setEndpointMarkerSize(5.0, 5.0);
  setLineWidth(2.0);
  if (prop_score()<scorethr) discard;
}
```

tingzhao:
Thanks! Where should I specify the property name such as prop_score? I saw that AnnotationPropertySpec can be used to define the type of a property, but couldn't find how to give a name to a property so that it can be used in the shader.

jbms:
You pass the list of properties to the constructor of MultiscaleAnnotationSource:

https://github.com/google/neuroglancer/blob/fed15a96fb2e0b91670cd10fa28923f2a0f33a8a/src/neuroglancer/datasource/precomputed/frontend.ts#L252

tingzhao:
I got stuck at 
https://github.com/google/neuroglancer/blob/fed15a96fb2e0b91670cd10fa28923f2a0f33a8a/src/neuroglancer/annotation/frontend_source.ts#L239

The DataView constructor rejected to take serializedAnnotations.data as its parameter. So I changed it to:
const dv = new DataView(serializedAnnotations.data.buffer);
It seemed solve the problem.

jbms:
Thanks --- your fix is correct.  I see that there are a few other bugs related to modifying annotations that I'm working on fixing right now --- hopefully will be done in a few hours.

tingzhao:
Thanks! I'm using AnnotationGeometryChunkSource with only one resolution level for handling point annotations in DVID. One rendering behavior confusing me is that when I zoom out the slice view, the annotations start disappearing in a certain order until all of them are gone. They will appear again if I zoom in. It's likely caused by something wrong in my code, but if you happen to know or be able to guess the reason, it will be great!

jbms:
Regarding the zooming out issue, there are a few things going on:

When using the spatial index, Neuroglancer automatically displays only a subsample of the annotations in order to limit the maximum density of displayed annotations per screen pixel.  This happens for both the cross-section and 3d projection views, and is controlled by the "Spacing" setting in the "Annotations" tab.  The AnnotationGeometryChunkSource specifies a `limit` value in the AnnotationGeometryChunkSpecification.  If `limit` is set to too high a value, though, subsampling will occur prematurely.  If you specify a `limit` of 1, that indicates a very low density and Neuroglancer won't perform any subsampling.

In general the way annotations are displayed was designed assuming a multi-level source.  If you only have a single-level source, then as you zoom out in either the cross section or 3d projection view, the view frustrum covers a larger and larger area, and the amount of data required increases until all chunks are required.  Note that even in the cross-section view, as you zoom out the depth range within which annotations are displayed also increases by default, but you can control this by unchecking "Zoom-relative" and/or manually adjusting the depth range.

In general, I'm not sure what the best solution is, though.

jbms:
I pushed a fix for a bunch of bugs and missing functionality related to write support for annotation sources.  Hopefully this makes things work for you.  However, for the only other data source that supports writes, brainmaps, I'm just using a single chunk for the entire volume, so some functionality is not fully tested; let me know if anything is still broken.

You will have to include chunkToMultiscaleTransform in the AnnotationGeometryChunkSpecification.  Also, to disable subsampling, you should set `limit` to 0, not `1` (1 should work most of the time, but 0 is more correct).

tingzhao:
Thank you! It's working well so far.

tingzhao:
@jbms I got another problem while using annotation source for DVID synapses. Since DVID doesn't have multiscale annotation datasource, I faked it by creating low-res sources with empty bounding boxes (lowerbound=upperbound=[0,0,0]) to avoid downloading too much data while zooming out. It worked fine with old versions of neuroglancer, but in the new version, there seem some significant changes in visible chunk computation and the trick no longer works. Neuroglancer will always try to download synapse data at any resolution level. Do you have any idea?

tingzhao:
I saw that https://github.com/google/neuroglancer/blob/4356aba77eeb2b0a7434e176dbef4bdc88f2183b/src/neuroglancer/annotation/renderlayer.ts#L722 uses transformedSources[0] only. Does that mean it only checks the source with the original resolution?

jbms:
In that case, transformedSources is an array of arrays --- the outer array selects between different chunk orientations, e.g. chunks aligned more to xy vs xz vs yz.  That is supported for normal volumetric data, but not yet supported for annotations (and not necessarily all that useful anyway).  The inner array is the array of different scales.

As far as how to avoid loading too much data, I think you could make that work by creating a fake scale with non-zero chunkDataSize and a non-zero limit value.  You need to pick the values such that the density is high enough that Neuroglancer doesn't attempt to load the next lower scale (the real scale) when you are sufficiently zoomed out.  Density is computed as limit / prod(chunkDataSize), so increase limit or decrease chunkDataSize to increase the density (and therefore make Neuroglancer less likely to load the real data).

jbms:
Note: The "Spacing" controls in the "Annotations" tab may be helpful to guide your choice of limit.  Also by adjusting the spacing value you can also control when the real data is loaded.

tingzhao:
Thank you! That's what I need!

PR #183, opened 2019-12-10
feat: Makes display of item value at cursor position in layer panel o…

moenigin:
…ptional

### DIFF
--- a/python/neuroglancer/viewer_config_state.py
+++ b/python/neuroglancer/viewer_config_state.py
@@ -151,7 +151,7 @@ class ConfigState(JsonObjectWrapper):
     show_panel_borders = showPanelBorders = wrapped_property('showPanelBorders',
                                                              optional(bool, True))
     scale_bar_options = scaleBarOptions = wrapped_property('scaleBarOptions', ScaleBarOptions)
-    display_layer_item_value = displayLayerItemValue = wrapped_property('displayLayerItemValue', optional(bool, True))
+    show_layer_hover_values = showLayerHoverValues = wrapped_property('showLayerHoverValues', optional(bool, True))
     viewer_size = viewerSize = wrapped_property('viewerSize', optional(array_wrapper(np.int64, 2)))
     prefetch = wrapped_property('prefetch', typed_list(PrefetchState))
 
--- a/python/neuroglancer/viewer_config_state.py
+++ b/python/neuroglancer/viewer_config_state.py
@@ -151,6 +151,7 @@ class ConfigState(JsonObjectWrapper):
     show_panel_borders = showPanelBorders = wrapped_property('showPanelBorders',
                                                              optional(bool, True))
     scale_bar_options = scaleBarOptions = wrapped_property('scaleBarOptions', ScaleBarOptions)
+    display_layer_item_value = displayLayerItemValue = wrapped_property('displayLayerItemValue', optional(bool, True))
     viewer_size = viewerSize = wrapped_property('viewerSize', optional(array_wrapper(np.int64, 2)))
     prefetch = wrapped_property('prefetch', typed_list(PrefetchState))
 
--- a/src/neuroglancer/layer_panel.ts
+++ b/src/neuroglancer/layer_panel.ts
@@ -31,6 +31,7 @@ import {float32ToString} from 'neuroglancer/util/float32_to_string';
 import {makeCloseButton} from 'neuroglancer/widget/close_button';
 import {makeIcon} from 'neuroglancer/widget/icon';
 import {PositionWidget} from 'neuroglancer/widget/position_widget';
+import {WatchableValueInterface} from 'neuroglancer/trackable_value';
 
 function destroyDropLayers(dropLayers: DropLayers, targetLayer?: ManagedUserLayer) {
   if (dropLayers.method === 'move') {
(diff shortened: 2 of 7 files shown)


src/neuroglancer/layer_panel.ts
```diff
@@ -354,31 +355,33 @@ export class LayerPanel extends RefCounted {
...
-              return 'null';
-            } else if (Math.fround(x) === x) {
-              // FIXME: Verify actual layer data type
-              return float32ToString(x);
-            } else {
-              return x;
-            }
-          });
-          text += value.join(', ');
+    if (this.displayLayerItemValue.value === true){
```
jbms: You could simplify this nesting by just returning early if the value is false.  Also I think we still need to call updateLayers() in case the layer names changed.

python/neuroglancer/viewer_config_state.py
```diff
@@ -151,6 +151,7 @@ class ConfigState(JsonObjectWrapper):
     show_panel_borders = showPanelBorders = wrapped_property('showPanelBorders',
                                                              optional(bool, True))
     scale_bar_options = scaleBarOptions = wrapped_property('scaleBarOptions', ScaleBarOptions)
+    display_layer_item_value = displayLayerItemValue = wrapped_property('displayLayerItemValue', optional(bool, True))
```
jbms: How about renaming this setting to show_layer_hover_values / showLayerHoverValues ?

src/neuroglancer/layer_panel.ts
jbms: Also, if you update an attribute of the layer panel element, e.g. this.element.dataset.showHoverValues = this.displayLayerItemValue.value.toString() , then you can use a CSS rule to hide that area of the layer widget.

src/neuroglancer/layer_panel.ts
```diff
@@ -258,8 +258,9 @@ export class LayerPanel extends RefCounted {
   constructor(
       public display: DisplayContext, public manager: LayerListSpecification,
       public viewerNavigationState: LinkedViewerNavigationState,
-      public selectedLayer: SelectedLayerState, public getLayoutSpecForDrag: () => any) {
-    super();
+      public selectedLayer: SelectedLayerState, public getLayoutSpecForDrag: () => any,
+      public displayLayerItemValue: WatchableValueInterface<boolean>) {
+    super(); 
     this.registerDisposer(selectedLayer);
```
jbms: Use e.g.:

this.registerDisposer(displayLayerItemValue.changed.add(() => this.handleLayerValuesChanged()));

so that the display updates as soon as the config setting changes.

jbms:
Thanks for this change --- overall it looks great, just a few comments.

python/neuroglancer/viewer_config_state.py
moenigin: Yes, it is easier to understand what this is good for. I changed it, but now it is somewhat inconsistent with the naming in the layer_panel.css ...

src/neuroglancer/layer_panel.ts
moenigin: Done.

src/neuroglancer/layer_panel.ts
moenigin: > Also, if you update an attribute of the layer panel element, e.g. this.element.dataset.showHoverValues = this.displayLayerItemValue.value.toString() , then you can use a CSS rule to hide that area of the layer widget.

I am not sure whether I properly understand what you are suggesting. Would you like to handle the entire hiding of the value display by switching between css classes? In this case the whole loop in update() would still be executed when hovering across the segementation which seems inefficient or at least unnecessary. Thanks!

python/neuroglancer/viewer_config_state.py
jbms: My original naming of the CSS classes was not so good.  If you want to fix that as well, that would be great, but I don't think it is that important.

src/neuroglancer/layer_panel.ts
```diff
@@ -415,8 +425,24 @@ export class LayerPanel extends RefCounted {
...
   addLayerMenu() {
     addNewLayer(this.manager, this.selectedLayer);
   }
+
+  updateItemValueDisplay() {
+    if (!this.valueElementUpdateNeeded){
+      return;
+    }
+    this.valueElementUpdateNeeded = false;
+    console.log(this.layerWidgets)
```
jbms: You can actually just set this on the parent element --- no need to set it on each individual widget element.   Then you can use a CSS selector like:

`.neuroglancer-layer-panel.hidden .neuroglancer-layer-item-value`

or if you used: this.element.dataset.hoverValuesHidden = "true";

.neuroglancer-layer-panel[data-hover-values-hidden="true"] .neuroglancer-layer-item-value

Also, please rename to neuroglancer-hidden, or use an attribute --- because class names are global, using an unprefixed name like "hidden" could result in unintended styling if Neuroglancer is embedded in another application that uses that class name for a different purpose.

src/neuroglancer/layer_panel.ts
jbms: I was suggesting to keep your current logic of not updating the values if they are hidden, but additionally use the CSS rules so that no space is taken up --- exactly as you have done.

python/neuroglancer/viewer_config_state.py
moenigin: Since I cannot really tell where your naming is not so good, I'd better keep my hands off this...

src/neuroglancer/layer_panel.ts
moenigin: Thanks a lot for your explanations and patience!

jbms:
Thanks!

PR #190, opened 2019-12-19
feat: @type 'neuroglancer_legacy_mesh' triggers legacy mesh format

william-silversmith:
Currently, neuroglancer treats missing mesh info files as legacy
format, but we found that it is necessary to add some attributes
to mesh info files in order to deduplicate chunk boundaries. However,
this leaves us with no way to read legacy formats. Thus, we add
neuroglancer_legacy_mesh as an explicit type to enable continued use.

### DIFF
--- a/src/neuroglancer/datasource/precomputed/README.md
+++ b/src/neuroglancer/datasource/precomputed/README.md
@@ -171,7 +171,7 @@ The actual storage of the manifest and mesh fragment data depends on whether the
 
 The `info` file is a JSON-format text file.  The root value must be a JSON object with the following
 members:
-- `"@type"`: Must be `"neuroglancer_multilod_draco"`.
+- `"@type"`: Must be `"neuroglancer_multilod_draco"` or `"neuroglancer_legacy_mesh"`.
 - `"vertex_quantization_bits"`: Specifies the number of bits needed to represent each vertex
   position coordinate within a mesh fragment.  Must be `10` or `16`.
 - `"transform"`: JSON array of 12 numbers specifying a 4x3 homogeneous coordinate transform from the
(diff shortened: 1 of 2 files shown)


william-silversmith:
Google storage returns 403 when a file is missing, so I added it as an or condition with 404.

src/neuroglancer/datasource/precomputed/README.md
```diff
@@ -171,7 +171,7 @@ The actual storage of the manifest and mesh fragment data depends on whether the
 
 The `info` file is a JSON-format text file.  The root value must be a JSON object with the following
 members:
-- `"@type"`: Must be `"neuroglancer_multilod_draco"`.
+- `"@type"`: Must be `"neuroglancer_multilod_draco"` or `"neuroglancer_legacy_mesh"`.
```
jbms: This shouldn't be mentioned here, since this documents the multiscale mesh format.

william-silversmith: Oops! Done.

src/neuroglancer/datasource/precomputed/frontend.ts
```diff
@@ -260,7 +263,7 @@ async function getMeshMetadata(
   try {
     metadata = await getJsonMetadata(chunkManager, url);
   } catch (e) {
-    if (e instanceof HttpError && e.status === 404) {
+    if (e instanceof HttpError && (e.status === 404 || e.status === 403)) {
```
jbms: Just curious, what sort of serving setup gives 403?

jbms: Regarding Google storage returning 403, does that happen when you grant access only via per-object ACLs, rather than per-bucket IAM permissions?

william-silversmith: "This bucket uses fine-grained access control, allowing you to specify access to individual objects. To control access uniformly at the bucket level, switch to uniform access control."

Honestly, it might be worth switching to uniform. The option didn't exist when we created some of them.

jbms: You don't have to switch --- the bucket-level IAM permissions apply even if you don't switch the policy.  A user is granted access to an object if either the bucket-level IAM permissions or object-level ACL grants access.  I recommend not switching to uniform at the moment if you benefit from HTTP cache control headers, because there is currently a limitation in GCS such that switching to uniform access control also prevents caching.

jbms:
Thanks!

Issue #185, opened 2019-12-13
SWC not parsed properly?


tingzhao:
I spot two potential problems in  decode_swc_skeletons.ts:

1. Function `decodeSwcSkeletonChunk` 
```
    glVertices[3 * i] = swc_obj.z;
    glVertices[3 * i + 1] = swc_obj.y;
    glVertices[3 * i + 2] = swc_obj.x;
```
I think it should be in (x, y, z) order instead of (z, y, x).

2. Function `parseSwc`

The conversion from 1-indexed to 0-indexed is a little ad hoc. I don't think there's any official standard to force SWC nodes to be 1-indexed, although many people use 1-indexed. This is broken in our application, however, since we're using 0-indexed in DVID.  In fact, there's no standard to state that the IDs should be contiguous, but this is more assumed by many applications than the indexing part.

tingzhao:
@jbms I'll create a pull request for fixing them if you confirm the problems.

jbms:
Sorry for the delay in responding.  This SWC code is actually only used for DVID and I believe was written by FlyEM, so you should feel free to fix it in whatever way you think makes sense.

PR #197, opened 2020-02-05
fix(decode_swc_skeleton): fix several problems in swc parsing

tingzhao:
This PR is for fixing issue #185. It fixes the parsing order of skeleton coordinates, and handles multiple roots and non-contiguous IDs. It also adds missing license comments.

image


jbms:
Thanks

PR #212, opened 2020-04-03
Add SWC support.

manoaman:
Hi jbms,

I am making a pull request for SWC support code.  If you can point me out what is missing, that would be great.

Thanks so much!
manoaman


src/neuroglancer/datasource/swc/frontend.ts
```diff
@@ -0,0 +1,542 @@
...
+            default: true,
+            subsource: {
+              mesh: options.chunkManager.getChunkSource(SWCSkeletonSource, {
+                parameters: {
+                  'baseUrl': baseUrl,
+                  'nodeKey': nodeKey,
+                  'dataInstanceKey': info.skeletonSrc,
+                }
+              })
+            },
```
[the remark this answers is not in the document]
manoaman: Does `getDataInstanceDetails` need to be replaced to prepare a `info` object?

```
        const box: BoundingBox = {
          lowerBounds: new Float64Array(info.lowerVoxelBound),
          upperBounds: Float64Array.from(info.upperVoxelBoundInclusive, x => x + 1)
        };
        const modelSpace = makeCoordinateSpace({
          rank: 3,
          names: ['x', 'y', 'z'],
          units: ['m', 'm', 'm'],
          scales: Float64Array.from(info.voxelSize, x => x / 1e9),
          boundingBoxes: [makeIdentityTransformedBoundingBox(box)],
        });
        const dataSource: DataSource = {
          modelTransform: makeIdentityTransform(modelSpace),
          subsources: [{
            id: 'default',
            subsource: {volume},
            default: true,
          }],
        };
```

src/neuroglancer/datasource/swc/backend.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@registerSharedObject() export class SWCSkeletonSource extends
+(WithParameters(SkeletonSource, SWCSourceParameters)) {
+  download(chunk: SkeletonChunk, cancellationToken: CancellationToken) {
+    const {parameters} = this;
+    let bodyid = `${chunk.objectId}`;
+    // Change the url pattern for SWC
+    // const url = `${parameters.baseUrl}/api/node/${parameters['nodeKey']}` +
+    //     `/${parameters['dataInstanceKey']}/key/` + bodyid + '_swc';
+    const url = `${parameters.baseUrl}/swc/` + bodyid ;
```
jbms, line 33: I recommend that you support data source URLS like:

"https://whatever/path/{id}.swc"  and then here you would replace {id} with the actual id converted to a string, e.g. via String.prototype.replace.

src/neuroglancer/datasource/swc/frontend.ts
```diff
@@ -0,0 +1,551 @@
...
+//     verifyObjectProperty(obj, 'TypeName', verifyString);
+//   }
+// }
+
+// export class DataInstanceInfo {
+//   constructor(public obj: any, public name: string, public base: DataInstanceBaseInfo) {}
+// }
+
+// class SWCVolumeChunkSource extends
+// (WithParameters(VolumeChunkSource, VolumeChunkSourceParameters)) {}
```
jbms, line 66: Delete

src/neuroglancer/datasource/swc/frontend.ts
```diff
@@ -0,0 +1,551 @@
...
+      //     }).map(spec => ({
+      //              chunkSource: chunkManager.getChunkSource(
+      //                  SWCVolumeChunkSource, {spec, parameters: volParameters}),
+      //              chunkToMultiscaleTransform,
+      //            }));
+      // sources.push(alternatives);
+    }
+    return transposeNestedArrays(sources);
+  }
+}
```
jbms, line 204: Delete

src/neuroglancer/datasource/swc/frontend.ts
```diff
@@ -0,0 +1,551 @@
...
+//   getSources(volumeSourceOptions: VolumeSourceOptions) {
+//     return this.info.getSources(
+//         this.chunkManager, {
+//           'baseUrl': this.baseUrl,
+//           'nodeKey': this.nodeKey,
+//           'dataInstanceKey': this.dataInstanceKey,
+//         },
+//         volumeSourceOptions);
+//   }
+// }
```
jbms, line 393: Delete

src/neuroglancer/datasource/swc/frontend.ts
```diff
@@ -0,0 +1,551 @@
...
+        // const box: BoundingBox = {
+        //   lowerBounds: new Float64Array(info.lowerVoxelBound),
+        //   upperBounds: Float64Array.from(info.upperVoxelBoundInclusive, x => x + 1)
+        // };
+        const modelSpace = makeCoordinateSpace({
+          rank: 3,
+          names: ['x', 'y', 'z'],
+          units: ['m', 'm', 'm'],
+          // scales: Float64Array.from(info.voxelSize, x => x / 1e9), ???
+          scales: Float64Array.from(vec3.create(), x => x / 1e9),
```
jbms, line 435: Since vec3.create() returns [0, 0, 0], then ends up being [0, 0, 0], which is invalid and the reason everything was displaying as a single point.  Instead, this should be Float64Array.of(1e-6, 1e-6, 1e-6) if you want to go by the spec here (http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html) and assume micrometers.

Some SWC files are in different units, but it doesn't really matter if this default is wrong because you can override by specifying a transform (e.g. in the UI).

src/neuroglancer/datasource/swc/frontend.ts
```diff
@@ -0,0 +1,551 @@
...
+          // boundingBoxes: [makeIdentityTransformedBoundingBox(box)], ???
+          boundingBoxes: [],
+        });
+        const dataSource: DataSource = {
+          modelTransform: makeIdentityTransform(modelSpace),
+          subsources: [{
+            id: 'default',
+            // subsource: {volume},
+            subsource: {},
+            default: true,
```
jbms, line 445: Delete this empty subsource.  Instead, the skeletons one added below should be given an id of "default".

PR #214, opened 2020-04-16
fix(datasource/dvid): remove unnecessary calls of getting dvid info

tingzhao:
This PR cleans up some code in dvid/frontend to remove unnecessary requests of getting data instance info from node level. Such information can be retrieved from previous repo calls. The extra requests were added to handle a bug in DVID that causes missing information at the repo level. Now the bug has been fixed and it's no longer necessary. The fix does not only improve the performance a little bit, but also facilitate future improvement which I will send through more PRs.

### DIFF
--- a/src/neuroglancer/datasource/dvid/frontend.ts
+++ b/src/neuroglancer/datasource/dvid/frontend.ts
@@ -57,6 +57,12 @@ export class DataInstanceBaseInfo {
 }
 
 export class DataInstanceInfo {
+  lowerVoxelBound: vec3;
+  upperVoxelBoundInclusive: vec3;
+  voxelSize: vec3;
+  blockSize: vec3;
+  numLevels: number;
+
   constructor(public obj: any, public name: string, public base: DataInstanceBaseInfo) {}
 }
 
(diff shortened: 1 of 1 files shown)


src/neuroglancer/datasource/dvid/frontend.ts
```diff
@@ -357,7 +367,7 @@ export function getDataInstanceDetails(
         return info;
       });
 }
-
+*/
```
jbms, line 370: Just delete this block if no longer needed.

src/neuroglancer/datasource/dvid/frontend.ts
```diff
@@ -57,6 +57,12 @@ export class DataInstanceBaseInfo {
 }
 
 export class DataInstanceInfo {
+  lowerVoxelBound: vec3;
+  upperVoxelBoundInclusive: vec3;
+  voxelSize: vec3;
+  blockSize: vec3;
+  numLevels: number;
+
```
jbms, line 65: Why are these moved to VolumeDataInstanceInfo?

jbms:
Thanks!

PR #218, opened 2020-05-06
docs: minor typos in meshes.md

perlman:


jbms:
Thanks

perlman:
@jbms: There is (at least) one more small fix needed. There is a missing parenthesis in the section on 'Multi-resolution mesh fragment data file format':

> Each position component `j`
> must be in the range `[0, 2**vertex_quantization_bits)`, where a value of `x` corresponds to
> `grid_origin[i] + (fragmentPosition[i] + x / (2**vertex_quantization_bits-1) * (2**lod)`.  The
> built-in Draco attribute quantization is not supported.

Adding the missing ')' is a choose your own adventure with quite a range of values.  ;-)

jbms:
Thanks for pointing that out, I fixed that also now in  9f475da0b2b061dbc6bea00794c7429eddf6b0e9 --- it was also missing a `chunk_shape[i] *`.

PR #221, opened 2020-05-29
Remove tornado<6 dep constraint

chrisroat:
The following fix should make sockjs-tornado compatible with tornado6
https://github.com/mrjoes/sockjs-tornado/pull/118

### DIFF
--- a/python/neuroglancer/server.py
+++ b/python/neuroglancer/server.py
@@ -33,7 +33,7 @@ import sockjs.tornado
 try:
     # Newer versions of tornado do not have the asynchronous decorator
     from sockjs.tornado.util import asynchronous
-except (ImportError, ModuleNotFoundError):
+except ImportError:
     from tornado.web import asynchronous
 
 from . import local_volume, static
--- a/python/neuroglancer/server.py
+++ b/python/neuroglancer/server.py
@@ -29,8 +29,12 @@ import tornado.netutil
 import tornado.web
 
 import sockjs.tornado
-from sockjs.tornado.util import asynchronous
 
+try:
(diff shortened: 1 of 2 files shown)


chrisroat:
@googlebot I signed it!

python/neuroglancer/server.py
```diff
@@ -29,6 +29,7 @@
 import tornado.web
 
 import sockjs.tornado
+from sockjs.tornado.util import asynchronous
```
jbms, line 32: Can you change this to conditionally import either this definition or the one from tornado.web if available, so that this continues to work with old versions of tornado and sockjs.tornado as well?

chrisroat: I assume that means you want me to drop the dependency constraint on sockjs-tornado.   Will there be problems with someone using the newest sockjs, but an older tornado (or vice versa)?

jbms: I think it is alright --- pip doesn't really strictly enforce dependency constraints anyway.

python/neuroglancer/server.py
```diff
@@ -30,6 +30,11 @@
 
 import sockjs.tornado
 
+try:
+    # Newer versions of tornado do not have the asynchronous decorator
+    from sockjs.tornado.util import asynchronous
+except (ImportError, ModuleNotFoundError):
```
jbms, line 36: ModuleNotFoundError is a subclass of ImportError, so there is no need to include it.  Additionally, it doesn't exist on Python2, which Neuroglancer still suports.

chrisroat:
Is this patch still of interest?

jbms:
Thanks!

PR #224, opened 2020-06-02
html colors for selected segments in Python API

austinhoag:
This feature adds a property: "segment_html_color_dict" to segmentation
layers via the Python API. The property returns a dictionary where
keys are segment ids and values are the html color codes,
e.g. #F9F607 (yellow) corresponding to each currently selected segment.
The colors are the same as what is rendered in the viewer for the segments,
given a color seed. This property uses the current color seed when
calculating the color codes.

### DIFF
--- a/python/neuroglancer/segment_colors.py
+++ b/python/neuroglancer/segment_colors.py
@@ -14,10 +14,6 @@
 
 import math
 
-def zero_fill_right_shift(val, n):
-    """ Python implementation of javascript's >>> operator """
-    return (val >> n) if val >= 0 else ((val + 0x100000000) >> n)
-
 def hash_function(state,value):
     """ Python implementation of hashCombine() function
     in src/neuroglancer/gpu_hash/hash_function.ts,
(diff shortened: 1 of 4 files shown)


austinhoag:
@googlebot I signed it!

python/neuroglancer/segment_colors.py
```diff
@@ -0,0 +1,85 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+
+def zero_fill_right_shift(val, n):
```
jbms, line 17: Since the hash function won't receive any negative numbers, you can just use regular right shift `val >> n` in Python for this purpose.

jbms:
If you can add a couple unit tests (e.g. just a couple examples where you verify that the result matches values manually obtained from the javascript code), that would be great.

austinhoag:
Sure, I have added a few tests and made sure they passed.

jbms:
Thanks!

PR #229, opened 2020-07-08
n5: work with AWS (403 instead of 404)

perlman:
Public on AWS with GetObject but not ListObject permissions will get a 403 instead of 404 when a files does not exist.

This prevents access to n5 volumes with a fatal error when the first missing `attributes.json` gets a 403 response.

(This is from debugging access to a s3 hosted n5 volume with @sharmishtaa).


jbms:
Thanks.

Issue #245, opened 2020-09-15
#uicontrol sliders not rendering properly under FireFox


perlman:
The range part of the slider is not currently showing up under FireFox 80 on MacOS. This is purely cosmetic. The controls continue to function.

[...]

PR #246, opened 2020-09-16
Make range track visible on FireFox 80/81 on MacOS.

perlman:
A possible fix for https://github.com/google/neuroglancer/issues/245 .

The sliders now look like this with FireFox 80 on MacOS dark mode:
image

These changes appear to have no effect on MacOS Chrome. I have not tested on Windows or Linux.


jbms:
Thanks!

PR #247, opened 2020-09-22
Include the right annotation id for error handling

tingzhao:
In [sendCommitRequest](https://github.com/google/neuroglancer/blob/2e4a2c18705de6dc3a6e9cfbc441bac4230c82cb/src/neuroglancer/annotation/frontend_source.ts#L474), `annotationId` is undefined in the case of adding an annotation, causing an undefined `annotationId` in the message for the error handling. This can keep [handleFailedUpdate](https://github.com/google/neuroglancer/blob/2e4a2c18705de6dc3a6e9cfbc441bac4230c82cb/src/neuroglancer/annotation/frontend_source.ts#L680) in annotation/frontend_source.ts from  showing the right error message because `localUpdate` will always be undefined in this case. As a result, the status message will get stuck at 'Committing annotations' when adding an annotation fails.

This pr tries to fix this problem by including the new annotation id in the error message when the update is not on an existing annotation, just like what the resolve function does, except that I expect both `annotationId` and `newAnnotation` can be undefined or null in a failed update (maybe?).

### DIFF
--- a/src/neuroglancer/annotation/backend.ts
+++ b/src/neuroglancer/annotation/backend.ts
@@ -253,9 +253,11 @@ registerRPC(ANNOTATION_COMMIT_UPDATE_RPC_ID, function(x: any) {
       },
       (error: Error) => {
         if (!obj.wasDisposed) {
-          this.invoke(
-              ANNOTATION_COMMIT_UPDATE_RESULT_RPC_ID,
-              {id: obj.rpcId, annotationId, error: error.message});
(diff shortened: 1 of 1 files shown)


jbms:
Thanks for reporting this.  It appears to me that if `annotationId` is undefined, then `newAnnotation` is guaranteed to be defined.  Neither of those values depends on the result of the update.  I think it is safe, therefore, to use `annotationId ?? newAnnotation!.id` (using the new `??` null coalescing operator supported by Typescript, which is similar to but slightly cleaner than `||`).

PR #256, opened 2020-10-20
Use relative paths for python data source.

perlman:
As @chrisroat noticed in the comment thread of #150, local data sources are no longer working through [jupyter-server-proxy](https://github.com/jupyterhub/jupyter-server-proxy).  The problem stems from the root-relative URLs.

An example URL for neuroglancer with jupyter-server-proxy would be `http://localhost:8888/proxy/58060/v/f9dd9e35803d93ee67997951bf4a138a9abb7983/`

This PR is a quick & dirty fix illustrating the problem. I'm not sure what a proper @jbms solution would be. 🙂

I know this worked ~1-2 years ago. Were the data URLs previously under the token?


jbms:
Thanks!

PR #287, opened 2021-03-04
feat(datasource/middleauth): add MiddleAuth credentials support

chrisj:
still working out details

### DIFF
--- a/src/neuroglancer/ui/url_hash_binding.ts
+++ b/src/neuroglancer/ui/url_hash_binding.ts
@@ -98,7 +98,7 @@ export class UrlHashBinding extends RefCounted {
         s = '#!{}';
       }
       // Handle remote JSON state
-      if (s.match(/^#!(http|https|gs|middleauth\+https):\/\//)) {
+      if (s.match(/^#!([a-z][a-z\d+-.]*):\/\//)) {
         const url = s.substring(2);
         const {url: parsedUrl, credentialsProvider} = parseSpecialUrl(url, this.credentialsManager);
         StatusMessage.forPromise(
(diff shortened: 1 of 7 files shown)


src/neuroglancer/ui/url_hash_binding.ts
```diff
@@ -98,7 +98,7 @@ export class UrlHashBinding extends RefCounted {
         s = '#!{}';
       }
       // Handle remote JSON state
-      if (s.match(/^#!(http|https|gs):\/\//)) {
+      if (s.match(/^#!(http|https|gs|middleauth\+https):\/\//)) {
```
jbms, line 101: Probably reasonable to just match anything that looks like a protocol -- that way we don't need the list of special protocols duplicated here.  I think I also forgot to properly support ngauth here as well...

src/neuroglancer/util/special_protocol_request.ts
```diff
@@ -67,6 +91,12 @@ export function parseSpecialUrl(url: string, credentialsManager: CredentialsMana
         credentialsProvider: getNgauthCredentialsProvider(credentialsManager, `https://${u.host}`, u.path),
         url: 'gs+xml:/' + u.path,
       };
+      case 'middleauth+https':
```
jbms, line 94: Now that there are a decent number of protocols, it might be nicer to add a registry mechanism so that they don't all have to be defined in this file.  That is also something I could take care of later after merging though.

chrisj:
@googlebot I signed it!

chrisj:
I realized all we need now is to look for an exact origin match for app urls. getMiddleAuthCredentialsProvider was already using `new URL(url).origin`. Thanks for pointing out that vulnerability.

jbms:
This looks good to me --- is it ready to merge or are you still working on it?

chrisj:
It is ready for merge, thanks.

PR #295, opened 2021-04-07
feat: add state saving/sharing ui

chrisj:
This pull request adds a "share" button next to the help button. It allows users to send their state to a server that accepts a json encoded version of the neuroglancer state. the "state server" is expected to return a url that will respond that same json string

The code is currently in a rough state because we currently have this functionality bundled with our code that removes the state from the url. I stripped out that functionality as well as tried to simplify the UI. We would like your thoughts before we polish this further. I believe you mentioned wanting to avoid users accidentally sending their state to a server that they didn't know was linked.

Currently if they click the share button, it immediately sends the request if the neuroglancer already has a "jsonStateServer" specified. Otherwise it allows them to set that state server. If we want to avoid them accidentally sending it to the wrong state, we either need to the state server next to the share button, or make it a two step confirmation. I'm also interested what you think about the overall UI/UX.

You can take a look at the version of this that is running at ngl.flywire.ai in case I stripped out more than what you would prefer.


jbms:
Thanks for submitting this.

I think it might make sense to specify a list of allowed state servers at build time as a map, from short identifier, to state server url.  If there are none in the map, then state saving is disabled.  If there is one, it is always used.  If there is more than one, then there is e.g. a <select> element next to the share button to indicate which one.  The JSON state could specify the default state server via the short identifier (or perhaps full url/some allowed list of aliases also configured at build time, for compatibility with your existing states).

src/neuroglancer/save_state/save_state.ts
```diff
@@ -0,0 +1,157 @@
...
+        content: viewer.jsonStateServer.value,
+        textId: 'neuroglancer-save-state-linkshare',
+        fieldTitle: '',
+        readonly: false,
+        btnName: 'Update',
+        btnTitle: 'Push to state server to get JSON URL.',
+        btnAct: () => {
+          const field =
+              <HTMLInputElement>document.getElementById('neuroglancer-save-state-linkshare');
+          const fieldBtn = <HTMLButtonElement>document.getElementById(
```
jbms, line 83: I think this should be passed in as a parameter.

chrisj:
I forgot to say if you want to try it out, you can use
`middleauth+https://globalv1.flywire-daf.com/nglstate/post` as the state server.

chrisj:
> I think it might make sense to specify a list of allowed state servers at build time as a map, from short identifier, to state server url.

Would this be done via the define argument or should I create a json file and pass that in?

jbms:
Yes, I'd suggest using --define --- some build script could then be used to take it from a file if that is desired.

src/neuroglancer/viewer.ts
```diff
@@ -524,6 +536,47 @@ export class Viewer extends RefCounted implements ViewerState {
       topRow.appendChild(button);
     }
 
+    if (typeof STATE_SERVERS !== 'undefined') {
```
jbms, line 539: Thanks for the changes.  This is looking pretty close.  Can you keep this in a separate file, though, since viewer.ts is already quite large?

chrisj: Yeah I'll do that.

I replaced the dialog with a very basic UI like you suggested.
state-share-ui

The select element only comes up when there is more than one state server option.

It is controlled by a config file like
```json
{
    "test": {
        "url": "middleauth+https://authsl1.middleauth.com/nglstate/post",
    },
    "production": {
        "url": "middleauth+https://global.daf-apis.com/nglstate/post",
        "default": true
    }
}
```
I bundled it using `node ./config/esbuild-cli.js --config=min --no-typecheck --define STATE_SERVERS=$(cat config/state_servers.json | tr -d ” \t\n\r”)`

I had to update esbuild for that define to work in github actions. >= 0.9 seems to work

When you share, it copies it to clipboard.

You can test it out here: https://state-share-2-dot-neuroglancer-dot-seung-lab.appspot.com

jbms: Thanks for these changes.  I've tried it out and it looks pretty good!

chrisj: Thanks, I moved the code back into a separate file. I kept the trackable value in view, not sure if it makes sense to move that into the StateShare object. I also wasn't sure about exporting a `stateShareEnabled` variable so that the viewer would know whether to instantiate the StateShare object.

jbms:
Thanks.  This looks good to me.  Is it ready to merge or are you planning to make any further changes?

chrisj:
> Thanks. This looks good to me. Is it ready to merge or are you planning to make any further changes?

It is ready to merge.

PR #317, opened 2021-06-11
feat(datasource/graphene)

chrisj:
I'm making this pull request now to open up discussion on the direction we should go with this feature. We are open to extensive changes.

The graphene format is powered by the Chunked Graph backend. The primary feature is remapping between root ids and supervoxel ids.

The map from root id to supervoxel id takes a bounding box and allows us to visualize the neurons. The mapping can be changed by using merge and cut operations that the chunked graph backend handles. There are other features that are not yet included in this branch, some of the most significant: multicut (a cut operation that allows setting multiple points the source and sink), time control which allows us to request prior states of the chunked graph, find path which allows you to find the most probable path between two points.

I was not that familiar with the graphene code before I worked on integrating it into the current neuroglancer codebase. This code is in a pretty rough state, hopefully it can be understood but we will explain or fix any areas that are confusing. There are some temporary hacks and there may be some unintended changes due to the rebase.

Explanations of some of the messier parts of the code:
* We include our own version of the draco decoding code, I'm not sure if our meshes are compatible with the existing draco code.
* We modified segmentation_user_layer to be able to do merges, cuts were never implemented. To do this in precomputed, we introduced `rootSegments` in  the  `VisibleSegmentsState`. For our graphene logic, these sets contain the same elements. Graphene has a single mesh manifest per root id while precomputed has a mesh manifest per segment id. I believe this is the primary driver in needing `rootSegments`.
* Annotation and skeleton code in graphene should be ignored for now. We want to take advantage of the precomputed annotation support but I haven't started working on that yet. It is currently commented out and and we may not include it for our first iteration of graphene.
* One of the ugliest things I did was in `ChunkedGraphLayer:filterVisibleSources` `forEachSelectedRootWithLeaves` needs to be able to access the sources so I pull it out during the call to `filterVisibleSources`. The previous version was passed the sources in the constructor's options parameter. I assume there is a better way to handle it.
* Some changes were made to base classes like `disjoint_sets`. We aren't expecting you to accept the majority of these changes but figure out a way to avoid them. We changed the ordering of disjoint_sets because it worked nicely with graphene in that the root id is always the largest id.

I'll add more to this explanation as I probably forgot something important.  I will clean up the git history as well.

`graphene://middleauth+https://prodv1.flywire-daf.com/segmentation/1.0/fly_v26` is a test graphene source that can be played with. It does require authentication


src/neuroglancer/datasource/graphene/backend.ts
```diff
@@ -0,0 +1,664 @@
...
+  shardUrl: string;
+}
+
+interface MinishardIndexSource extends
+    GenericSharedDataSource<string, DecodedMinishardIndex|undefined> {
+  sharding: ShardingParameters;
+  credentialsProvider: SpecialProtocolCredentialsProvider;
+}
+
+function getMinishardIndexDataSource(
```
jbms, line 72: Is this different from the precomputed sharded format?

src/neuroglancer/datasource/graphene/backend.ts
```diff
@@ -0,0 +1,664 @@
...
+        error => {
+          if (error instanceof HttpError && error.status == 404) {
+            chunk.source!.removeChunk(chunk);
+          }
+          Promise.reject(error);
+        });
+  }
+}
+
+async function fetchByUint64(
```
jbms, line 472: If possible it would be best to avoid duplication of code from precomputed/backend.ts

jbms:
Hi Chris, thanks for your work on this, I can see that this is a large task!

jbms:
I have taken a look at this, but will still need to work through the details a bit more.

Regarding draco: the neuroglancer precomputed multiscale mesh format does use draco in a slightly specialized way.  However, it should be relatively easy to add another entry point to neuroglancer_draco.cc that just performs the normal decoding --- that way the draco library itself can be shared.

I have some commits on the nggraph branch that are applicable, as far as the disjoint sets reversal, a "tool binding" mechanism for commands like split and merge, and an abstraction for integrating an agglomeration graph with segmentation_user_layer.  I am still looking into it, though, as it doesn't perfectly fit what is needed here.

chrisj:
Hi Jeremy, I'm going to look into the draco change and I will pull in some people that can better explain various parts.

src/neuroglancer/datasource/graphene/backend.ts
william-silversmith: Hi Jeremy,

I only worked on this from the Python side, but the graphene sharded format is identical to precomputed with the exception that shard names are assigned by bounding box, not by hashing. That, and we use the regular Draco compression. This modification has only been used for meshes, I don't think there are any other differences.

jbms:
Over the past week I have polished and integrated several pending changes that I originally developed for the in-progress "nggraph" segmentation graph service, which I think should help integrate graphene in a cleaner way:

- ability to bind layer-specific "tools" to shift+letter.  This makes it possible to bind e.g. shift+x to merge and shift+s to split and shift+c to toggle whether base segments vs agglomerated segments are shown.  While shift+letter is held down, the tool is "activated", a status message can be shown and the tool can bind its own key/mouse bindings as an overlay.  In addition to the new split and merge tools described below, most existing controls (e.g. opacity, shader ui controls, etc.) can now be bound as tools.  I think this will also work well for annotations in the future, e.g. allowing you to bind a key to tag an annotation with a specific tag.
- SegmentationGraphSource abstraction: the graphene data source can provide a SegmentationGraphSource as an additional subsource.  The existing SegmentationUserLayer then creates a  SegmentationGraphSourceConnection, which allows the source to monitor and manipulate its visibleSegments and segmentEquivalences.  You will surely have to change the interface to add the necessary hooks for graphene, but I hope that the basic abstraction will work for you.
- Support in the disjoint sets data structure for toggling between the normal min-element representative behavior /assuming that the high bit is set to indicate the representative element (i.e. max-element).   This is controlled by the active SegmentationGraphSource.
- Split and merge tools that use the SegmentationGraphSource.  The merge tool would probably work as is for graphene.  The split tool implements "one click" split behavior (basically assuming the graph source maintains a spanning tree).  I'm not sure it would make sense to support this for graphene, and in any case you would want a separate tool for the existing multicut splitting logic.  But it may still serve as a useful example of how to integrate split functionality.
- Temporary visible segment set and equivalence set: these are used by the merge and split tools to display a preview of the split or merge.  You might have to add additional temporary options, but I think this is cleaner than swapping in the separate SupevoxelRenderLayer.
- local://equivalences data source that provides a SegmentationGraphSource corresponding to the existing "equivalences" layer property.  As far as graphene is concerned, this mostly just serves as an example/demo of the split/merge actions I've added.  Note that currently it is not possible to use this in combination with any other graph source --- that would probably be useful in some cases, but would increase the implementation complexity and could be added in the future.

chrisj:
Thanks, I'm excited to integrate these changes into Graphene!

chrisj:
Hi Jeremy, I'm back to working on this. One thing I wasn't sure about was that we create a `ChunkedGraphChunkSource` for viewing the agglomerated segmentation. `SegmentationUserLayerWithGraph` combines the `MultiscaleVolumeChunkSource` and `ChunkedGraphChunkSource` so that both can respond to viewport changes. The `ChunkedGraphChunkSource` loads only the equivalences in the visible chunked graph chunks.

Do you see this as part of the necessary hooks that I need to add to `SegmentationGraphSourceConnection` or do we need to keep our `ChunkedGraphChunkSource`?

jbms:
I think you should keep the ChunkedGraphChunkSource, but make it available via the SegmentationGraphSource.

I also think SegmentationUserLayerWithGraph should just be merged into SegmentationUserLayer.

chrisj:
thanks, that makes sense!

PR #320, opened 2021-06-21
http_path_completion: Loosens check for content-type

Tomaz-Vieira:
This patch allows the Content-Type header to have any case
and also allows charset specification, like so:

`Content-type: text/html; charset=UTF-8`

which wasn't allowed before

PR #322, opened 2021-06-22
Allows query params in precomputed chunks

### LINKED ISSUE #316: Allow query strings for precomputed chunks URLs
Currently, neuroglancer will error out on datasets of the kind `precomputed://http://` if they contains query parameters. This is particularly annoying when serving data from Opentsack Swift's temporary URLs, since those come with signatures in the query parameters.

It seems natural to allow URLs to be derived from the base URL as usual, but also appending the query params to the string, but I haven't given too much thought on what consequences that could have.

Tomaz-Vieira:
- Creates the `ParsedUrl` class to streamline handling URLs 
 - uses `ParsedUrl` everywhere the PrecomputedChunks datasource composes URLs, concatenating paths and preserving query strings
 
Fixes  #316 

Needs #320

TODO:
- Wait for legal to give me the green light on licensing/CLA and for #320 to be merged
- Get some architectural feedback and comments =)
- Some more testing

### DIFF
--- a/src/neuroglancer/util/http_request.ts
+++ b/src/neuroglancer/util/http_request.ts
@@ -120,13 +120,77 @@ export function getByteRangeHeader(startOffset: Uint64|number, endOffset: Uint64
   return {'Range': `bytes=${startOffset}-${endOffsetStr}`};
 }
 
-export function parseUrl(url: string): {protocol: string, host: string, path: string} {
-  const urlProtocolPattern = /^([^:\/]+):\/\/([^\/]+)((?:\/.*)?)$/;
-  let match = url.match(urlProtocolPattern);
-  if (match === null) {
-    throw new Error(`Invalid URL: ${JSON.stringify(url)}`);
+export class ParsedUrl{
+  public readonly protocol: string;
+  public readonly host: string;
+  public readonly path: string;
+  public readonly search: string;
+  public readonly hash: string;
+  public readonly href: string;
+
+  public static readonly pattern = /^(?<protocol>[^:]+):\/\/(?<host>[^/]+)(?<path>\/[^?]*)(?<search>\?[^#]*)?(?<hash>#.*)?$/;
+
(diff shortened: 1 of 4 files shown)


src/neuroglancer/util/http_request.ts
```diff
@@ -120,13 +120,77 @@ export function getByteRangeHeader(startOffset: Uint64|number, endOffset: Uint64
   return {'Range': `bytes=${startOffset}-${endOffsetStr}`};
 }
 
-export function parseUrl(url: string): {protocol: string, host: string, path: string} {
-  const urlProtocolPattern = /^([^:\/]+):\/\/([^\/]+)((?:\/.*)?)$/;
-  let match = url.match(urlProtocolPattern);
-  if (match === null) {
-    throw new Error(`Invalid URL: ${JSON.stringify(url)}`);
+export class ParsedUrl{
```
jbms, line 123: Can we use the existing URL class already provided by browsers, or is it unsuitable for some reason?

https://developer.mozilla.org/en-US/docs/Web/API/URL

Tomaz-Vieira: I figured eventually `ParsedUrl` (or whatever other name you might want to give it) could be used to handle URLs in more places throughout the code where URL manipulation is required. The native implementation would be insufficient because it silently breaks on funny protocols:

```
new URL("gs://example.com/some/path")
// produces URL { (...) host: "", hostname: "", port: "", pathname: "//example.com/some/path"}
// note the empty hostname, which was interpreted as being part of the pathname
```

So even if we extended `URL`, we'd still need to do the parsing ourselves.

Still, if we split handling the special Neuroglancer URLs (e.g. `precomputed://http://...`, `gs://`) from the expanded, canonical URLs (i.e.: `https?:`), then extending the native `URL` would be enough. But I think unifying all URL handling under a single class would be simpler to follow.

PR #328, opened 2021-06-30
Update compressed morton code algorithm

chrisroat:
The current algorithm adds an extra bit for grid sizes that are a power of 2.  This was discovered by me and @william-silversmith when I was debugging grids of that size.  I was seeing missing/swapped data when reading sharded tensorstore datasets with cloud-volume, which had implemented the current version of the algorithm.  See https://github.com/seung-lab/cloud-volume/pull/482


jbms:
Thanks!

PR #329, opened 2021-06-30
missing dependency for video_tool and typecheck for basestring in python3

moenigin:


jbms:
Thanks.  I think we can drop python 2 compatibility and just replace basestring with str

moenigin:
ok!

jbms:
Thanks!

PR #331, opened 2021-07-19
Remove 'Loading...' for Annotations

ang-be-taylor:
Updates to allow annotation information to be fetched more quickly to avoid the 'Loading...' text in the UI.

### DIFF
--- a/src/neuroglancer/ui/annotations.ts
+++ b/src/neuroglancer/ui/annotations.ts
@@ -1317,7 +1317,6 @@ export function UserLayerWithAnnotationsMixin<TBase extends {new (...args: any[]
           !this.annotationStates.states.includes(annotationLayer)) {
         return;
       }
-      state.rank = mouseState.coordinateSpace.rank;
 
       state.annotationId = mouseState.pickedAnnotationId;
       state.annotationType = mouseState.pickedAnnotationType;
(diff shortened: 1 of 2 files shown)


ang-be-taylor:
@googlebot I fixed it.

src/neuroglancer/ui/annotations.ts
```diff
@@ -1316,7 +1317,12 @@ export function UserLayerWithAnnotationsMixin<TBase extends {new (...args: any[]
           !this.annotationStates.states.includes(annotationLayer)) {
         return;
       }
+      state.rank = mouseState.coordinateSpace.rank;
```
jbms, line 1320: This is the rank of the global coordinate space, which is not necessarily the same as the rank of the annotation source.  I think we don't actually need to store the rank here, instead you can obtain the rank from `annotationLayer.source.rank` inside of `displayAnnotationState`.

ang-be-taylor: Updated.

src/neuroglancer/ui/annotations.ts
```diff
@@ -1342,12 +1347,30 @@ export function UserLayerWithAnnotationsMixin<TBase extends {new (...args: any[]
                                                   }))),
                   ({annotation, chunkTransform}, parent, context) => {
                     if (annotation == null) {
-                      const statusMessage = document.createElement('div');
-                      statusMessage.classList.add('neuroglancer-selection-annotation-status');
-                      statusMessage.textContent =
-                          (annotation === null) ? 'Annotation not found' : 'Loading...';
-                      parent.appendChild(statusMessage);
-                      return;
+                      if (state.annotationType && state.annotationSerialized) {
```
jbms, line 1350: Check `state.annotationType !== undefined && state.annotationSerialized !== undefined`.

Otherwise the check will fail for AnnotationType.POINT which happens to have a value of 0.

ang-be-taylor: Updated.

PR #332, opened 2021-07-22
Correct link to zarr array dimension spec

unidesigner:


unidesigner:
Additional comment: How would I specify the resolution/unit of a dataset? In [the source](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/zarr/frontend.ts#L146) it seems to use the shape value which seems wrong?
Also, the README does not clearly describe where the `_ARRAY_DIMENSIONS` attribute should bi in `.zattrs` or `.zarray`.
Maybe it would be useful to link an example zarr dataset?

unidesigner:
@googlebot I signed it!

PR #347, opened 2021-10-25
Support default keys for makeToolButton

kaikue:
This is part of the seung-lab merge project (see #340).

It would be convenient to have default keybinds for operations like merge and split, so that the user doesn't have to set them manually. This PR adds an optional `defaultKey` parameter to the `ToolBindingWidget` constructor and `makeToolButton` function, which will fill in the tool button with that key the first time Neuroglancer is loaded, but respect the user's choice and use a different key if one has been set manually.

Caveat: If the user unsets the tool (by double-clicking) then reloads Neuroglancer, it will be set back to the default. The `LayerToolBinder.jsonToKey` isn't structured to support tool -> `undefined` mappings, so I'm not sure if there's a good way around this, but it's a minor annoyance.


chrisj:
Closing this in favor of https://github.com/google/neuroglancer/pull/464
@kaikue was on our team, this was accidently left around after we chose a different approach

PR #354, opened 2021-11-11
Support toggle activation for tools

kaikue:
This is part of the seung-lab merge project.

For some tools (such as merge and split), it is more convenient to have the tool be activated by toggle (press and release the key once to activate, then press and release the key again to deactivate) rather than hold (press the key to activate, then release the key to deactivate). This PR adds a `toggle` parameter (defaulting to `false`) in the `Tool` constructor which supports this behavior. To preserve default behavior, I don't set this to `true` for any currently existing tools, but I could if you think that's a good idea (it would be a small change to `MergeSegmentsTool` and `SplitSegmentsTool`, or any other tool you think would be suited to this behavior).

It also makes the `deactivate` method on the `ToolBinder` public. This is needed for tools to be able to disable themselves upon the completion of an operation (i.e. finishing a merge, then the merge tool gets automatically deactivated). Also, it calls the `Tool.deactivate()` method in `ToolBinder.deactivate()`, so that tools can clear their own internal state.

### DIFF
--- a/src/neuroglancer/ui/tool.ts
+++ b/src/neuroglancer/ui/tool.ts
@@ -264,7 +264,8 @@ export class ToolBinder extends RefCounted {
       return;
     }
     else if (this.activeTool) {
-      StatusMessage.showTemporaryMessage(`Can't activate tool ${tool} while tool ${this.activeTool} is active`);
+      StatusMessage.showTemporaryMessage(
+        `Can't activate tool ${tool.description} while tool ${this.activeTool.tool.description} is active`);
       return;
     }
     const activation = new ToolActivation(tool, inputEventMapBinder);
--- a/src/neuroglancer/ui/tool.ts
+++ b/src/neuroglancer/ui/tool.ts
@@ -213,7 +213,7 @@ export class ToolBinder extends RefCounted {
   bindings = new Map<string, Borrowed<Tool>>();
   changed = new Signal();
   private activeTool: Owned<ToolActivation>|undefined;
-  private queuedTool: ToolActivation|undefined;
+  private queuedTool: Tool|undefined;
   private debounceDeactivate = this.registerCancellable(debounce(() => this.deactivate_(), 1));
   private debounceReactivate = this.registerCancellable(debounce(
     (inputEventMapBinder: InputEventMapBinder) => this.reactivateQueuedTool(inputEventMapBinder), 1));
@@ -271,7 +271,7 @@ export class ToolBinder extends RefCounted {
     }
     else if (this.activeTool) {
       if (this.activeTool.tool.toggle && !tool.toggle) {
-        this.queuedTool = this.activeTool;
+        this.queuedTool = this.activeTool.tool;
       }
       this.deactivate_();
     }
@@ -296,9 +296,9 @@ export class ToolBinder extends RefCounted {
 
   private reactivateQueuedTool(inputEventMapBinder: InputEventMapBinder) {
     if (this.queuedTool) {
-      const activation = new ToolActivation(this.queuedTool.tool, inputEventMapBinder);
+      const activation = new ToolActivation(this.queuedTool, inputEventMapBinder);
       this.activeTool = activation;
-      this.queuedTool.tool.activate(activation);
+      this.queuedTool.activate(activation);
       this.queuedTool = undefined;
     }
   }
--- a/src/neuroglancer/ui/tool.ts
+++ b/src/neuroglancer/ui/tool.ts
@@ -50,7 +50,7 @@ export class ToolActivation<ToolType extends Tool = Tool> extends RefCounted {
 export abstract class Tool<LayerType extends UserLayer = UserLayer> extends RefCounted {
   changed = new Signal();
   keyBinding: string|undefined = undefined;
-  constructor(public layer: LayerType) {
+  constructor(public layer: LayerType, public toggle: boolean = false) {
     super();
   }
   get mouseState() {
@@ -258,19 +258,28 @@ export class ToolBinder extends RefCounted {
     }
     this.debounceDeactivate.cancel();
     if (tool === this.activeTool?.tool) {
+      if (tool.toggle) {
+        this.deactivate();
+      }
+      return;
+    }
+    else if (this.activeTool) {
+      StatusMessage.showTemporaryMessage(`Can't activate tool ${tool} while tool ${this.activeTool} is active`);
       return;
     }
     const activation = new ToolActivation(tool, inputEventMapBinder);
     this.activeTool = activation;
-    const expectedCode = `Key${key}`;
-    activation.registerEventListener(window, 'keyup', (event: KeyboardEvent) => {
-      if (event.code === expectedCode) {
+    if (!tool.toggle) {
+      const expectedCode = `Key${key}`;
+      activation.registerEventListener(window, 'keyup', (event: KeyboardEvent) => {
+        if (event.code === expectedCode) {
+          this.debounceDeactivate();
+        }
+      });
+      activation.registerEventListener(window, 'blur', () => {
         this.debounceDeactivate();
-      }
-    });
-    activation.registerEventListener(window, 'blur', () => {
-      this.debounceDeactivate();
-    });
+      });
+    }
     tool.activate(activation);
     return tool;
   }
@@ -287,11 +296,12 @@ export class ToolBinder extends RefCounted {
     super.disposed();
   }
 
-  private deactivate() {
+  deactivate() {
     this.debounceDeactivate.cancel();
     const activation = this.activeTool;
     if (activation === undefined) return;
     this.activeTool = undefined;
+    activation.tool.deactivate();
     activation.dispose();
   }
 }
--- a/src/neuroglancer/ui/tool.ts
+++ b/src/neuroglancer/ui/tool.ts
@@ -45,6 +45,9 @@ export class ToolActivation<ToolType extends Tool = Tool> extends RefCounted {
   bindInputEventMap(inputEventMap: EventActionMap) {
     this.inputEventMapBinder(inputEventMap, this);
   }
+  cancel() {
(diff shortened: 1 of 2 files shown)


src/neuroglancer/ui/tool.ts
```diff
@@ -287,11 +297,12 @@ export class ToolBinder extends RefCounted {
     super.disposed();
   }
 
-  private deactivate() {
+  deactivate() {
     this.debounceDeactivate.cancel();
     const activation = this.activeTool;
     if (activation === undefined) return;
     this.activeTool = undefined;
+    activation.tool.deactivate();
```
jbms, line 305: I just pushed 8ebac69881ab5a28d948daf743d6856d7a53c4db to eliminate the `Tool.deactivate` method since it was not previously used and is redundant.

src/neuroglancer/ui/tool.ts
```diff
@@ -258,19 +258,29 @@ export class ToolBinder extends RefCounted {
...
     this.debounceDeactivate.cancel();
     if (tool === this.activeTool?.tool) {
+      if (tool.toggle) {
+        this.deactivate();
+      }
+      return;
+    }
+    else if (this.activeTool) {
+      StatusMessage.showTemporaryMessage(
+        `Can't activate tool ${tool.description} while tool ${this.activeTool.tool.description} is active`);
```
jbms, line 268: Perhaps it would be better to just deactivate the existing tool in this case?

Another question is whether there should now be a way to encode in the JSON state which tool is activated.

src/neuroglancer/ui/tool.ts
```diff
@@ -287,11 +297,12 @@ export class ToolBinder extends RefCounted {
     super.disposed();
   }
 
-  private deactivate() {
+  deactivate() {
```
jbms, line 300: I think this should remain private, and instead you could add a `cancel` method to `ToolActivation`.

src/neuroglancer/ui/tool.ts
kaikue: Makes sense, I've removed this part.

src/neuroglancer/ui/tool.ts
kaikue: There has been some discussion of this. An example scenario is a multicut in progress, where the user does a significant amount of work with the tool active, and may not want it to be easily deactivated by accident. That could be mitigated by not clearing the multicut state when it's deactivated, but that may be confusing. We're still discussing the ideal behavior from a user's perspective for this.

jbms: I think there definitely should be a way to deactivate without clearing the state, because the user may wish to use another tool to adjust display settings, etc.

Possibly when you activate a non-toggle tool while a toggle tool is activated, the toggle tool should be temporarily deactivated, but then re-activated once there are no non-toggle tools activated.  I would suppose that activating another toggle tool should permanently deactivate the previous toggle tool.

As far as multicut, my understanding is that it was specifically desired to be able to share the multicut points by sharing the neuroglancer state, and that was the reason for encoding the multicut points in the neuroglancer state in the first place.  That means at the very least either the toggled-on tool needs to be encoded in the JSON state, or the multicut points need to be kept even when the tool is deactivated.

kaikue: The team agreed the behavior you suggest is good, so I pushed a change to make it work that way. Now, if a toggle tool is active and another toggle tool is activated, the first one gets disabled and the new one becomes active. If a toggle tool is active and a non-toggle tool is activated, the first one gets disabled and the new one becomes active, with the first one becoming queued to reactivate as soon as the new one is deactivated. So from a user's perspective, you can only have one toggle tool at a time, but activate any non-toggle tool "on top of" that active tool. (Also, multicut points will be preserved when the tool is disabled, but the graphene changes aren't included in this PR.)

src/neuroglancer/ui/tool.ts
kaikue: I am not sure how to do that while preserving the ideal behavior. For tools to be able to deactivate themselves, it seems like they need to be setting `activeTool` in `ToolBinder` to `undefined`. I can add another public method to `ToolBinder` which just calls `deactivate` and leave `deactivate` private, but that would just be redundant from making `deactivate` itself public. Is there another strategy here that I'm not seeing?

jbms: My thinking was just that by making the "public" interface a method of ToolActivation, it is clearer that it is intended only for a tool to deactivate itself, and also allows you to check that the tool is actually active, so that it isn't deactivated by mistake if event handlers/signals are run in an unexpected order.

You raise a good point that some public method of `ToolBinder` is needed --- unfortunately the public/private access control provided by typescript is not quite adequate for this case.  I would suggest renaming `deactivate` to `deactivate_` and making it public, but indicating it is for internal use only.  Then you can call it from a `cancel` method of ToolActivation.

kaikue: That makes sense. The only issue I can foresee is that in our branch, we have extra controls in the sidebar for multicut submit/cancel (Chris added these but says he is not opposed to removing them). If we keep those, they need to be able to deactivate the tool on submit or cancel, but they don't have a reference to the `ToolActivation` (since the tool was created with `makeToolButton`, and these buttons are outside the tool). They do have a reference to the layer, so they can call `layer.manager.root.toolBinder.deactivate_()`, but if that's an "internal" method then they probably should not be calling that. Do you have any thoughts on this case? In general it seems like it might be useful to deactivate the current tool from elsewhere in the code.

jbms: I think the multicut tool, like all tools when activated, should open a status area at the bottom that shows the relevant key bindings and current multicut state.  I think that would also be a good place to put any additional buttons for cancel/submit, and you will have access to the ToolActivation there as well.

kaikue: Okay, the team agrees that is a good idea (in fact we already had buttons there for cancel/submit; the sidebar buttons were extra but we can just remove those). I pushed a change to support this.

src/neuroglancer/ui/tool.ts
```diff
@@ -253,41 +258,65 @@ export class ToolBinder extends RefCounted {
   activate(key: string, inputEventMapBinder: InputEventMapBinder): Borrowed<Tool>|undefined {
```
jbms, line 258: Now that `inputEventMapBinder` is no longer tied to the activation of a single tool, since it is saved by `debounceReactivate`, let's just make it a constructor parameter of `ToolBinder` instead.  The way you have it now works fine, but given that it will always be a fixed value I think it is clearer to make it a constructor parameter.

src/neuroglancer/ui/tool.ts
```diff
@@ -45,20 +45,22 @@ export class ToolActivation<ToolType extends Tool = Tool> extends RefCounted {
   bindInputEventMap(inputEventMap: EventActionMap) {
     this.inputEventMapBinder(inputEventMap, this);
   }
+  cancel() {
+    this.tool.layer.manager.root.toolBinder.deactivate_();
```
jbms, line 49: Can you add a check that `this == toolBinder.activeTool` to prevent spurious canceling?

src/neuroglancer/ui/tool.ts
```diff
@@ -211,7 +213,10 @@ export class ToolBinder extends RefCounted {
   bindings = new Map<string, Borrowed<Tool>>();
   changed = new Signal();
   private activeTool: Owned<ToolActivation>|undefined;
-  private debounceDeactivate = this.registerCancellable(debounce(() => this.deactivate(), 1));
+  private queuedTool: ToolActivation|undefined;
```
jbms, line 216: Just make this a `Tool` rather than `ToolActivation` since the `ToolActivation` disposers will already be run when the tool is first deactivated.

src/neuroglancer/ui/tool.ts
kaikue: `toolBinder.activeTool` is private to `ToolBinder`, should I make that public-with-an-underscore too?

jbms: yeah that would be fine.

src/neuroglancer/ui/tool.ts
kaikue: Good call, I have made this change.

src/neuroglancer/ui/tool.ts
kaikue: How would that affect `makeToolActivationStatusMessage`? I don't fully understand what that does, but it modifies `activation.inputEventMapBinder` in a way that seems to depend on `activation`.

jbms: `makeToolActivationStatusMessage` arranges for some additional code to be called whenever `inputEventMapBinder` is called, so that it can show the key bindings as part of the status message.

To be clear, we should retain the `inputEventMapBinder` member of `ToolActivation` --- you can just set it from the `inputEventMapBinder` set from the `ToolBinder` constructor.

kaikue: That makes sense, I have made that change.

src/neuroglancer/ui/tool.ts
kaikue: I have made that change.

src/neuroglancer/ui/tool.ts
```diff
@@ -250,48 +260,72 @@ export class ToolBinder extends RefCounted {
...
+      this.queuedTool = undefined;
+    }
+  }
+
   destroyTool(tool: Owned<Tool>) {
-    if (this.activeTool?.tool === tool) {
-      this.deactivate();
+    if (this.activeTool_?.tool === tool) {
+      this.deactivate_();
     }
```
jbms, line 314: All check for queueTool here

kaikue: Done

jbms:
Thanks!

PR #362, opened 2022-01-07
feat(segmentation_user_layer): adds segment selection tool

tingzhao:
This pr adds tools for selecting/deselecting segments by mouse moving. The tools are added in the seg panel and can be triggered by key binding like other tools.

image

### DIFF
--- a/src/neuroglancer/segmentation_user_layer.ts
+++ b/src/neuroglancer/segmentation_user_layer.ts
@@ -59,6 +59,7 @@ import {rangeLayerControl} from 'neuroglancer/widget/layer_control_range';
 import {renderScaleLayerControl} from 'neuroglancer/widget/render_scale_widget';
 import {colorSeedLayerControl, fixedColorLayerControl} from 'neuroglancer/widget/segmentation_color_mode';
 import {registerLayerShaderControlsTool} from 'neuroglancer/widget/shader_controls';
+import { registerSegmentSelectTools } from './ui/segment_select_tools';
 
 const SELECTED_ALPHA_JSON_KEY = 'selectedAlpha';
 const NOT_SELECTED_ALPHA_JSON_KEY = 'notSelectedAlpha';
@@ -927,3 +928,4 @@ registerLayerShaderControlsTool(
     SKELETON_RENDERING_SHADER_CONTROL_TOOL_ID);
 
 registerSegmentSplitMergeTools();
+registerSegmentSelectTools();
--- a/src/neuroglancer/segmentation_user_layer.ts
+++ b/src/neuroglancer/segmentation_user_layer.ts
@@ -59,7 +59,7 @@ import {rangeLayerControl} from 'neuroglancer/widget/layer_control_range';
 import {renderScaleLayerControl} from 'neuroglancer/widget/render_scale_widget';
 import {colorSeedLayerControl, fixedColorLayerControl} from 'neuroglancer/widget/segmentation_color_mode';
 import {registerLayerShaderControlsTool} from 'neuroglancer/widget/shader_controls';
-import { registerSegmentSelectTools } from './ui/segment_select_tools';
+import {registerSegmentSelectTools} from 'neuroglancer/ui/segment_select_tools';
 
 const SELECTED_ALPHA_JSON_KEY = 'selectedAlpha';
 const NOT_SELECTED_ALPHA_JSON_KEY = 'notSelectedAlpha';
--- a/src/neuroglancer/ui/segment_list.ts
+++ b/src/neuroglancer/ui/segment_list.ts
@@ -41,7 +41,7 @@ import {clampToInterval, computeInvlerp, dataTypeCompare, DataTypeInterval, data
 import {CdfController, getUpdatedRangeAndWindowParameters, RangeAndWindowIntervals} from 'neuroglancer/widget/invlerp';
 import {makeToolButton} from 'neuroglancer/ui/tool';
 import {ANNOTATE_MERGE_SEGMENTS_TOOL_ID, ANNOTATE_SPLIT_SEGMENTS_TOOL_ID} from 'neuroglancer/ui/segment_split_merge_tools';
-import { SELECT_SEGMENTS_TOOLS_ID, DESELECT_SEGMENTS_TOOLS_ID } from './segment_select_tools';
+import { SELECT_SEGMENTS_TOOLS_ID } from './segment_select_tools';
 
 const tempUint64 = new Uint64();
 
@@ -806,12 +806,7 @@ export class SegmentDisplayTab extends Tab {
     toolbox.appendChild(makeToolButton(this, layer, {
       toolJson: SELECT_SEGMENTS_TOOLS_ID,
       label: 'Select',
-      title: 'Select segments'
-    }));
-    toolbox.appendChild(makeToolButton(this, layer, {
-      toolJson: DESELECT_SEGMENTS_TOOLS_ID,
-      label: 'Deselect',
-      title: 'Deselect segments'
+      title: 'Select/Deselect segments'
     }));
     element.appendChild(toolbox);
 
--- a/src/neuroglancer/ui/segment_list.ts
+++ b/src/neuroglancer/ui/segment_list.ts
@@ -41,7 +41,7 @@ import {clampToInterval, computeInvlerp, dataTypeCompare, DataTypeInterval, data
 import {CdfController, getUpdatedRangeAndWindowParameters, RangeAndWindowIntervals} from 'neuroglancer/widget/invlerp';
 import {makeToolButton} from 'neuroglancer/ui/tool';
 import {ANNOTATE_MERGE_SEGMENTS_TOOL_ID, ANNOTATE_SPLIT_SEGMENTS_TOOL_ID} from 'neuroglancer/ui/segment_split_merge_tools';
-import { SELECT_SEGMENTS_TOOLS_ID } from './segment_select_tools';
+import { SELECT_SEGMENTS_TOOLS_ID } from 'neuroglancer/ui/segment_select_tools';
 
 const tempUint64 = new Uint64();
 
--- a/src/neuroglancer/ui/segment_list.ts
+++ b/src/neuroglancer/ui/segment_list.ts
@@ -41,6 +41,7 @@ import {clampToInterval, computeInvlerp, dataTypeCompare, DataTypeInterval, data
 import {CdfController, getUpdatedRangeAndWindowParameters, RangeAndWindowIntervals} from 'neuroglancer/widget/invlerp';
 import {makeToolButton} from 'neuroglancer/ui/tool';
 import {ANNOTATE_MERGE_SEGMENTS_TOOL_ID, ANNOTATE_SPLIT_SEGMENTS_TOOL_ID} from 'neuroglancer/ui/segment_split_merge_tools';
+import { SELECT_SEGMENTS_TOOLS_ID, DESELECT_SEGMENTS_TOOLS_ID } from './segment_select_tools';
 
 const tempUint64 = new Uint64();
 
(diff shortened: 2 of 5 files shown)


jbms:
Thanks --- this works well and seems pretty useful.  One question: I wonder if it would make sense to have a single tool for both select and deselect, where e.g. you left click and drag to select and right click and drag to deselect, or hold down some other modifier key?  Or do you think it is more convenient to have two separate tools?

tingzhao:
Right, a single tool looks more user friendly. I prefer using a modifier key for deselection. How should I add the modifier key checking?

jbms:
Modifiers keys are slightly tricky in JavaScript because there isn't a way to directly query which modifier keys are currently pressed, except from a KeyboardEvent or MouseEvent.

Therefore the way to do it would be to add some global tracking of the modifier states in Neuroglancer, by registering keydown and keyup handlers on the top-level `window` object, independent of any specific tool (e.g. an additional `modifiers` field of `MouseState`).

There is the issue that combining multiple modifier keys may interfere with OS shortcuts.

An alternative would be to make the select tool require holding down a mouse button --- e.g. left button to select, right button to unselect.

tingzhao:
For the alternative solution using button down, how can I make mouse dragging respond to the selection tool when the tool is activated, instead of triggering the default action such as rotation?

jbms:
You can use `activation.bindInputEventMap` --- refer to `segment_split_merge_tools.ts` for an example.

tingzhao:
In the updated PR I combined select and deselect tools into one tool. When the tool is activated, the user can select a segment by a left mouse click or select multiple segments by dragging with the left mouse button pressed. It works the same way for deselecting with the right mouse button.

src/neuroglancer/ui/segment_list.ts
```diff
@@ -41,6 +41,7 @@ import {clampToInterval, computeInvlerp, dataTypeCompare, DataTypeInterval, data
 import {CdfController, getUpdatedRangeAndWindowParameters, RangeAndWindowIntervals} from 'neuroglancer/widget/invlerp';
 import {makeToolButton} from 'neuroglancer/ui/tool';
 import {ANNOTATE_MERGE_SEGMENTS_TOOL_ID, ANNOTATE_SPLIT_SEGMENTS_TOOL_ID} from 'neuroglancer/ui/segment_split_merge_tools';
+import { SELECT_SEGMENTS_TOOLS_ID } from './segment_select_tools';
```
jbms, line 44: Please change to `neuroglancer/ui/segment_select_tools`

src/neuroglancer/ui/segment_select_tools.ts
```diff
@@ -0,0 +1,142 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {SegmentationUserLayer} from 'neuroglancer/segmentation_user_layer';
+import {removeChildren} from 'neuroglancer/util/dom';
+import {makeToolActivationStatusMessageWithHeader, registerLayerTool, Tool, ToolActivation} from 'neuroglancer/ui/tool';
+import { EventActionMap } from '../util/event_action_map';
```
jbms, line 25: please change to `neuroglancer/util/event_action_map`

src/neuroglancer/ui/segment_select_tools.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export const SELECT_SEGMENTS_TOOLS_ID = 'selectSegments';
+
+const selectEvent = 'mousedown0';
+const deselectEvent = 'mousedown2';
+const SELECT_SEGMENTS_INPUT_EVENT_MAP = EventActionMap.fromObject({
+  [`at:shift?+${selectEvent}`]: 'drag-select-segments',
```
jbms, line 32: key should be string, not array.

src/neuroglancer/ui/segment_select_tools.ts
```diff
@@ -0,0 +1,142 @@
...
+  [`at:shift?+${deselectEvent}`]: 'drag-deselect-segments',
+  'at:shift?+mouseup0': 'deactivate-drag-select-segments',
```
jbms, line 34: Rather than use separate mouseup events to end the drag, which can be fragile (e.g. mouseup may happen outside the browser window), you can use `startRelativeMouseDrag` (from util/mouse_drag.ts) which installs its own mouseup handlers to end the drag, as is used elsewhere in neuroglancer for mouse drag handling.

`startRelativeMouseDrag` will also invoke a callback on each mouse move event, which in this case you can ignore.

src/neuroglancer/ui/segment_select_tools.ts
```diff
@@ -0,0 +1,142 @@
...
+      const {segmentSelectionState} = layer.displayState;
+      if (segmentSelectionState.hasSelectedSegment) {
+        const segment = segmentSelectionState.selectedSegment;
+        const {visibleSegments} = layer.displayState.segmentationGroupState.value;
+        switch (this.state) {
+          case ToolState.SELECT:
+            visibleSegments.add(segment);
+            break;
+          case ToolState.DESELECT:
+            if (visibleSegments.has(segment)) {
```
jbms, line 87: No need for the `has` condition since that is checked anyway inside `delete`

src/neuroglancer/ui/segment_select_tools.ts
tingzhao: Here the key is not an array because [`at:shift?+${selectEvent}`] means the key is the evaluated result from the expression `at:shift?+${selectEvent}`.

src/neuroglancer/ui/segment_select_tools.ts
```diff
@@ -0,0 +1,142 @@
...
+              visibleSegments.delete(segment);
+            }
+        }
+      }
+    };
+
+    const startSelecting = () => {
+      updateStatus();
+      trySelectSegment();
+      activation.registerDisposer(
```
jbms, line 97: Since this event handlers are never unregistered, `startSelecting` should not be called more than once per activation as it currently stands.

src/neuroglancer/ui/segment_select_tools.ts
```diff
@@ -0,0 +1,142 @@
...
+});
+
+enum ToolState {
+  IGNORE,
+  SELECT,
+  DESELECT,
+};
+
+export class SelectSegmentsTool extends Tool<SegmentationUserLayer> {
+  constructor(layer: SegmentationUserLayer, private state: ToolState) {
```
jbms, line 45: It might make more sense for `state` to be a local variable within `activate` rather than a member of `SelectSegmentsTool`.

src/neuroglancer/ui/segment_select_tools.ts
jbms: Ah yes, my mistake!

src/neuroglancer/ui/segment_select_tools.ts
tingzhao: Before pushing the commit, I'd like to make sure I'm using startRelativeMouseDrag correctly:

```
  const startSelecting = (event: ActionEvent<MouseEvent>) => {
      updateStatus();
      trySelectSegment();
      activation.registerDisposer(
        layer.displayState.segmentSelectionState.changed.add(trySelectSegment));
      activation.registerDisposer(
        layer.displayState.segmentationGroupState.value.visibleSegments.changed.add(updateStatus));
      startRelativeMouseDrag(
        event.detail,
        (_event, _deltaX, _deltaY) => {},
        (_event) => {
          state = ToolState.IGNORE;
          updateStatus();
        }
      );
    };
```

Since startRelativeMouseDrag requires mouse event, I need to cast the event in action binding like this:
```
  activation.bindAction('drag-select-segments', event => {
      event.stopPropagation();
      state = ToolState.SELECT;
      startSelecting(event as ActionEvent<MouseEvent>);
    });
```
Hope this will not cause any problem.

jbms: Yes, that's right --- rather than using `as` you can also do:

`activation.bindAction('drag-select-segments', (event: ActionEvent<MouseEvent>) => { ... });`

But either way it is true that if you were to bind something that is not a mouse event to trigger `drag-select-segments` then it would not work correctly.

tingzhao:
Thank you for the suggestions! I just updated the PR. Please let me know if any more revision is needed.

src/neuroglancer/ui/segment_select_tools.ts
```diff
@@ -0,0 +1,131 @@
...
+    });
+    activation.bindAction('drag-deselect-segments', (event: ActionEvent<MouseEvent>) => {
+      event.stopPropagation();
+      state = ToolState.DESELECT;
+      startSelecting(event);
+    });
+  }
+
+  get description() {
+    return 'select/deselect segments';
```
jbms, line 123: Since this is shown in the top bar maybe make it more concise, e.g. just `select`.

jbms:
Thanks!  This is pretty close, there are just a few comments still unaddressed.

tingzhao:
Another issue I noticed is that shift+mousedown2 always opens the context menu in Firefox. It turned out that Firefox used the shift modifier on purpose to suppress the contextmenu event so that javascript could not override it. So the current implementation will have a downside of not fully supporting Firefox. Any advice?

stuarteberg:
>the current implementation will have a downside of not fully supporting Firefox

If it's not a great inconvenience, I humbly request that you maintain Firefox support.  BTW, I'm very excited about this new feature!

jbms:
I agree that supporting Firefox is important, and it appears that as @tingzhao said there is intentionally no way to override the context menu appearing in Firefox.  Therefore we should choose a different way to trigger deselect after all --- perhaps mousedown0 drag to select, alt+mousedown0 drag to deselect.

In that case I would suggest adding global modifier tracking to Neuroglancer, e.g. at the top level of util/keyboard_bindings.ts:

```typescript
export const globalModifiers = new WatchableValue<number>(0);
window.addEventListener('keydown', event => {  globalModifiers.value = getEventModifierMask(event); });
window.addEventListener('keyup', event => {  globalModifiers.value = getEventModifierMask(event); });
```

Then you will be able to monitor the modifier state from within the select tool.

jbms:
As far as the issue you mentioned about deselect in the 3d view deselecting multiple objects on top of each other:

If we view this as a continuous painting-type operation then it makes sense that it would deselect all of the objects.  After all, there is already doubleclick to deselect a single object.  Additionally, if after clicking to deselect you move the mouse slightly it would anyway deselect another object.

But it sounds like maybe you want some sort of threshold, where when you first click you just select/deselect once, like doubleclicking, and only after moving more than a certain amount does it switch to continuous "painting" mode?

tingzhao:
I changed deselection to alt+mousedown0. Please let me know if there is any problem. In this update I also used a threshold to trigger painting as you suggested. I set the distance threshold to 5 screen pixels. I'm not sure if it's optimal because I only tested it on my macbook with touchpad.

jbms:
Thanks!

Can you add a class definition for the tool to `python/neuroglancer/viewer_state.py` --- refer to `SplitSegmentsTool` for an example?

src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -59,6 +59,7 @@ import {rangeLayerControl} from 'neuroglancer/widget/layer_control_range';
 import {renderScaleLayerControl} from 'neuroglancer/widget/render_scale_widget';
 import {colorSeedLayerControl, fixedColorLayerControl} from 'neuroglancer/widget/segmentation_color_mode';
 import {registerLayerShaderControlsTool} from 'neuroglancer/widget/shader_controls';
+import { registerSegmentSelectTools } from './ui/segment_select_tools';
```
jbms, line 62: Please change to `neuroglancer/ui/segment_select_tools` for consistency.

tingzhao:
> Thanks!
> 
> Can you add a class definition for the tool to `python/neuroglancer/viewer_state.py` --- refer to `SplitSegmentsTool` for an example?

No problem. Is it something like this?

```
@export_tool
class SelectSegmentsTool(Tool):
    __slots__ = ()
    TOOL_TYPE = 'selectSegments'
```

jbms:
Yeah exactly.

jbms:
Thanks, Ting!

stuarteberg:
selection-tool-guide

PR #363, opened 2022-01-10
feat(tools): default tool keybinds

chrisj:
also added a fix to help menu so that it picks up changes to the tool global keybinds. We want to modify the defaultGlobalBindings to allow non-shift keybinds for certain tools.

I don't like this line ```const globalBindings = [...bindings][0][1].parents[0];``` I was hoping you had a better idea on how to handle it.

Also, the change I made to `src/neuroglancer/help/input_event_bindings.ts` is not enough. We only want to apply the default keybinds to layers that are created by the user, not layers loaded from the state.

If users change the layer type to 'auto' or 'segmentation' before entering the layer url, this code won't fire. I'm not sure the correct place to differentiate between newly created layers and those loaded from the state.

### DIFF
--- a/src/neuroglancer/help/input_event_bindings.ts
+++ b/src/neuroglancer/help/input_event_bindings.ts
@@ -151,7 +151,14 @@ export class InputEventBindingHelpDialog extends SidePanel {
         layerBindings = [];
         layerToolBindingsMap.set(tool.layer, layerBindings);
       }
-      layerBindings.push([`shift+key${key.toLowerCase()}`, tool.description]);
+      const globalBindings = [...bindings][0][1].parents[0];
+      const bindingsForTool = [...globalBindings.bindings.entries()].filter(([event, eventAction]) => {
+        return eventAction.action === `tool-${key.toUpperCase()}` && event.startsWith('at');
+      });
+      const bindingIdentifiersForTool = bindingsForTool.map(([event, eventAction]) => {
+        return eventAction.originalEventIdentifier || event;
+      });
+      layerBindings.push([bindingIdentifiersForTool.join(', '), tool.description]);
     }
     const layerToolBindings = Array.from(layerToolBindingsMap.entries());
     if (layerToolBindings.length > 0) {
--- a/src/neuroglancer/ui/layer_data_sources_tab.ts
+++ b/src/neuroglancer/ui/layer_data_sources_tab.ts
@@ -321,6 +321,7 @@ function changeLayerTypeToDetected(userLayer: UserLayer) {
     const layerConstructor = userLayer.detectedLayerConstructor;
     if (layerConstructor !== undefined) {
       changeLayerType(userLayer.managedLayer, layerConstructor);
+      userLayer.managedLayer.layer!.toolBinder.loadDefaultKeybinds();
       return true;
     }
   }
@@ -362,7 +363,6 @@ export class LayerDataSourcesTab extends Tab {
       layerTypeDetection.classList.add('neuroglancer-layer-data-sources-tab-type-detection');
       layerTypeDetection.addEventListener('click', () => {
         changeLayerTypeToDetected(layer);
-        layer.managedLayer.layer!.toolBinder.loadDefaultKeybinds();
       });
     }
     const reRender = this.reRender = animationFrameDebounce(() => this.updateView());
--- a/src/neuroglancer/ui/layer_data_sources_tab.ts
+++ b/src/neuroglancer/ui/layer_data_sources_tab.ts
@@ -362,6 +362,7 @@ export class LayerDataSourcesTab extends Tab {
       layerTypeDetection.classList.add('neuroglancer-layer-data-sources-tab-type-detection');
       layerTypeDetection.addEventListener('click', () => {
         changeLayerTypeToDetected(layer);
+        layer.managedLayer.layer!.toolBinder.loadDefaultKeybinds();
       });
     }
     const reRender = this.reRender = animationFrameDebounce(() => this.updateView());
--- a/src/neuroglancer/ui/tool.ts
+++ b/src/neuroglancer/ui/tool.ts
@@ -33,6 +33,17 @@ import {Signal} from 'neuroglancer/util/signal';
 
 const TOOL_KEY_PATTERN = /^[A-Z]$/;
 
+type DefaultKeybindsForLayerType = {
+  [key: string]: any
+};
+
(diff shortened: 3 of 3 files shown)


src/neuroglancer/ui/layer_data_sources_tab.ts
```diff
@@ -362,6 +362,7 @@ export class LayerDataSourcesTab extends Tab {
       layerTypeDetection.classList.add('neuroglancer-layer-data-sources-tab-type-detection');
       layerTypeDetection.addEventListener('click', () => {
         changeLayerTypeToDetected(layer);
+        layer.managedLayer.layer!.toolBinder.loadDefaultKeybinds();
```
jbms, line 365: Should probably be done in `changeLayerTypeToDetected` rather than here, since `changeLayerTypeToDetected` is also called when you press enter again with the URL input area focused

jbms:
One of the challenges here is that just based on the layer type (e.g. `"segmentation"`) we don't know if it is a graphene layer or not.  But perhaps for your purposes that is not a problem?

chrisj:
It would be cleaner to look at the datasource. Something like `userLayer.dataSources[0].spec.url.split(':')[0]`

src/neuroglancer/ui/tool.ts
```diff
@@ -397,6 +408,25 @@ export class LayerToolBinder {
     this.clear();
   }
 
+  loadDefaultKeybinds() {
+    if (typeof DEFAULT_KEYBINDS !== 'undefined') {
+      const obj: DefaultKeybindsForLayerType = {};
+      const defaultKeybindsForLayer = DEFAULT_KEYBINDS[this.layer.type] || {};
+      for (const [key, value] of Object.entries(defaultKeybindsForLayer)) {
+        const toolAlreadyBound = [...this.bindings.values()].includes(value);
```
jbms, line 416: You can check if the tool is already bound by checking `this.jsonToKey`.

src/neuroglancer/ui/tool.ts
```diff
@@ -397,6 +408,25 @@ export class LayerToolBinder {
...
+        if (toolAlreadyBound) {
+          continue;
+        }
+        const existingTool = this.globalBinder.get(key);
+        if (existingTool && !existingTool.layer.managedLayer.archived) {
```
jbms, line 421: I can see that it might potentially be desired, but is there a specific reason why you want to override an existing binding for an archived layer?

jbms:
It looks like currently the DEFAULT_KEYBINDS logic doesn't actually support non-shift bindings, but you have already made some modifications to the help panel to try to support non-shift bindings.  Is that correct?

chrisj:
I felt like the easiest way to handle non shift keys was to modify `getDefaultGlobalBindings` or override it with a similar function for our deployment. I could create an additional config file for that, does that sound good?

src/neuroglancer/ui/tool.ts
chrisj: I will remove that, I was unfamiliar with how to fully remove layers when I implemented that.

src/neuroglancer/ui/tool.ts
chrisj: I deleted this check since it wasn't necessary after I changed it to only run on new layers.

jbms:
Regarding non-shift bindings, currently the way tool bindings work is that there are global bindings for:

shift+keya -> tool-A
shift+keyb -> tool-B
...

Then the handler for `tool-A` looks up which tool is bound to "A" and activates it.

Therefore I'm wondering how you intended the support for non-shift tool bindings to work --- will you just add an additional binding for e.g. keym -> tool-M, so that "M" tools can be used either with or without shift?

chrisj:
Yeah, I was thinking that would be good enough.

src/neuroglancer/help/input_event_bindings.ts
```diff
@@ -151,7 +151,14 @@ export class InputEventBindingHelpDialog extends SidePanel {
         layerBindings = [];
         layerToolBindingsMap.set(tool.layer, layerBindings);
       }
-      layerBindings.push([`shift+key${key.toLowerCase()}`, tool.description]);
+      const globalBindings = [...bindings][0][1].parents[0];
```
jbms, line 154: Rather than iterate over all bindings for each tool, it would be better to build a Map<string, string[]> of bindings for each tool (i.e. where the action matches "^tool-[A-Z]$") outside the loop.

You can iterate over all entries of an EventActionMap (including any parents) using the `entries()` method which returns an iterator.

src/neuroglancer/ui/tool.ts
```diff
@@ -33,6 +33,17 @@ import {Signal} from 'neuroglancer/util/signal';
...
 
+type DefaultKeybindsForLayerType = {
+  [key: string]: any
+};
+
+type DefaultKeybinds = {
+  [layerType: string]: DefaultKeybindsForLayerType
+};
+
+declare const DEFAULT_KEYBINDS: DefaultKeybinds|undefined;
```
jbms, line 44: I'd suggest renaming to DEFAULT_TOOL_BINDINGS_FOR_LAYER_TYPE

chrisj:
Our requirements are a bit different than I had thought, we want bindings for our tools that persist even if the user wants to bind to a shift+key.

Something like this would probably accomplish what we want and we could close this pull request.

```typescript
const bindActionToTool = (action: string, toolType: string, layerType: UserLayerConstructor) => {
  this.bindAction(action, () => {
    const segmentationLayers = this.layerManager.managedLayers.filter((managedLayer) => {
      return managedLayer.layer instanceof layerType;
    });

    if (segmentationLayers.length > 0) {
      const firstLayer = segmentationLayers[0];

      if (this.toolBinder.activeTool_?.tool.toJSON() === toolType) {
        this.toolBinder.activate(this.toolBinder.activeTool_.tool);
      } else {
        const tool = restoreTool(firstLayer.layer!, toolType);
        this.toolBinder.activate(tool);
      }
    }
  });
}

bindActionToTool('multicut', 'multicutSegments', SegmentationUserLayer);
```

This would only be for our own custom deployments. It's not perfect, for excample, ToolBinder does not know about this bind so if a user opens up with our custom keybind, then presses the shift+keybind for that tool, it will open up a new tool rather than close the existing tool. I don't think it's worth fixing this since this feature goes against the pattern you intended.

I'm hoping our requirements change so that we can drop this band-aid solution.

PR #366, opened 2022-01-20
fix(mouse_bindings): Fix control+mousedown1 support on macOS Firefox

### LINKED ISSUE #365: Unable to add manual annotations in FireFox on macOS
We (@shbhas and I) are currently unable to add manual annotations with FireFox on macOS.

`control+mousedown0` is being treated as `control+mousedown1`, bringing up the selection pane instead of adding an annotation. I suspect Firefox is doing some "helpful" remapping of the mouse event when control is held.

Annotations work as expected in Chrome and Safari (on macOS 12).

Platform: macOS 10.15.7 & 12.1
Firefox 96.0.2 & 97.0b6

image

jbms:
Fixes #365.

### DIFF
--- a/src/neuroglancer/util/mouse_bindings.ts
+++ b/src/neuroglancer/util/mouse_bindings.ts
@@ -43,7 +43,36 @@ export class MouseEventBinder<EventMap extends EventActionMapInterface> extends 
     });
     this.registerEventListener(target, 'mousedown', (event: MouseEvent) => {
       if (commonHandler !== undefined) commonHandler(event);
-      this.dispatch(`mousedown${event.button}`, event);
+      let button = event.button;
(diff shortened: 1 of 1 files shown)


jbms:
@perlman Please let me know if this fixes the problem for you.

perlman:
No, still not working, but it's half way:
The selection pane no longer pops up on `control+mousedown1`; but no annotations are added.

`control+mousedown2` continues to function as expected (with the selection pane opening).

jbms:
Thanks for testing --- I was setting the button to 1 instead of 0.  Should be fixed now.

perlman:
It works.

I appreciate the lengthy comment in the code on how you do the disambiguation.

PR #367, opened 2022-01-25
feat(compresso): support format version 1 (backwards compatible)

william-silversmith:
Hi Jeremy,

I was updating the compresso codec with an optional index to make random access to a z slice possible, which I might use to make certain operations in CloudVolume faster (such as fetching single voxels) which I have already done for compressed segmentation (!). 

This change is the minimum necessary to support the format. It only supports full decode of both format version 0 (the original in #318) and the new indexed format, version 1. It does not support random access to z slices (that will be in the cython version).

Backwards compatible.

Thanks for reviewing this!


jbms:
Thanks!

william-silversmith:
Thanks for the speedy merge Jeremy!

PR #370, opened 2022-02-03
feat: WASM png support

william-silversmith:
Hi Jeremy,

I added PNG support for Neuroglancer and tested it with grayscale images using a newly compiled libpng.wasm module. I modeled this PR after #318 

The important differences are:
- I used two third party libraries [spng](https://github.com/randy408/libspng/tree/master/spng) and [miniz](https://github.com/richgel999/miniz) which have permissive licenses (BSD-2 and MIT respectively) in order to get PNGs to decode to a raw buffer. 
- The codec is in C rather than C++.

I have a working PR for CloudVolume that can write PNGs.

I ran PNG on CREMI aplus, which was a bigger volume than my previous small tests and again saw a ~25% improvement in storage size, so I think this is real.

I'm a bit unhappy with how I had to use a macro in order to safely free ctx at each if statement. If you know a better C idiom, I'd be happy to update it.

Thanks!


jbms:
Thanks.

I'm assuming spng and miniz are unmodified.  If that is the case, can you download them (and check their sha256 checksum) as part of the build script rather than including their source code directly?

jbms:
Note: Best way to download them is probably to do it as part of the `Dockerfile`.

william-silversmith:
I can give that a shot. I assumed you might want me to modify spng to remove the encoder part though to shrink the WASM size. 

My colleague Ran mentioned that maybe I should experiment with webp too which claims on its website to generate even smaller lossless images https://developers.google.com/speed/webp/. Maybe I should check that out before we merge?

jbms:
As far as removing the encoder, I think it will make no difference, the webassembly optimizer should remove code that it can statically determine to be unused.

There are other codecs to explore --- webp, avif, jpeg2k, jpeg-xl.  However, I can imagine that for different situations, different codecs will be preferred.

It isn't ideal to add a bunch of codecs to the neuroglancer precomputed format, since it means every implementation needs to support them, but I imagine if we add them to neuroglancer they will also be useful for other formats like zarr.  The imagecodecs Python package already semi-officially adds support for them to zarr under the names `imagecodecs_*`.

william-silversmith:
I guess that makes sense, though it's possible WebP might mostly dominate PNG according to this study (on data that is pretty different): https://developers.google.com/speed/webp/docs/webp_lossless_alpha_study 

Do you anticipate neuroglancer and/or Precomputed eventually supporting all the codecs you listed? Based on the [denoising paper](https://www.biorxiv.org/content/10.1101/2021.05.29.445828v2.full.pdf) I've been expecting people to want to at least support JPEG-XL or AVIF.

jbms:
It seems plausible that all would be supported by Neuroglancer, though I would just assume wait until there is a specific use case for each one.

It sounds like the webp compression study you linked (https://developers.google.com/speed/webp/docs/webp_lossless_alpha_study) may have been inaccurate as far as webp decoding speed: https://groups.google.com/a/webmproject.org/g/webp-discuss/c/FPOfZs2cCS4/m/2F0vs7qGiKIJ
It looks like PNG decoding may actually be about twice as fast.

Also it would be nice to get code splitting working in Neuroglancer before adding too many codecs, so that users don't have to download all codecs even if they aren't using them.  Unfortunately that is currently blocked by the fact that esbuild only supports code splitting for the esm bundle format (https://github.com/evanw/esbuild/issues/16) and Firefox does not support esm for workers (https://bugzilla.mozilla.org/show_bug.cgi?id=1247687).

Though we could actually split out the wasm files even if we can't split out the javascript --- originally I had the wasm bundled into the javascript, and then split the javascript, but now that splitting the javascript can't be done, just splitting the wasm could be a reasonable alternative.

william-silversmith:
Interesting, it looks like that groups discussion is from 2013 and the study says it was last updated in 2017 (but doesn't say what was updated). They may have improved the codec implementation in the intervening years, so I guess we'll have to test it ourselves (we have different data than they tested anyway).

I was thinking similarly regarding splitting out the WASM. So far the burden isn't too high (similar to a about two big chunks).

```
148K  ./src/neuroglancer/mesh/draco/neuroglancer_draco.wasm
43K   ./src/neuroglancer/sliceview/compresso/compresso.wasm
43K   ./src/neuroglancer/sliceview/png/libpng.wasm
```

jpegjs is  39.3 KB unminified for the decoder.

In any case, I'll finish up this PR later today.

william-silversmith:
I think this PR is ready for review. I moved the libraries to the docker.

src/neuroglancer/sliceview/png/index.ts
```diff
@@ -0,0 +1,89 @@
...
+    throw new Error(`Failed to decode png image size. image size: ${nbytes}`);
+  }
+
+  const imagePtr = (m.instance.exports.malloc as Function)(nbytes);
+
+  // heap must be redefined after creating imagePtr because
+  // memory growth can detatch the buffer.
+  heap = new Uint8Array((m.instance.exports.memory as WebAssembly.Memory).buffer);
+  heap.set(buffer, bufPtr);
+
```
jbms, line 64: The javascript code should be able to get the dimensions of the image and the number of components, and the bits per component, so that it can be validated against the expected chunk shape.

I believe PNG supports 1 to 4 components, and both 8-bit and 16-bit.

william-silversmith: Won't that be error prone if the PNG isn't compressed with the expected settings? That could generate a buffer overflow.

src/neuroglancer/sliceview/png/LICENSE
```diff
@@ -0,0 +1,206 @@
+The following license covers all code in this png directory
```
jbms, line 1: Since the third-party code is no longer copied here, we can exclude this additional license file.

src/neuroglancer/sliceview/png/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -xve
+
+# This script builds `compresso.wasm` using emsdk in a docker container.
```
jbms, line 3: libpng.wasm

william-silversmith:
Okay, I moved the header logic into a JS function.

jbms:
Sorry, I think I wasn't clear.  Currently the javascript `asyncComputation` interface that you have does not allow the caller to determine or validate the data type, the dimensions, or the number of channels.   To fix that, in addition to the encoded `data`, the caller of the async request should also pass in `width`, `height`, `num_components`, and `dataType` so that those can be validated when decoding the png (see `decode_jpeg_request.ts` for an example).  The actual header decoding does not have to happen in javascript.

william-silversmith:
Okay! I validated the chunk size and added support for `convertToGrayscale`. However, I'm having trouble figuring out how to access `dataType` and it seems the jpeg decoder omits this check as well. Would you have a tip for how to do that? I looked in the debugger and its not attached to VolumeChunk or the immediate context around `await this.chunkDecoder(chunk, cancellationToken, response);`

jbms:
The jpeg library only supports 8-bit so there is no need to check data type.

The data type can be obtained from `this.source.spec.dataType`

william-silversmith:
Thanks! That worked. I think that takes care of the comments so far.

src/neuroglancer/sliceview/png/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  PALETTE = 3,
+  GRAYSCALE_ALPHA = 4,
+  RGBA = 6
+}
+
+// header constants
+const magicSpec = [ 137, 80, 78, 71, 13, 10, 26, 10 ];
+const validHeaderCode = [ 'I', 'H', 'D', 'R' ];
+
+// not a full implementation of read header, just the parts we need
```
jbms, line 49: I know you put in some export to implement this header parsing in javascript, but is there any advantage over just relying on spng to do that?

william-silversmith: The main advantage I can think of is that the error message is easier to write in an informative way. I can revert it back if you'd like.

william-silversmith:
One other refinement I could make is to set `covertToGrayscale = true` if the expected number of channels is 1. This would allow some laxness in decoding PNGs that were sloppily set to RGB mode but are still supposed to be grayscale. Not sure how you'd feel about that. It might make implementations a bit easier, though they would be inefficient.

jbms:
I think it is better not to convert to grayscale for the precomputed format --- that will just encourage laxness and force every implementation to handle that case.

src/neuroglancer/sliceview/png/png_wasm.c
```diff
@@ -0,0 +1,63 @@
...
+	// can we even read the header?
+	if (num_bytes < 8) { return 3; }
+
+	/* Create a decoder context */
+	spng_ctx *ctx = spng_ctx_new(0);
+    if (ctx == NULL) { return 4; }
+
+	/* Set an input buffer */
+	if (spng_set_png_buffer(ctx, buf, num_bytes)) {
+		RET(5);
```
jbms, line 38: A common C pattern for dealing with this without a macro would be to define:

`int retval = 0;`

at the beginning of the function.

And then at the end put:

`done: return retval;`

Then you can use:

`retval = 5; goto done;`

in the case of an error.

src/neuroglancer/sliceview/png/png_wasm.c
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+#include "spng.h"
+#include <stdbool.h>
+
+#define RET(val) spng_ctx_free(ctx); return (val);
+
+int png_decompress(
+	unsigned char* buf, unsigned int num_bytes, 
+    void* out, bool convert_to_grayscale
```
jbms, line 24: Should also pass in the size of the output buffer, even though in this case it will have already been validated by javascript code.

william-silversmith:
Okay! I fixed up those items.

src/neuroglancer/sliceview/png/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+FROM emscripten/emsdk@sha256:81ec54b7a096d28f24d906955dbf98ff336cca47658d980c243baa36f6484f9f
+
+RUN mkdir -p /usr/src/spng \
+    && curl -SL https://github.com/randy408/libspng/archive/refs/tags/v0.7.1.tar.gz \
+    | tar -xzC /usr/src/spng --strip-components=1
+
+RUN mkdir -p /usr/src/miniz \
+    && curl -SL https://github.com/richgel999/miniz/archive/refs/tags/2.2.0.tar.gz \
+    | tar -xzC /usr/src/miniz --strip-components=1
+
```
jbms, line 11: Can you also validate the checksums using sha256sum?

william-silversmith: Whoops! Sorry I forgot. I'll do that now.

william-silversmith: I seem to be having some trouble getting it to exit. Am I doing this right?

ENV SPNG_0_7_1_SHA256 0726a4914ad7155028f3baa94027244d439cd2a2fbe8daf780c2150c4c951d8e

RUN mkdir -p /usr/src/spng \
    && curl -SL -o /usr/src/spng.tar.gz https://github.com/randy408/libspng/archive/refs/tags/v0.7.1.tar.gz 

RUN if [ $(echo "$(SPNG_0_7_1_SHA256) /usr/src/spng.tar.gz" | sha256sum --check --status) ]; then exit 1; fi;
RUN tar -xzC /usr/src/spng -f /usr/src/spng.tar.gz --strip-components=1 \
    && rm /usr/src/spng.tar.gz

jbms: The `[` command is actually a shell command that treats its arguments as a condition.  That is useful for comparing strings or numbers, but in this case we are just testing the exit code of sha256sum so we don't need that.  Instead you can just use:

```dockerfile
RUN echo "${SPNG_0_7_1_SHA256} /usr/src/spng.tar.gz" | sha256sum --check --status
```
Note that you should also use `${...}` instead of `$(...)` for environment variable substitution.

william-silversmith: Thank you Jeremy! Fixed. My bash skills could use some work. I tried something like that, but it didn't work for me before. It's probably the curlies that made the difference.

jbms:
Thanks!

william-silversmith:
Thank you Jeremy! I'll release PNG support in CloudVolume today.

PR #378, opened 2022-02-22
feat(segmentation_user_layer): add  individual segment color picker tool

chrisj:
adding some comments on parts I'm unsure about

Currently there is no way to reset the color (remove the stated color) and it persists if the segment is removed and  re-added.

### DIFF
--- a/src/neuroglancer/segmentation_display_state/frontend.ts
+++ b/src/neuroglancer/segmentation_display_state/frontend.ts
@@ -31,7 +31,7 @@ import {observeWatchable, registerNestedSync, TrackableValue, WatchableValueInte
 import {isWithinSelectionPanel} from 'neuroglancer/ui/selection_details';
 import {Uint64Map} from 'neuroglancer/uint64_map';
 import {setClipboard} from 'neuroglancer/util/clipboard';
-import {useWhiteBackground} from 'neuroglancer/util/color';
+import {packColor, serializeColor, TrackableRGB, useWhiteBackground} from 'neuroglancer/util/color';
 import {RefCounted} from 'neuroglancer/util/disposable';
 import {measureElementClone} from 'neuroglancer/util/dom';
 import {kOneVec, vec3, vec4} from 'neuroglancer/util/geom';
@@ -40,6 +40,7 @@ import {Uint64} from 'neuroglancer/util/uint64';
 import {withSharedVisibility} from 'neuroglancer/visibility_priority/frontend';
 import {makeCopyButton} from 'neuroglancer/widget/copy_button';
 import {makeFilterButton} from 'neuroglancer/widget/filter_button';
+import {ColorWidget} from 'neuroglancer/widget/color';
 
 export class Uint64MapEntry {
   constructor(public key: Uint64, public value?: Uint64, public label?: string|undefined) {}
@@ -273,6 +274,10 @@ const segmentWidgetTemplate = (() => {
   filterElement.classList.add('neuroglancer-segment-list-entry-filter');
   const filterIndex = template.childElementCount;
   template.appendChild(filterElement);
+
+  const colorWidgetIndex = template.childElementCount;
+  template.appendChild(ColorWidget.template());
+
   return {
     template,
     copyContainerIndex,
@@ -282,6 +287,7 @@ const segmentWidgetTemplate = (() => {
     idIndex,
     labelIndex,
     filterIndex,
+    colorWidgetIndex,
     unmappedIdIndex: -1,
     unmappedCopyIndex: -1
   };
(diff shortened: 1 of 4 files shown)


src/neuroglancer/segmentation_display_state/frontend.ts
```diff
@@ -414,6 +420,20 @@ function makeRegisterSegmentWidgetEventHandlers(displayState: SegmentationDispla
...
+
+      const idString = element.dataset.id!;
+      const id = tempStatedColor;
+      id.tryParseString(idString);
+      displayState.segmentStatedColors.value.delete(id);
+      displayState.segmentStatedColors.value.set(id, testU);
+    });
+
+    // TODO, need to reigster disposer?
+    new ColorWidget(trackableRGB, undefined, children[template.colorWidgetIndex] as HTMLInputElement);
```
chrisj, line 436: new ColorWidget is usually added using registerDisposer though I don't see a way to do that in `makeRegisterSegmentWidgetEventHandlers`. Perhaps that is the wrong function to construct it in.

I made changes to ColorWidget so that you could request a template and provide it with an element in the constructor.

src/neuroglancer/util/color.ts
```diff
@@ -93,8 +93,8 @@ export function unpackRGBA(value: number) {
       ((value >>> 24) & 0xff) / 255);
 }
 
-export function serializeColor(x: vec3|vec4) {
-  if (x[3] === undefined || x[3] === 1) {
+export function serializeColor(x: vec3|vec4, ignoreAlpha=false) {
```
chrisj, line 96: added ignore alpha because that is essentially what is done in `getCSSColor`, I'm not sure why the color originally has a alpha of 0. I also considered making `getCssColor` optionally return the hex representation.

src/neuroglancer/segmentation_display_state/frontend.ts
```diff
@@ -414,6 +420,20 @@ function makeRegisterSegmentWidgetEventHandlers(displayState: SegmentationDispla
...
     element.addEventListener('action:select-position', selectHandler);
+
+    const trackableRGB = new TrackableRGB(vec3.fromValues(0, 0, 0));
+    trackableRGB.changed.add(() => {
+      const testU = new Uint64(packColor(trackableRGB.value));
+
+      const idString = element.dataset.id!;
+      const id = tempStatedColor;
+      id.tryParseString(idString);
+      displayState.segmentStatedColors.value.delete(id);
```
chrisj, line 431: I'm curious why uint64_map doesn't trigger the changed signal if the value changes for an existing key. HashMapUint64 does not update values for existing keys. Not sure if those are both intentional.

src/neuroglancer/segmentation_user_layer.css
```diff
@@ -99,6 +99,18 @@
...
+  border-color: transparent;
+  -webkit-appearance: none;
+  background-color: transparent;
+  padding: 0;
+  margin: 0;
+  margin-left: 3px;
+  height: 19px;
+  width: 20px
+}
+
```
chrisj, line 113: example of the styling
Screen Shot 2022-02-22 at 3 55 43 PM

jbms: Since the color is already displayed on the id itself, and space is tight in the segment list, I think it would be better to just use a modifier+click, e.g. alt+click, to change the color without having any visible control.

src/neuroglancer/segmentation_display_state/frontend.ts
jbms: Yeah, HashMapUint64 should be fixed to update existing keys.  That is a bug that was not previously triggered because the hash map was only used for segment colors, which never changed after initial construction.

src/neuroglancer/util/color.ts
jbms: Instead of ignoreAlpha you can just pass a vec3, e.g. color.subarray(0, 3) as vec3

src/neuroglancer/segmentation_user_layer.css
chrisj: after talking with our group, while that seems very clean, we think the discoverability of the feature would be too limited with that approach

jbms: In that case perhaps we can use a build-time option to enable the visible widget, but still allow a modifier+click to change the color in any case?

src/neuroglancer/segmentation_display_state/frontend.ts
```diff
@@ -414,6 +420,20 @@ function makeRegisterSegmentWidgetEventHandlers(displayState: SegmentationDispla
     stickyChildren[template.visibleIndex].addEventListener('click', visibleCheckboxHandler);
     children[template.filterIndex].addEventListener('click', filterHandler);
     element.addEventListener('action:select-position', selectHandler);
```
jbms, line 422: For efficiency (because segment widgets are created and updated very frequently in the segment list) these segment widgets are handled in a special way --- where we have a template that we clone and then add the required event handlers (as minimally as possible).

I think it would probably be best to avoid creating a ColorWidget at all (but still obtain the template from it), and just implement the necessary event handlers separately.

jbms:
Regarding removing the color --- if we suppose that alt+click is used to change the color (if the widget isn't being shown), then alt+rightclick could be used to remove the color.  And we can have the tooltip show the key combinations so that they are discoverable.

When the widget is shown, rightclick without modifier could also reset the color.

As far as resetting colors automatically when the segment is no longer visible:
per the discussion today with Sven about a "persistent segment list" which the seung lab branch has but the upstream branch does not, I agree that for manual setting of colors it probably makes sense that the colors are lost when you remove the segment from the persistent list.  However when the colors are configured in the state directly (currently the only way to do it in the upstream branch), e.g. using the Python API, then we don't want the colors to so easily go away.  Though perhaps it would be reasonable for every segment with a specified color to automatically get added to the persistent list in which case that would not be an issue.

PR #383, opened 2022-03-26
feat(datasource/graphene) view only

chrisj:
I wasn't sure what to call the "ways of iterating over the visible segments" so called it VISIBLE_SEGMENT_TYPE and I probably should rename highBitRepresentation as well.

I'll add some comments to the code.


src/neuroglancer/datasource/graphene/backend.ts
```diff
@@ -0,0 +1,325 @@
...
+  const source =
+      GenericSharedDataSource.get<string, DecodedMinishardIndex|undefined>(
+          chunkManager, stableStringify({
+            type: 'graphene:shardedDataSource',
+            url,
+            sharding,
+            credentialsProvider: getObjectId(credentialsProvider),
+          }),
+          {
+            download: async function(
```
jbms, line 60: I think this logic may be identical to what is in the precomputed data source.  If so it would be better to refactor to avoid duplication.

src/neuroglancer/datasource/graphene/backend.ts
```diff
@@ -0,0 +1,325 @@
...
+        msg = (await response.json())['message'];
+      } catch {
+        msg = await response.text();
+      }
+      throw new Error(`[${response.status}] ${errorPrefix}${msg}`);
+    }
+  }
+}
+
+function getVerifiedFragmentPromise(
```
jbms, line 219: Can you comment what verified means?

src/neuroglancer/datasource/graphene/base.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+import {fetchSpecialHttpByteRange as fetchSpecialHttpByteRangeOrig} from 'neuroglancer/util/byte_range_http_requests';
+import { Uint64 } from 'neuroglancer/util/uint64';
+
+export function fetchSpecialHttpByteRange(
+  credentialsProvider: SpecialProtocolCredentialsProvider, url: string,
+  startOffset: Uint64|number, endOffset: Uint64|number,
+  cancellationToken: CancellationToken): Promise<ArrayBuffer> {
+    if ((new URL(url)).origin === GCS_ORIGIN) {
```
jbms, line 88: This seems kind of risky in terms of avoiding leaking the credentials --- e.g. if Neuroglancer made GCS requests slightly differently.  If I understand correctly, the intention is that you have one credentials provider for calls to the chunk graph and mesh services, and then don't use it for fetching the static data from GCS.  I think it would be better for the callers to track explicitly which credentials provider they should use, since in the future you might not use GCS for the static data, and you might want to supply separate credentials to the static data service.

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -0,0 +1,820 @@
...
+        outputParts.length = outputParts.length - 1;
+        continue;
+      }
+    }
+    outputParts.push(part);
+  }
+  return outputParts.join('/');
+}
+
+class ScaleInfo {
```
jbms, line 74: It would be nice to refactor this to share code with the precomputed data source rather than duplicating it, if possible.

chrisj: I pushed these changes
https://github.com/google/neuroglancer/pull/383/commits/3c0028054bedd55afb89d20c5a83b09be8d56a88
https://github.com/google/neuroglancer/pull/383/commits/24ad0ddb6d15d2a71dd50169de8b844debfdfa0f

I think I could re-use PrecomputedDataSource.get though I need to override `getVolumeDataSource`. One way to do that would be to make it a method on PrecomputedDataSource though if that made sense, it would mean moving all of the other getDataSource functions into that class.

src/neuroglancer/datasource/graphene/backend.ts
chrisj: I removed this since I realized it was only being by unverified sharded meshes. I'm still working on a good explanation of verify/unverified meshes, essentially our chunked graph can give us fast but unreliable meshes (unverified) which may have gaps/holes. We are using it immediately after proofreading to provide more real time feedback. Since I removed the proofreading logic from this PR, I also stripped out the unverified mesh logic.

src/neuroglancer/datasource/graphene/base.ts
chrisj: I just removed these functions, they were only intended to be temporary, if I still need similar conditional logic on when to pass credentials, I'll handle it in the calling function.

src/neuroglancer/gpu_hash/hash_table.ts
```diff
@@ -134,17 +136,15 @@ export abstract class HashTableBase {
   /**
    * Iterates over the Uint64 keys contained in the hash set.
    *
-   * The same temp value will be modified and yielded at every iteration.
+   * Creates a new Uint64 object at every iteration (otherwise spread and Array.from() fail)
    */
-  * keys(temp = new Uint64()) {
+  * keys() {
```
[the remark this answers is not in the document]
jbms: (The same also applies to `entries`)

src/neuroglancer/segmentation_graph/source.ts
```diff
@@ -18,13 +18,22 @@ import {VisibleSegmentsState} from 'neuroglancer/segmentation_display_state/base
 import {WatchableValueInterface} from 'neuroglancer/trackable_value';
 import {Disposer, Owned, RefCounted} from 'neuroglancer/util/disposable';
 import {Uint64} from 'neuroglancer/util/uint64';
+import {RenderLayer } from 'neuroglancer/renderlayer';
+import {RenderLayerTransformOrError} from 'neuroglancer/render_coordinate_transform';
+import {MultiscaleVolumeChunkSource} from 'neuroglancer/sliceview/volume/frontend';
+
+export enum VISIBLE_SEGMENT_TYPE {
```
jbms, line 25: Can you change to `VisibleSegmentType` and add a brief comment to describe the behavior of each one?

src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -482,9 +492,14 @@ export class SegmentationUserLayer extends Base {
             loadedSubsource.activate(refCounted => {
               this.graphConnection = refCounted.registerDisposer(
                   segmentationGraph.connect(this.displayState.segmentationGroupState.value));
-              refCounted.registerDisposer(() => {
-                this.graphConnection = undefined;
-              });
+              const segmentationRenderlayer = this.someSegmentationRenderLayer();
```
jbms, line 495: `someSegmentationRenderLayer` isn't the right approach, because if another non-graphene data source has been added to the same layer it may find the render layer from that.

Additionally, if the segmentation volume has been manually unchecked by the user, there won't be a segmentation layer either.

Is there a reason why the graph connection can't just supply everything it needs directly?

src/neuroglancer/sliceview/chunked_graph/backend.ts
```diff
@@ -0,0 +1,334 @@
...
+import {registerRPC, registerSharedObject, RPC, SharedObjectCounterpart} from 'neuroglancer/worker_rpc';
+import {WatchableValueInterface} from 'neuroglancer/trackable_value';
+
+import * as vector from 'neuroglancer/util/vector';
+import {isBaseSegmentId} from 'neuroglancer/datasource/graphene/base';
+
+export class ChunkedGraphChunk extends SliceViewChunk {
+  backendOnly = true;
+  source: ChunkedGraphChunkSource|null = null;
+  mappings: Map<string, Uint64[]|null>|null = null;
```
jbms, line 36: In Neuroglancer the existing framework for downloading and caching chunks is based on the assumption that they are immutable and cacheable and basically correspond to a single (or small fixed number) of network requests.

ChunkedGraphChunk would fit better with the existing framework if for each (level, grid_pos, root_id) there were a separate chunk, rather than the current system where there is a single chunk for each (level, grid_pos).  Then the ChunkedGraphSource would not depend on the rootSegments at all, and the ChunkedGraphLayer would be responsible for requesting chunks for the current set of root segments.

I imagine one reason for the current design of ChunkedGraphChunk is to be able to re-use the existing mechanism in SliceView for deciding which chunks to load, and at the time graphene was first implemented in the seung branch there wasn't a viable alternative.

However, note that the logic in SliceView has been factored out so that it can be used mostly independently --- see `forEachPlaneIntersectingVolumetricChunk` defined in sliceview/base.ts.

chrisj: It seems we had some intentions to make bulk /leaves requests for multiple segments within a region but we never used it in neuroglancer and we don't see a need for it currently so your proposition sounds good.

Could you guide me on the change needed to make ChunkedGraphLayer request chunks per root segment?

I see GrapheneMultiscaleVolumeChunkSource.getChunkedGraphSources which creates a single source which references visibleSegments. That is passed to the frontend ChunkedGraphLayer which has getSources method which we just return this.sources for.  Then on the backend ChunkedGraphLayer, we have filterVisibleSources which has some hacks I was hoping could be replaced.

So one guess is that we no longer generate the sources immediately on layer creation and getSources creates a new list of sources based on the current visible segments?

jbms: There should still be one ChunkSource per chunk graph level, but its chunks should be keyed by position+root id, rather than just position (SliceViewChunkSource keys chunks by position only, but since this will be operating separately from SliceView you don't have to do that here).  What you describe of having a separate chunk source per root id could work, but it is simpler to just use a single chunk source.

src/neuroglancer/segmentation_user_layer.ts
jbms: Thinking about this more, I see that there is a need to have some dummy renderlayer that will be picked up by the SliceView.

src/neuroglancer/sliceview/chunked_graph/backend.ts
chrisj: ChunkedGraphChunkSource.getChunk takes a key chunkGridPosition, it generates a key `vec3Key(chunkGridPosition)` creates a chunk passing it all the root segments and returns it. Do I need to change my getChunk to `getChunk(chunkGridPosition: Float32Array, segment: Uint64) {`?

jbms: Yeah that's what I have in mind.

chrisj: Would ChunkedGraphLayer still extend SliceViewRenderLayer or do I need to replace SliceViewRenderLayer and all other SliceView classes/factor it out.

chrisj: you also mentioned above the "dummy renderlayer" that could be attached to a slice view which seems applicable to the chunked graph layer.

chrisj: @jbms I just added a commit that makes I think the changes to ChunkedGraphChunk that you were expecting, though I wasn't sure how to trigger what I called `handleChunkedGraphLayer`. I did a crude hack to get something working. I'm not sure what changes I should be making on the front end. I'm not sure what I should be doing as far as a ChunkManager as well. 

One thing I noticed is that `ChunkedGraphChunkSource.computeChunkBounds` is essentially identical to `VolumeChunkSource`.

I think the changes once you get inside ChunkedGraphChunk are a lot cleaner than our previous implementation.

jbms: Yes --- I think `computeChunkBounds` could be factored out and made a non-member function --- then it wouldn't have to be duplicated.


See my comment on your commit here:

https://github.com/google/neuroglancer/commit/35916543e1e8877dea4af598fa3cba892125fd1f#r71754865

I think with a relatively small change the integration of ChunkedGraphLayer can be cleaner.

Overall this looks very close to being ready.

src/neuroglancer/segmentation_graph/source.ts
jbms: Pinging again regarding this.

chrisj:
@jbms thanks Jeremy, I've mostly been looking at the skeleton code. I'm a little stuck at the moment, I'm inside of updateChunkPriorities in my equivalent of the backend SkeletonLayer. It seems like the place that I should run forEachPlaneIntersectingVolumetricChunk though I'm not sure where I should get/calculate the arguments for that function.

I see class AnnotationSpatiallyIndexedRenderLayerBackend extends withChunkManager uses most of the same parameters for forEachVisibleAnnotationChunk. Main difference is that it has a renderScaleTarget and forEachPlaneIntersectingVolumetricChunk requires a chunkLayout.

jbms:
The transformedSources should be set up on the frontend side and then sent to the frontend, similar to what is done in the annotation layer:

https://github.com/google/neuroglancer/blob/5787d018c2dbf9804fefb2da73a7c37a49b89842/src/neuroglancer/annotation/renderlayer.ts#L685

https://github.com/google/neuroglancer/blob/5787d018c2dbf9804fefb2da73a7c37a49b89842/src/neuroglancer/annotation/renderlayer.ts#L699

https://github.com/google/neuroglancer/blob/5787d018c2dbf9804fefb2da73a7c37a49b89842/src/neuroglancer/annotation/backend.ts#L347

Is it correct that there is only a single chunk grid used for fetching equivalences, i.e. it is not a multiscale grid?  In that case you don't need to worry about any logic for deciding which transformedSources to use, since there will only be one.

Then you can invoke forEachPlaneIntersectingVolumetricChunk similarly to how it is done by sliceview/backend.ts:

https://github.com/google/neuroglancer/blob/5787d018c2dbf9804fefb2da73a7c37a49b89842/src/neuroglancer/sliceview/backend.ts#L118

chrisj:
thanks
that is correct it is a single grid

chrisj:
Here is my current progress on using SliceViewPanelRenderLayer. It is in a messy state. https://github.com/seung-lab/neuroglancer/commit/668911cad5e5092149a1b63e1df9f6409ea7622d

I'm stuck at trying to make the `forEachPlaneIntersectingVolumetricChunk` call. I have `view.projectionParameters` from the attachment but it is a plain `ProjectionParameters` rather than a `SliceViewProjectionParameters` which lacks the centerDataPosition property used to calculate the priority. I just noticed you have a `// FIXME: calculate priority` which is perhaps due to the same reason. I'll see if I can get by with priority = 0.

We were also using SliceViewProjectionParameters.pixelSize to stop loading the chunked graph when zoomed far out within `filterVisibleSources`. Assuming I can access the pixel size or some other zoom calcuation, would I handle that now by just doing an early return within `updateChunkPriorities` or on the frontend?

jbms:
I believe the `projectionParameters` is actually a `SliceViewProjectionParameters` since it is coming from SliceViewBackend --- you just need to cast it.  Probably there is room for improving the typing here.

For the annotation layer I just didn't get around to implementing the priority computation.

For the cutoff, it is probably best to do that on the backend, though I suppose then you would have to duplicate the logic in the frontend in order to show a message.

chrisj:
Thanks, it was a SliceViewProjectionParameters. It is mostly working now. I have a question about `loadedSubsource.getRenderLayerTransform`.  Previously I was passing in `segmentationRenderlayer.channelCoordinateSpace` which is why I was doing that `someSegmentationRenderLayer` call. Though transform is also a property on the layer display states and those are being done without a coordinate space argument. Was my use of `segmentationRenderlayer.channelCoordinateSpace` unnecessary?

Also, does displayDimensionRenderInfo make sense for the chunked graph render layer? I'm not sure if it is doing anything.

jbms:
`channelCoordinateSpace` is not supported for segmentation layers, only for image layers, so you can just rely on the default value of `emptyValidCoordinateSpace` when calling `getRenderLayerTransform`.

`displayDimensionRenderInfo` is still applicable, since it determines which of the global dimensions correspond to the cross section being displayed, and what scale factors are applied.

src/neuroglancer/gpu_hash/hash_table.ts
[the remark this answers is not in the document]
chrisj: All of the calls to existing calls to unsafeEntries are from `[Symbol.iterator]` and being made from many `for...of` loops. Should [Symbol.iterator] be safe?

If so I would probably do something like

```typescript
[Symbol.iterator]() {
  return this.hashTable.entries();
}

unsafeEntries() {
  return this.hashTable.unsafeEntries();
}
```
```typescript
for (const [key, value] of segmentStatedColors.unsafeEntries()) {
  j[key.toString()] = serializeColor(unpackRGB(value.low));
}
```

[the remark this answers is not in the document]
jbms: Yes that sounds good.

src/neuroglancer/segmentation_graph/source.ts
```diff
@@ -23,18 +23,18 @@ import {RenderLayerTransformOrError} from 'neuroglancer/render_coordinate_transf
...
 import { ChunkedGraphLayerDisplayState } from 'neuroglancer/sliceview/chunked_graph/frontend';
 
-export enum VISIBLE_SEGMENT_TYPE {
-  SIMPLE_EQUIVALENCES = 1,
-  HIGH_BIT_REPRESENTATIVE_EXCLUDED,
-  HIGH_BIT_REPRESENTATIVE_ONLY
+export enum VisibleSegmentType {
+  SIMPLE_EQUIVALENCES = 1,          // all members of a set are accessed as individual elements
+  HIGH_BIT_REPRESENTATIVE_EXCLUDED, // equivalences with the highest bit set are ignored
+  HIGH_BIT_REPRESENTATIVE_ONLY      // only the representive
```
chrisj, line 29: HIGH_BIT_REPRESENTATIVE_ONLY doesn't seem to be the correct way to describe the option we need for graphene. VISIBLE_SEGMENT_ONLY would be more fitting. We only want to interact with the visibleSegment set. Graphene replaces non root segments in that set with root segments.

Graphene has it's own isBaseSegmentId. It would probably makes sense to move that into a method on SegmentationGraphSourceConnection that graphene can override.

Another possible name could work BASE_SEGMENT_EXCLUDED/BASE_SEGMENT_ONLY

jbms: Currently this enum controls two things:

- The behavior of DisjointUint64Sets, namely which element of the set is considered the representative (either min or max)
- Which segments in the visible equivalence sets are iterated over by `forEachVisibleSegment` in `segmentation_display_state/base.ts`.  That is used for the existing mesh, skeleton, and annotation layers.

Maybe it wold be better to decouple them, so that we just have a MIN_REPRESENTATIVE / MAX_REPRESENTATIVE option for DisjointUint64Sets, and something else for `forEachVisibleSegment`.

I believe for graphene you want `forEachVisibleSegment` to only iterate over the representative.  So we could potentially have:

VisibleSegmentEquivalencePolicy with

INCLUDE_ALL
EXCLUDE_REPRESENTATIVE
REPRESENTATIVE_ONLY

Would that work?

chrisj: Yes, those names make sense to me. Though it may make sense to additionally modify forEachVisibleSegment to ignore base segments in the graphene REPRESENTATIVE_ONLY case. Currently there is an extraneous mesh manifest request made for the original segment id before graphene swaps it with the root id. Since it is a non root id, the manifest just returns an empty array. It is not significant.

VisibleSegmentType is also used in the MergeSegmentsTool. We may support MergeSegmentsTool in the future though our first iteration of graphene proofreading will probably only use our graphene specific merge tool.

chrisj:
I did a sight refactor of computeChunkBounds 1f5483fb8ae8f63da2d4e28e4455d43230c01156, though it is a bit janky. I added two interfaces and the function requires you to pass in two temp variables.

src/neuroglancer/segmentation_graph/source.ts
jbms: That is a good point.  One tricky thing if you need a call to the graph source to determine which segments to include is that the graph source currently exists only on the frontend, but this function is used on the backend as well.

jbms:
> I did a sight refactor of computeChunkBounds [1f5483f](https://github.com/google/neuroglancer/commit/1f5483fb8ae8f63da2d4e28e4455d43230c01156), though it is a bit janky. I added two interfaces and the function requires you to pass in two temp variables.

Thanks --- looks okay -- an alternative would be to pass in the `source` (and define an interface for that, including the temp variables) rather than the separate arguments.  I think either way is okay.

src/neuroglancer/segmentation_graph/source.ts
chrisj: Would you be opposed to having a back-end counterpart to the chunked graph?

One issue I find with our implementation is that our chunked_graph had graphene logic abstracted out. I think the reason was because it was based on sliceview so it was moved with the rest of the sliceview code. I would be surprised if anyone else decided to use chunked_graph and I don't think it is generic enough to support other implementations.

I am currently assuming it is a graphene datasource in chunked_graph backend because of the isBaseSegment call. Do you think I should moved chunked_graph into graphene? and maybe combine GrapheneChunkedGraphChunkSource and ChunkedGraphChunkSource.

jbms: Yes I think combining could make sense, as could having a backend counterpart for a graph connection.

chrisj:
Regarding computeChunkBounds, I tried that and seemed a little cleaner to me though so I went with it.

I looked into adding a backend for SegmentationGraphSource or SegmentationGraphSourceConnection but when I looked at the call to forEachVisibleSegment in MeshLayer, it seemed it would be tricky to access the graphConnection from that class. So I decided to just handle the minor issue (of making unnecessary requests) in GrapheneMeshSource.download by avoiding the request if it was a base segment. bd1254ed6bc896a9d1244b55780b9dd5606b07fd

If we split up VisibleSegmentEquivalencePolicy/HighBitRepresentative, how would we go about accessing the VisibleSegmentEquivalencePolicy since we are currently pulling that enum value by accessing the disjoint set within forEachVisibleSegment
```const highBitRepresentative = segmentEquivalences.disjointSets.highBitRepresentative.value;```

**Comment by jbms** [comment]
The `VisibleSegmentEquivalencePolicy` could be accessed via some property on the graphSource or graphConnection.  But a simpler option is probably to keep just the single value in `disjointSets`, but make it a bitvector so that we can store several flags as part of the same number.

**Comment by jbms** [comment]
To access the graphSource or graphConnection, it would need to be added to the `VisibleSegmentsState` interface.  But since you have a workable solution without that, it is probably simpler to not add a backend counterpart yet.

**Comment by chrisj** [comment]
I was considering adding the graph to the VisibleSegmentsState but it seemed like a major change in the intention of VisibleSegmentsState and graph, as the graph used the state to communicate with the other layers. It does seem to be the easiest way to implement it.

I made VisibleSegmentEquivalencePolicy a bitvector. Not sure if it is how you were expecting.

The default value 0 = min representative, simple equivalences. The logic in forEachVisibleSegment is a little odd, I'm assuming you either have `NONREPRESENTATIVE_EXCLUDED` or `REPRESENTATIVE_EXCLUDED`. I don't mind making them separate properties.

File: src/neuroglancer/ui/segment_split_merge_tools.ts [hunk 10]
```diff
@@ -58,9 +58,9 @@ export class MergeSegmentsTool extends Tool<SegmentationUserLayer> {
       if (!Uint64.equal(segmentSelectionState.selectedSegment, mappedAnchorSegment)) return;
       const base = segmentSelectionState.baseSelectedSegment;
       const isBase = isBaseSegmentId(base);
-      const highBitVal = segmentEquivalences.disjointSets.highBitRepresentative.value;
```
jbms: This logic isn't quite right, since currently the bitvector only talks about representative vs non-representative, not high bit.  But in practice this will still work.  Can you just add a TODO, like:

// TODO: This would ideally rely on a separate `HIGH_BIT_REPRESENTATIVE` flag, but it nonetheless still works correctly for nggraph and local equivalences.

src/neuroglancer/chunk_manager/backend.ts
```diff
@@ -386,7 +386,7 @@ export class ChunkSource extends ChunkSourceBase {
   }
 }
 
-function startChunkDownload(chunk: Chunk) {
+export function startChunkDownload(chunk: Chunk) {
```
jbms, line 389: Is this change to export these functions still needed?

src/neuroglancer/chunk_manager/backend.ts
```diff
@@ -411,7 +411,7 @@ function startChunkDownload(chunk: Chunk) {
           });
 }
 
-function cancelChunkDownload(chunk: Chunk) {
+export function cancelChunkDownload(chunk: Chunk) {
```
jbms, line 414: Does this function still need to be exported?

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -0,0 +1,746 @@
...
+
+class GrapheneMeshSource extends
+(WithParameters(WithCredentialsProvider<SpecialProtocolCredentials>()(MeshSource), MeshSourceParameters)) {
+  getFragmentKey(objectKey: string|null, fragmentId: string) {
+    objectKey;
+    return getGrapheneFragmentKey(fragmentId);
+  }
+}
+
+function resolvePath(a: string, b: string) {
```
jbms, line 62: Just use the function from precomputed/frontend instead.

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -0,0 +1,746 @@
...
+      // Dealing with a prehistoric graph server with no version information
+      this.supported_api_versions = [0];
+    }
+    if (PYCG_APP_VERSION in this.supported_api_versions === false) {
+      const redirectMsgBox = new StatusMessage();
+      const redirectMsg = `This Neuroglancer branch requires Graph Server version ${
+          PYCG_APP_VERSION}, but the server only supports version(s) ${
+          this.supported_api_versions}.`;
+
+      if (location.hostname.includes('neuromancer-seung-import.appspot.com')) {
```
jbms, line 105: I understand the value of this, but I'd rather not have this included.

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -0,0 +1,746 @@
...
+  }
+}
+
+interface GrapheneMultiscaleVolumeInfo extends MultiscaleVolumeInfo {
+  dataUrl: string;
+  app: AppInfo;
+  graph: GraphInfo;
+}
+
+function parseSpecialUrlOld(url: string): string { // TODO: brought back old parseSpecialUrl
```
jbms, line 136: What is the reason for this?

chrisj: It looks like it was needed because we hadn't adopted cancellableFetchSpecialOk which does the transformation from gs to https. It seems to be working correctly if I swap to that function and use undefined for the credentialsProvider but this goes back to your original comment about a function where I had previously set the credentialsProvider to undefined if the url was GCS. I need to confirm when we may need credentials for such requests.

jbms:
Thanks for your perseverance with this.  Any remaining fixes can be done in a followup PR.

PR #395, opened 2022-06-07
feat(datasource/graphene) added multicut tool

chrisj:
I wanted to get this tool implemented first since it requires quite a few changes to neuroglancer

There may still be some sections I can clean up though I think there are some important things that are worth discussing again. I'll also add code comments for smaller issues.

* the 2d coloring features we added. I think this was our last discussion on it https://github.com/google/neuroglancer/pull/340#issuecomment-964545143
* how the graphene tab is added. it now responds to the graph source being disabled however the tool and the tool keybind continue to work. Sven suggested that we make the subsource required (non disable-able)
* the graphene state. we discussed this in detail so it might be in good shape
* the multicut segment list (I'll add a code comment about a temporary hack for this)


src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -83,6 +83,7 @@ const MESH_SILHOUETTE_RENDERING_JSON_KEY = 'meshSilhouetteRendering';
 const LINKED_SEGMENTATION_GROUP_JSON_KEY = 'linkedSegmentationGroup';
 const LINKED_SEGMENTATION_COLOR_GROUP_JSON_KEY = 'linkedSegmentationColorGroup';
 const SEGMENT_DEFAULT_COLOR_JSON_KEY = 'segmentDefaultColor';
+// const HIGHLIGHT_COLOR_JSON_KEY = 'highlightColor';
```
chrisj, line 86: I have this commented out because multicut overrides the highlightColor, so I save the previous highlightColor (as I do with a few other values and they would have similar issues) so it can be restored when multicut tool is closed. The issue is that if neuroglancer is refreshed, those prior values are lost. The easiest thing in this case was to just not have highlightColor be part of the state. The robust solution would be also adding tempHightlightColor, tempSegmentStatedColors, etc.

src/neuroglancer/segmentation_display_state/base.ts
```diff
@@ -31,6 +31,10 @@ export interface VisibleSegmentsState {
   temporarySegmentEquivalences: SharedDisjointUint64Sets;
   useTemporaryVisibleSegments: SharedWatchableValue<boolean>;
   useTemporarySegmentEquivalences: SharedWatchableValue<boolean>;
+
+  // TODO explanation, currently they cause non focus segments to be hidden in 2d, segmentation rendered transparent, with everything else darkened
+  // to bring attention to the EM for that segment
+  focusSegments: Uint64Set;
```
chrisj, line 37: I initially had this in SegmentationDisplayState since it was just used for changing the 2d rendering, but then I also decided I could use this set to prevent users from accidentally deselecting the active multicut segment while the tool is active. I do that in the graph connection visible segment listener by re-adding the segment if it is removed (and is in the focus set).

src/neuroglancer/ui/annotations.ts
```diff
@@ -1333,7 +1334,7 @@ export function UserLayerWithAnnotationsMixin<TBase extends {new (...args: any[]
         boolean {
       if (state.annotationId === undefined) return false;
       const annotationLayer = this.annotationStates.states.find(
-          x => x.sourceIndex === state.annotationSourceIndex &&
+          x => x.subsourceIndex === state.annotationSubsourceIndex &&
```
chrisj, line 1337: This temp hack was made because there was an issue with the selection side panel displaying "Annotation not found" on the second annotationState in the same AnnotationLayerView. I don't know the proper way to handle this.

src/neuroglancer/sliceview/volume/segmentation_renderlayer.ts
```diff
@@ -149,6 +166,7 @@ uint64_t getMappedObjectId(uint64_t value) {
 `);
     }
     builder.addUniform('highp uvec2', 'uSelectedSegment');
+    builder.addUniform('highp uvec2', 'uBaseSelectedSegment');
```
jbms, line 169: I don't think there is a need for a separate parameter --- instead `uSelectedSegment` can just be set to the base selected segment if `baseSegmentHighlighting` is enabled.

src/neuroglancer/sliceview/volume/segmentation_renderlayer.ts
```diff
@@ -220,26 +274,33 @@ uint64_t getMappedObjectId(uint64_t value) {
...
     gl.uniform2ui(shader.uniform('uSelectedSegment'), selectedSegmentLow, selectedSegmentHigh);
+    gl.uniform2ui(shader.uniform('uBaseSelectedSegment'), baseSelectedSegmentLow, baseSelectedSegmentHigh);
     gl.uniform1ui(
         shader.uniform('uShowAllSegments'),
         visibleSegments.hashTable.size || !ignoreNullSegmentSet ? 0 : 1);
     this.hashTableManager.enable(
         gl, shader,
         segmentationGroupState.useTemporaryVisibleSegments.value ? this.gpuTemporaryHashTable :
                                                                    this.gpuHashTable);
+    this.hashTableManagerFocusSegments.enable(gl, shader, this.gpuHashTableFocusSegments);
```
jbms, line 303: condition on `parameters.hasFocusSegments`

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -702,3 +1002,317 @@ class SliceViewPanelChunkedGraphLayer extends SliceViewPanelRenderLayer {
...
+    activation.bindAction('swap-group', event => {
+      event.stopPropagation();
+      multicutState.swapGroup();
+    });
+
+    activation.bindAction('set-anchor', event => {
+      event.stopPropagation();
+      
+      const {segmentSelectionState: {baseValue, value}} = this.layer.displayState;
+      if (!baseValue || !value) return; // could this happen or is this just for type checking
```
jbms, line 1272: If the user triggers this action while not hovering over a segment, then this condition will occur.

jbms:
I think it would be preferable if we can avoid adding `focusSegments` to `VsibleSegmentsState`, since it seems to be very specific to a single use case.  If I understand correctly, the desired behavior could instead be implemented by adding the following:

- `tempSegmentStatedColors2d`
- `useTempSegmentStatedColors2d`
- ``tempSegmentDefaultColor2d`
-

chrisj:
I think that would work as long as I changed segment stated colors to support alpha, multiply that alpha value by the existing alpha value in the 2d shader. I would ignore it in 3d though I know you want consistency between 2d and 3d. For multicut, making the target segment in 2d completely transparent is useful, but for 3d, it would be useless. I'll start working on that but let me know if you would prefer a different approach.

I would also have to add notSelectedColor that you previously suggested.

chrisj:
Just re-read your comment and realized you specified 2d variable names so the only talking point is transparency.

jbms:
There is an existing `segmentDefaultColor` that I believe does what the previously mentioned `notSelectedColor` was proposed to do (i.e. catch all for `segmentStatedColors`) --- therefore I proposed `tempSegmentDefaultColor2d` because I believe you need to control that just for 2d.

Agreed that `segmentStatedColors` would need to also include an alpha value which would be multiplied in the way you describe.

chrisj:
Rather than multiply the alpha, I chose to override the alpha if the alpha value is non-zero. If the alpha is zero, it uses the existing alpha.

This made it easier to darken outside the multicut neuron without having to also increase the off opacity, which would warrant a temporary off opacity.

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -702,3 +1011,313 @@ class SliceViewPanelChunkedGraphLayer extends SliceViewPanelRenderLayer {
...
+
+export class GrapheneTab extends Tab {
+  private annotationLayerView =
+      this.registerDisposer(new MulticutAnnotationLayerView(this.layer, this.layer.annotationDisplayState));
+
+  constructor(public layer: SegmentationUserLayer) {
+    super();
+    const {graphConnection} = layer;
+    if (graphConnection instanceof GraphConnection) {
+      const states = graphConnection.initializeAnnotations(this.layer);
```
jbms, line 1085: I think the intention is that the multicut markers are shown all the time if present, regardless of whether the annotation tab is open or the multicut tool has been activated.  If so, then `initializeAnnotations` should instead happen earlier, e.g. when the graph connection is created.

src/neuroglancer/ui/annotations.ts
jbms: It would be simpler if the annotations were stored in a single annotation source, since the existing annotation code assumes a single annotation source per subsource.  However, I guess the reason for two annotation sources is so that they can be displayed in separate colors, and also shown in separate lists.  I think there are two possible solutions:

1. Use a single annotation source for both "red" and "blue" points.  The colors could be handled via an annotation property.  Some change would be needed to filter them in the list, though.
2. Add an additional optional string identifier in order to identify multiple annotation sources within the same subsource --- basically the same as you have here, except using `string | undefined` as the identifier type in place of `number`.

I think option 2 is probably the best way to go.

src/neuroglancer/datasource/graphene/frontend.ts
jbms: Yes, I think adding a `layer` paremeter to `SegmentationGraphSource.connect` would be reasonable.

I'm not too fond of making subsources required, as I'd rather preserve user control over that ---- but I think it is reasonable if the meshes don't work with the graph subsource disabled.  Potentially there could be a way to specify that a subsource depends on another subsource, e.g. to prevent the mesh subsource from being enabled without the graph subsource.  But I think that may not be worth the trouble to implement.

src/neuroglancer/ui/annotations.ts
chrisj: For option 2, any idea what I should call it? I was going to name it annotationSubsourceId but in annotations.ts it has.  `state.annotationSubsource = annotationLayer.subsourceId` I could rename the existing AnnotationLayerState.subsourceId to just subsource.

What does AnnotationLayerState currently use subsourceId and subsourceIndex for?

src/neuroglancer/datasource/graphene/frontend.ts
chrisj: does https://github.com/google/neuroglancer/pull/395/commits/b16feb6df1ac27844f6a2fa1102e02fe96073ef4 seem reasonable? I replaced segmentsState with layer since you can access the segmentState from the layer and having them as two separate arguments opens it up to sending unconnected arguments. Also, transform was accidentally added in the graphene view PR so I removed that.

src/neuroglancer/sliceview/volume/segmentation_renderlayer.ts
```diff
@@ -247,13 +274,17 @@ uint64_t getMappedObjectId(uint64_t value) {
           gl, shader,
           useTemp ? this.gpuTemporaryEquivalencesHashTable : this.gpuEquivalencesHashTable);
     }
-    if (segmentDefaultColor === undefined) {
-      this.segmentColorShaderManager.enable(gl, shader, segmentColorHash);
+    let activeSegmentDefaultColor = tempSegmentDefaultColor2d || segmentDefaultColor;
+    if (activeSegmentDefaultColor) {
+      // convert to vec4, is there a better way?
+      activeSegmentDefaultColor = vec4.clone([...activeSegmentDefaultColor]);
+      gl.uniform4fv(shader.uniform('uSegmentDefaultColor'), activeSegmentDefaultColor);
```
jbms, line 281: Can just use `gl.uniform4f` instead and avoid creating a new vec4.

src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -289,6 +303,9 @@ class SegmentationUserLayerDisplayState implements SegmentationDisplayState {
   selectSegment = this.layer.selectSegment;
   transparentPickEnabled = this.layer.pick;
   baseSegmentColoring = new TrackableBoolean(false, false);
+  baseSegmentHighlighting = new TrackableBoolean(false, false);
+  useTempSegmentStatedColors2d =
+      this.layer.registerDisposer(SharedWatchableValue.make(this.layer.manager.rpc, false)); // where does this line belong? Also should it be in segmentationColorGroupState?
```
jbms, line 308: I think it depends on whether the temporary colors should also apply to other layers for which the segment color state is linked to this layer.  Probably not too likely to matter either way, so I think it is fine to leave it here for now.

src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -493,6 +513,19 @@ export class SegmentationUserLayer extends Base {
               for (const renderLayer of graphRenderLayers) {
                 loadedSubsource.addRenderLayer(renderLayer);
               }
+              const graphTab = segmentationGraph.tab;
```
jbms, line 516: One thing that I am a bit concerned about is how it will work to have this tab selected, or dragged out as a separate panel, given that it is added only once the data source is loaded, rather than when the layer is first created.

Have you tested that?

chrisj: I just tested pinning the graph tab and toggling the graph source off and back on. The pinned tab stays, the tool UI inside the tab hides but the list of annotations remains. It reverts back to normal when toggling the source back on.

jbms: What about if you reload the browser while the graph tab is selected or dragged out as a separate panel?

chrisj: It disappears/reverts back to the source tab.

src/neuroglancer/datasource/graphene/frontend.ts
jbms: Yes, thanks.

src/neuroglancer/segmentation_user_layer.ts
jbms: I think to get better behavior with this tab a slightly different approach may be required --- I think for the layer tabs we may need a concept of a "hidden tab", which is shown only when selected --- if not selected, no tab is shown.

Then when a graph subsource with a graph tab is activated, the tab is marked non-hidden.

Additionally the graph connection would not supply the actual top-level tab, but instead something that would included as the contents of the tab, e.g. via DependentViewWidget.

chrisj: What about making the graphConnection property on the segmentationLayer a watchable value? That way I can create a tab class that listens to that and replaces the graph tab content and toggle it's own visibility. Rather than accessing the tab from the segmentationLayer.

Also currently I had the tab method on the source rather than connection but it shouldn't be a problem moving it to the connection.

Here is my current WIP https://github.com/seung-lab/neuroglancer/commit/3d6c0726972557da4add5fa926209ade3795393f

jbms: Yes I think that change makes sense.

To be clear I was suggesting that the `DependentViewWidget` stuff be done by segmentation_user_layer, not by graphene-specifically --- graphene would supply its tabContents, and the segmentation layer would ensure it is visible when that graph source is activated.

chrisj: I converted layer.graphConnection to a WatchableValue but I ended up using `layer.displayState.segmentationGroupState.value.graph` instead. I forgot that the graph was accessible through that. Should I revert the change to graphConnection or does it seem useful to keep?

https://github.com/google/neuroglancer/pull/395/commits/f25632e3ed592dfd9bb908357e43dca5ecd459c3

src/neuroglancer/segmentation_graph/source.ts
```diff
@@ -30,12 +32,35 @@ export enum VisibleSegmentEquivalencePolicy {
   NONREPRESENTATIVE_EXCLUDED = 1 << 2, // filter out non representative elements when iterating over visible segments
 }
 
+export class SegmentationGraphSourceTab extends Tab {
+  constructor(public layer: SegmentationUserLayer) {
+    super();
+    const {element} = this;
+
+    element.classList.add('neuroglancer-annotations-tab');
```
chrisj, line 40: this class adds display flex which is needed for my annotation list to render properly. Setting the property on the parent inside the graphene provided dependent view requires a !important to override existing display: contents.

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -577,6 +873,30 @@ class GrapheneGraphSource extends SegmentationGraphSource {
     return this.graphServer.getRoot(segment);
   }
 
+  tabContents(layer: SegmentationUserLayer) {
+    return new DependentViewWidget(
```
chrisj, line 877: I'm using 2 DependentViewWidgets. One in the SegmentationGraphSourceTab, the other provided by the graph implementation. It seemed to make sense here rather than trying to just return an element because I need the graph and the context.

src/neuroglancer/segmentation_graph/source.ts
chrisj: I took a look at this again and I think I was mistaken, it seems display flex needs to be on the neuroglancer-tab-content div.

I think tabContents needs to be able to add/toggle classes on the tab element. I could pass the tab or the element to tabContents.

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -874,10 +873,30 @@ class GrapheneGraphSource extends SegmentationGraphSource {
     return this.graphServer.getRoot(segment);
   }
 
-  tab(layer: SegmentationUserLayer) {
-    return new GrapheneTab(layer);
+  tabContents(layer: SegmentationUserLayer) {
+    return new DependentViewWidget(
+                                layer.displayState.segmentationGroupState.value.graph,
+                                (graph, parent, context) => {
+                                  if (graph === undefined) return;
```
jbms, line 880: I think SegmentationGraphSourceTab should be responsible for only including the tab contents if these conditions hold, so that individual graph sources don't need these checks.  Then you wouldn't need a second layer of dependent view widget.

chrisj: The primary reason I used it was that it was also supplying the context, should I just add context as a parameter? Also perhaps element/tab to solve the css class issue I mentioned above?

```tabContents(layer: SegmentationUserLayer, context: DependentViewContext, tab: SegmentationGraphSourceTab) {```

Comment (chrisj) [reply src/neuroglancer/datasource/graphene/frontend.ts]: this is what I was thinking https://github.com/google/neuroglancer/pull/395/commits/3cfdc6dda9c9cdfcada5156952ec8245266081da

File: src/neuroglancer/ui/layer_side_panel_state.ts [hunk 14]
```diff
@@ -70,7 +70,10 @@ export class UserLayerSidePanelState extends RefCounted {
     for (const id of this.tabs) {
       const {hidden} = this.layer.tabs.options.get(id)!;
       if (hidden) {
-        this.registerDisposer(hidden.changed.add(() => this.panels.updateTabs()));
+        this.registerDisposer(hidden.changed.add(() => {
+          this.panels.updateTabs();
+          this.tabsChanged.dispatch();
```
chrisj, line 75: I had to add tabsChanged.dispatch because otherwise the tab label was not appearing on refresh if the graph tab was alone in its own panel.

I also noticed an issue where turning off the graph source would cause the graph tab to stay selected even though the tab content and tab label went away.

I tried to fix it by adding this code in TabView.makeTabs but then I got the unwanted behavior of the graph tab being deselected on reload since the tab isn't immediately available. Seems like that could be a tricky fix without adding complexity.
```ts
// swap to the first tab if the selected tab becomes hidden
    const tabIds = this.tabs.value.map(x => x.id);
    const selectedTabId = this.selectedTab.value;
    if (selectedTabId && tabIds.length && !tabIds.includes(selectedTabId)) {
      this.selectedTab.value = tabIds[0];
    }
```

chrisj: another issue I noticed in regards to disabling the graph source is that the tool keybind stays active and the multicut tool is assuming there is a graph source. I can handle that in the tool, is that what you would suggest? Or something similar to the DependentView, a DependentTool.

jbms: I think for consistency with how reloading is handled, it makes sense that disabling the graph source should not deselect the graph tab if it is already selected.

I would say that if a hidden tab is selected, its tab label should also be shown (to avoid confusion).

Similarly any key bindings for tools that require a graph should also stay, but the tool itself should not work.  If the multicut tool is actually activated, rather than just bound to a key, if the graph source is disabled then I would say the tool should become deactivated (but still remain as a key binding).  I think it would be reasonable to just handle this behavior in the tool itself for now, though if you think it would be cleaner to factor out the logic somehow that would also be fine.

jbms: To be clear, if a "hidden" tab is *not* selected then its label should be hidden.  But if a "hidden" tab is selected, then its label should be shown.  The content hiding doesn't necessarily need to be implemented by `tab_view` itself since the SegmentationGraphSourceTab can already handle that with its DependentViewWidget.

chrisj: Does this seem right? https://github.com/google/neuroglancer/pull/395/commits/6248302d75d8d590997a2476fe77b39b021150ee
I did have to listen to the selectedTab event to hide the tab if they select off of a hidden tab. I could also add a check to only redo the tabs if the previous value was a hidden tab.

jbms: Thanks --- I think it would be slightly better to only regenerate the tabs if the previous value was a hidden tab.

chrisj: done in https://github.com/google/neuroglancer/pull/395/commits/cede0520e9c70ad77228b859ce8c5a8c2a660088

chrisj: forgot to update the prev variable, fixed in https://github.com/google/neuroglancer/commit/bc2a8d7a4980e5ce50832ba55f660619e6445bc8

jbms:
Thanks!

PR #408, opened 2022-08-17
feat(segmentation_user_layer): hide segments

chrisj:
the checkbox now hides the visible segments but does not remove it from the list.

this is a rough version to gauge your thoughts on the feature/implementation. I threw in a random removal method: ctrl + left click.

I think a more proper implementation would make the visibleSegment act similar to a sql view on top of selected segments but still conform to the uint64S, and selectedSegments could be changed to be a Uint64Map: segment id -> visibility (0/1). Another option is to change selectedSegments to hiddenSegments. There are many ways the JSON state could work, obviously it must be backwards compatible.

I didn't touch the temporary sets but it seems it would justify another temporary set and useTemporaryVisibleSegments would enable both.

### DIFF
--- a/src/neuroglancer/datasource/graphene/frontend.ts
+++ b/src/neuroglancer/datasource/graphene/frontend.ts
@@ -646,11 +646,11 @@ class GraphConnection extends SegmentationGraphSourceConnection {
       public state: GrapheneState) {
     super(graph, layer.displayState.segmentationGroupState.value);
     const segmentsState = layer.displayState.segmentationGroupState.value;
-    segmentsState.visibleSegments.changed.add((segmentIds: Uint64[]|Uint64|null, add: boolean) => {
+    segmentsState.selectedSegments.changed.add((segmentIds: Uint64[]|Uint64|null, add: boolean) => {
       if (segmentIds !== null) {
         segmentIds = Array<Uint64>().concat(segmentIds);
       }
-      this.visibleSegmentsChanged(segmentIds, add);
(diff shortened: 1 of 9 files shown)


jbms:
Thanks.

In general I think this approach is good --- as far as the in-memory representation, I think having `selectedSegments` and `visibleSegments`, rather than separate `visibleSegments` and `hiddenSegments`, is reasonable, since compared to a map or other representation, it minimizes the code changes required.

As far as the temporary sets, what do you see as the use cases for those?  I would think even during a merge we would probably not want to change which segments are displayed in the UI --- for example, currently the temporary sets don't affect what is displayed in the UI.  I guess the issue is just that for graphene there is a need to load the chunks for the temporary visible segments; it looks like at the moment you have it tracking the selected segments, rather than the visible segments.  Is there a reason to load the supervoxels for selected but not visible segments?

chrisj:
With the way it is set up, I think you are right, temporary selected segment isn't necessary, since there is no restriction that visible segments also have to be selected.

I have a few fixes for this PR that I made on another branch and a little bit more to clean up.

I thought it would make more sense to operate on selected segments and leave visible segments to be controlled primarily by the UI. The chunksource is only looking at visible segments. https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/graphene/backend.ts#L339

chrisj:
Replaced by https://github.com/google/neuroglancer/pull/425

PR #409, opened 2022-08-18
feat(python): support TensorStore data sources in LocalVolume

jan-matthis:
Adds support for TensorStore data sources in `LocalVolume`.

### DIFF
--- a/python/neuroglancer/local_volume.py
+++ b/python/neuroglancer/local_volume.py
@@ -20,6 +20,10 @@ import threading
 
 import numpy as np
 import six
+try:
+    import tensorstore as ts
+except ImportError:
+    ts = None
 
 from . import downsample, downsample_scales
 from .chunks import encode_jpeg, encode_npz, encode_raw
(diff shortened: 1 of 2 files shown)


python/neuroglancer/local_volume.py
```diff
@@ -83,8 +87,11 @@ def __init__(self,
         """
         super(LocalVolume, self).__init__()
         self.token = make_random_token()
-        self.data = data
-        self.shape = data.shape
+        if ts is not None and isinstance(data, ts.TensorStore):
+            self.data = TensorStoreLocalVolume(data)
+        else:
+            self.data = data
+        self.shape = self.data.shape
```
jbms, line 94: Can you also add support for choosing a default `CoordinateSpace` from the additional data provided by TensorStore, namely `data.domain.labels`, and `data.dimension_units`?

Also can you add support for non-zero origin, e.g. set `voxel_offset = data.origin` and then set `data = data[ts.d[:].translate_to[0]]` ?

Finally can you add a few unit tests (just of LocalVolume directly perhaps)?

python/neuroglancer/local_volume.py
```diff
@@ -244,3 +251,80 @@ def invalidate(self):
...
+        super().__init__(data=data[ts.d[:].translate_to[0]])
+        self.dtype = self._data.dtype.numpy_dtype
+        self._origin_before_translation = data.origin
+
+    def __getitem__(self, item):
+        return self._data.__getitem__(item).read().result()
+
+    @property
+    def labels(self):
+        if 'labels' in self.schema.to_json()['domain']:
```
jbms, line 307: Rather than convert the schema to json it would be better to just say:

`if not any(self.data.domain.labels)`

jan-matthis: Thanks, done

python/neuroglancer/local_volume.py
```diff
@@ -244,3 +251,80 @@ def invalidate(self):
...
+        super().__init__(data=data[ts.d[:].translate_to[0]])
+        self.dtype = self._data.dtype.numpy_dtype
+        self._origin_before_translation = data.origin
+
+    def __getitem__(self, item):
+        return self._data.__getitem__(item).read().result()
+
+    @property
+    def labels(self):
+        if any(self._data.domain.labels):
```
jbms, line 307: Sorry, I gave the wrong advice actually --- technically some dimensions can be labeled and others can be unlabeled (empty string).  So maybe it is better to instead just choose `dN` labels to fill in any missing ones.

python/neuroglancer/local_volume.py
```diff
@@ -244,3 +251,80 @@ def invalidate(self):
...
+        return self._default_labels
+
+    @property
+    def origin(self):
+        return self._origin_before_translation
+
+    @property
+    def scales(self):
+        try:
+            return [d.multiplier for d in self._data.dimension_units]
```
jbms, line 318: The self._data.dimension_units property itself could raise an error, in which case we should just use the default scales.  But also each unit can be `None` to indicate unknown.  In that case we should convert it to `1` as a default.  Similarly for `units`.

PR #410, opened 2022-08-30
Choose best available BossDB requested encoding

j6k4m8:
BossDB can render u8 imagery as JPEG filmstrips, which saves ~50% on egress filesize. All other datatypes should be returned as npygz.

This also adds support for non-u8 imagery :)

> cc @BrockWester @dxenes1 @hannahgooden


jbms:
Okay.  Is it safe to assume that uin8 won't be used for segmentations/masks or other data where the exact values need to be reproduced?

j6k4m8:
That is correct!

PR #412, opened 2022-09-03
Update write_annotations.py

kleined:
Update to allow this module to be compatible with python versions < 3.8. This is important because Colab runs python 3.7.


python/neuroglancer/write_annotations.py
```diff
@@ -24,6 +24,10 @@
 import numpy as np
 import struct
 
+try:
```
jbms, line 27: Can just unconditionally import typing_extensions; on newer versions of Python that is an alias to `typing.Literal`.

Note: module name is misspelled here.

Also need to add `typing_extensions` to install_requires in setup.py

kleined:
Thanks!

PR #425, opened 2022-11-01
feat(segmentation_user_layer): starred segments

chrisj:
Replaces https://github.com/google/neuroglancer/pull/408

We aren't planning to have this feature added to neuroglancer until the multi list view feature is ready but I think it is good to work out issues with the implementation of the starred/selected/saved list, whichever you prefer to call it, I'll rename.

I'll add some PR comments to the code. Some of the UI changes still don't play entirely nicely with the segment property map. I may make those fixes on this PR or on the other.

As far as the next step, getting the multi list view working. I'm not sure the best way to implement it. I got it working by copying most of the UI code and pasting it underneath. It was very repetitive. Maybe I should abstract it into some UI class? The ListSource only required me to pass an empty watchable string for the query param and I added an additional param to control whether it should include the explicit segments.

### DIFF
--- a/src/neuroglancer/datasource/graphene/backend.ts
+++ b/src/neuroglancer/datasource/graphene/backend.ts
@@ -45,6 +45,9 @@ import { DisplayDimensionRenderInfo } from 'neuroglancer/navigation_state';
 import { forEachVisibleSegment } from 'neuroglancer/segmentation_display_state/base';
 import { computeChunkBounds } from 'neuroglancer/sliceview/volume/backend';
 
+import {Uint64OrderedSet} from 'neuroglancer/uint64_ordered_set';
+console.log('test - this is needed becuase Uint64OrderedSet is not usecd by any backend threads', Uint64OrderedSet);
+
 function getVerifiedFragmentPromise(
     credentialsProvider: SpecialProtocolCredentialsProvider,
     chunk: FragmentChunk,
--- a/src/neuroglancer/datasource/graphene/backend.ts
+++ b/src/neuroglancer/datasource/graphene/backend.ts
@@ -45,9 +45,6 @@ import { DisplayDimensionRenderInfo } from 'neuroglancer/navigation_state';
 import { forEachVisibleSegment } from 'neuroglancer/segmentation_display_state/base';
 import { computeChunkBounds } from 'neuroglancer/sliceview/volume/backend';
 
-import {Uint64OrderedSet} from 'neuroglancer/uint64_ordered_set';
-console.log('test - this is needed becuase Uint64OrderedSet is not usecd by any backend threads', Uint64OrderedSet);
-
 function getVerifiedFragmentPromise(
     credentialsProvider: SpecialProtocolCredentialsProvider,
     chunk: FragmentChunk,
--- a/src/neuroglancer/segmentation_user_layer.ts
+++ b/src/neuroglancer/segmentation_user_layer.ts
@@ -170,7 +170,7 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
 
   localGraph = new LocalSegmentationGraphSource();
   visibleSegments = this.registerDisposer(Uint64Set.makeWithCounterpart(this.layer.manager.rpc));
-  selectedSegments = this.registerDisposer(Uint64OrderedSet.makeWithCounterpart(this.layer.manager.rpc));
+  selectedSegments = this.registerDisposer(new Uint64OrderedSet());
 
   segmentPropertyMap = new WatchableValue<PreprocessedSegmentPropertyMap|undefined>(undefined);
   graph = new WatchableValue<SegmentationGraphSource|undefined>(undefined);
--- a/src/neuroglancer/segmentation_user_layer.ts
+++ b/src/neuroglancer/segmentation_user_layer.ts
@@ -72,8 +72,7 @@ const BASE_SEGMENT_COLORING_JSON_KEY = 'baseSegmentColoring';
 const IGNORE_NULL_VISIBLE_SET_JSON_KEY = 'ignoreNullVisibleSet';
 const MESH_JSON_KEY = 'mesh';
 const SKELETONS_JSON_KEY = 'skeletons';
-const SELECTED_SEGMENTS_JSON_KEY = 'segments';
-const VISIBLE_SEGMENTS_JSON_KEY = 'visibleSegments';
+const SEGMENTS_JSON_KEY = 'segments';
 const EQUIVALENCES_JSON_KEY = 'equivalences';
 const COLOR_SEED_JSON_KEY = 'colorSeed';
 const SEGMENT_STATED_COLORS_JSON_KEY = 'segmentColors';
@@ -120,23 +119,20 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
       this.localGraph.restoreState(value);
     });
 
-    verifyOptionalObjectProperty(specification, SELECTED_SEGMENTS_JSON_KEY, segmentsValue => {
-      const {segmentEquivalences, selectedSegments} = this;
-      // backwards compatibility, assume all selected are visible
-      const missingVisible = !Object.prototype.hasOwnProperty.call(specification, VISIBLE_SEGMENTS_JSON_KEY);
+    verifyOptionalObjectProperty(specification, SEGMENTS_JSON_KEY, segmentsValue => {
+      const {segmentEquivalences, selectedSegments, visibleSegments} = this;
       parseArray(segmentsValue, value => {
-        let id = Uint64.parseString(String(value), 10);
-        selectedSegments.add(segmentEquivalences.get(id));
-        if (missingVisible) {
-          this.visibleSegments.add(segmentEquivalences.get(id));
+        let stringValue = String(value);
+        const hidden = stringValue.startsWith('!');
+        if (hidden) {
+          stringValue = stringValue.substring(1);
         }
-      });
-    });
-    verifyOptionalObjectProperty(specification, VISIBLE_SEGMENTS_JSON_KEY, segmentsValue => {
-      const {segmentEquivalences, visibleSegments} = this;
-      parseArray(segmentsValue, value => {
-        let id = Uint64.parseString(String(value), 10);
-        visibleSegments.add(segmentEquivalences.get(id));
+        const id = Uint64.parseString(stringValue, 10);
+        const segmentId = segmentEquivalences.get(id);
+        selectedSegments.add(segmentId);
+        if (!hidden) {
+          visibleSegments.add(segmentId);
+        }
       });
     });
   }
@@ -144,17 +140,17 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
   toJSON() {
     const x: any = {};
     x[HIDE_SEGMENT_ZERO_JSON_KEY] = this.hideSegmentZero.toJSON();
-    let {selectedSegments} = this;
+    let {selectedSegments, visibleSegments} = this;
     if (selectedSegments.size > 0) {
-      x[SELECTED_SEGMENTS_JSON_KEY] = this.selectedSegments.toJSON();
+      x[SEGMENTS_JSON_KEY] = [...selectedSegments].map(segment => {
+        if (visibleSegments.has(segment)) {
+          return segment.toString();
+        } else {
+          return '!' + segment.toString();
+        }
+      });
     } else {
-      x[SELECTED_SEGMENTS_JSON_KEY] = [];
-    }
-    let {visibleSegments} = this;
-    if (visibleSegments.size > 0) {
-      x[VISIBLE_SEGMENTS_JSON_KEY] = this.visibleSegments.toJSON();
-    } else {
-      x[VISIBLE_SEGMENTS_JSON_KEY] = [];
+      x[SEGMENTS_JSON_KEY] = [];
     }
     let {segmentEquivalences} = this;
     if (this.localSegmentEquivalences && segmentEquivalences.size > 0) {
--- a/src/neuroglancer/segmentation_user_layer.ts
+++ b/src/neuroglancer/segmentation_user_layer.ts
@@ -60,6 +60,7 @@ import {renderScaleLayerControl} from 'neuroglancer/widget/render_scale_widget';
 import {colorSeedLayerControl, fixedColorLayerControl} from 'neuroglancer/widget/segmentation_color_mode';
 import {registerLayerShaderControlsTool} from 'neuroglancer/widget/shader_controls';
 import {registerSegmentSelectTools} from 'neuroglancer/ui/segment_select_tools';
+import {Uint64OrderedSet} from 'neuroglancer/uint64_ordered_set';
 
 const SELECTED_ALPHA_JSON_KEY = 'selectedAlpha';
 const NOT_SELECTED_ALPHA_JSON_KEY = 'notSelectedAlpha';
@@ -146,6 +147,8 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
     let {selectedSegments} = this;
     if (selectedSegments.size > 0) {
       x[SELECTED_SEGMENTS_JSON_KEY] = this.selectedSegments.toJSON();
+    } else {
+      x[SELECTED_SEGMENTS_JSON_KEY] = [];
     }
     let {visibleSegments} = this;
     if (visibleSegments.size > 0) {
@@ -171,7 +174,8 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
 
   localGraph = new LocalSegmentationGraphSource();
   visibleSegments = this.registerDisposer(Uint64Set.makeWithCounterpart(this.layer.manager.rpc));
-  selectedSegments = this.registerDisposer(Uint64Set.makeWithCounterpart(this.layer.manager.rpc));
+  selectedSegments = this.registerDisposer(Uint64OrderedSet.makeWithCounterpart(this.layer.manager.rpc));
+
   segmentPropertyMap = new WatchableValue<PreprocessedSegmentPropertyMap|undefined>(undefined);
   graph = new WatchableValue<SegmentationGraphSource|undefined>(undefined);
   segmentEquivalences = this.registerDisposer(SharedDisjointUint64Sets.makeWithCounterpart(
--- a/src/neuroglancer/segmentation_user_layer.ts
+++ b/src/neuroglancer/segmentation_user_layer.ts
@@ -71,7 +71,8 @@ const BASE_SEGMENT_COLORING_JSON_KEY = 'baseSegmentColoring';
 const IGNORE_NULL_VISIBLE_SET_JSON_KEY = 'ignoreNullVisibleSet';
 const MESH_JSON_KEY = 'mesh';
 const SKELETONS_JSON_KEY = 'skeletons';
-const SEGMENTS_JSON_KEY = 'segments';
+const SELECTED_SEGMENTS_JSON_KEY = 'segments';
+const VISIBLE_SEGMENTS_JSON_KEY = 'visibleSegments';
 const EQUIVALENCES_JSON_KEY = 'equivalences';
 const COLOR_SEED_JSON_KEY = 'colorSeed';
 const SEGMENT_STATED_COLORS_JSON_KEY = 'segmentColors';
@@ -93,9 +94,21 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
   constructor(public layer: SegmentationUserLayer) {
     super();
     const {specificationChanged} = this;
-    this.visibleSegments.changed.add(specificationChanged.dispatch);
     this.hideSegmentZero.changed.add(specificationChanged.dispatch);
     this.segmentQuery.changed.add(specificationChanged.dispatch);
+
+    const {visibleSegments, selectedSegments} = this;
+    visibleSegments.changed.add(specificationChanged.dispatch);
+    selectedSegments.changed.add(specificationChanged.dispatch);
+    selectedSegments.changed.add((x, add) => {
+      if (!add) {
+        if (x) {
+          visibleSegments.delete(x);
+        } else {
+          visibleSegments.clear();
+        }
+      }
+    });
   }
 
   restoreState(specification: unknown) {
@@ -106,23 +119,39 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
       this.localGraph.restoreState(value);
     });
 
-    verifyOptionalObjectProperty(specification, SEGMENTS_JSON_KEY, segmentsValue => {
+    verifyOptionalObjectProperty(specification, SELECTED_SEGMENTS_JSON_KEY, segmentsValue => {
+      const {segmentEquivalences, selectedSegments} = this;
+      // backwards compatibility, assume all selected are visible
+      const missingVisible = !Object.prototype.hasOwnProperty.call(specification, VISIBLE_SEGMENTS_JSON_KEY);
+      parseArray(segmentsValue, value => {
+        let id = Uint64.parseString(String(value), 10);
+        selectedSegments.add(segmentEquivalences.get(id));
+        if (missingVisible) {
+          this.visibleSegments.add(segmentEquivalences.get(id));
+        }
+      });
+    });
+    verifyOptionalObjectProperty(specification, VISIBLE_SEGMENTS_JSON_KEY, segmentsValue => {
       const {segmentEquivalences, visibleSegments} = this;
       parseArray(segmentsValue, value => {
         let id = Uint64.parseString(String(value), 10);
         visibleSegments.add(segmentEquivalences.get(id));
       });
     });
-    verifyOptionalObjectProperty(
-        specification, SEGMENT_QUERY_JSON_KEY, value => this.segmentQuery.restoreState(value));
   }
 
   toJSON() {
     const x: any = {};
     x[HIDE_SEGMENT_ZERO_JSON_KEY] = this.hideSegmentZero.toJSON();
+    let {selectedSegments} = this;
+    if (selectedSegments.size > 0) {
+      x[SELECTED_SEGMENTS_JSON_KEY] = this.selectedSegments.toJSON();
+    }
     let {visibleSegments} = this;
     if (visibleSegments.size > 0) {
-      x[SEGMENTS_JSON_KEY] = visibleSegments.toJSON();
+      x[VISIBLE_SEGMENTS_JSON_KEY] = this.visibleSegments.toJSON();
+    } else {
+      x[VISIBLE_SEGMENTS_JSON_KEY] = [];
     }
     let {segmentEquivalences} = this;
     if (this.localSegmentEquivalences && segmentEquivalences.size > 0) {
@@ -135,12 +164,14 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
   assignFrom(other: SegmentationUserLayerGroupState) {
     this.maxIdLength.value = other.maxIdLength.value;
     this.hideSegmentZero.value = other.hideSegmentZero.value;
+    this.selectedSegments.assignFrom(other.selectedSegments);
     this.visibleSegments.assignFrom(other.visibleSegments);
     this.segmentEquivalences.assignFrom(other.segmentEquivalences);
   }
 
   localGraph = new LocalSegmentationGraphSource();
   visibleSegments = this.registerDisposer(Uint64Set.makeWithCounterpart(this.layer.manager.rpc));
+  selectedSegments = this.registerDisposer(Uint64Set.makeWithCounterpart(this.layer.manager.rpc));
   segmentPropertyMap = new WatchableValue<PreprocessedSegmentPropertyMap|undefined>(undefined);
   graph = new WatchableValue<SegmentationGraphSource|undefined>(undefined);
   segmentEquivalences = this.registerDisposer(SharedDisjointUint64Sets.makeWithCounterpart(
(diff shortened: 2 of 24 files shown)


src/neuroglancer/widget/icon.ts
```diff
@@ -24,6 +24,19 @@ export interface MakeIconOptions {
   href?: string;
 }
 
+export interface MakeHoverIconOptions extends MakeIconOptions {
+  svgHover?: string
+}
+
+export function makeHoverIcon(options: MakeHoverIconOptions): HTMLElement {
```
chrisj, line 31: this class may be overkill, I wanted the star icon to change to a x to show it was a destructive action. I could replace it with two makeIcons.

src/neuroglancer/segmentation_user_layer.ts
```diff
@@ -92,9 +93,21 @@ export class SegmentationUserLayerGroupState extends RefCounted implements Segme
...
+
+    const {visibleSegments, selectedSegments} = this;
+    visibleSegments.changed.add(specificationChanged.dispatch);
+    selectedSegments.changed.add(specificationChanged.dispatch);
+    selectedSegments.changed.add((x, add) => {
+      if (!add) {
+        if (x) {
+          visibleSegments.delete(x);
+        } else {
+          visibleSegments.clear();
```
chrisj, line 107: originally I had this listener adding and deleting to visible segments but we switched to wanting the star button to not make the segment visible. I'm not sure if this just adds more confusion, it means that in some other places I have to specifically add to visible but not when deleting from selected.

jbms:
Can you explain a bit more what you mean by the "multi list view"?

jbms:
There are a few issues with regard to the JSON state representation:
- It would be nice to avoid duplicating the visible segment ids in both the visible list and the starred list, since this could lead to a significant increase in state size.  An easy solution would be to just have a starredSegments list that excludes the visible segments.  But this does not allow us to preserve order.
- Currently Neuroglancer treats the visible segments set as an unordered set; when they are displayed, they are ordered either numerically or ordered by a segment property.  However, there might be use cases where it is useful to treat the starred list as ordered --- for example when it effectively functions as a (modifiable) history of segments that were explored, then preserving the order may be helpful to users.  The starred list might also be used as an ordered list of segments for a user to check.   However, this does introduce new UI challenges.  It also means we would need a single combined segments list --- since the segment ids are already json-encoded as strings rather than numbers, (to allow 64-bit numbers to be stored), possibly it could be represented as: ["1", "!2", "3", "4"], where "!" prefix indicates a non-visible segment.  But this representation, while concise, may be confusing.

chrisj:
> Can you explain a bit more what you mean by the "multi list view"?
I mean viewing the segment list query results together with what was the visible list but now the selected list, stacked vertically.

Are you suggesting changing starred segments from a set to a list or something else to allow it to preserve order? I know some who would prefer it was ordered by insertion.

I like the idea of using !, seems like it would be hard to find a better solution.

jbms:
> > Can you explain a bit more what you mean by the "multi list view"?
> I mean viewing the segment list query results together with what was the visible list but now the selected list, stacked vertically.

Previously, the visible list was concatenated with the query results as a single VirtualList list.  But with the introduction of the starred list, I think that should be removed ---instead there should be one VirtualList for the query results, and one VirtualList for the starred segments.  The code in ui/segment_list.ts likely needs to be refactored some to avoid duplication of code.

> 
> Are you suggesting changing starred segments from a set to a list or something else to allow it to preserve order? I know some who would prefer it was ordered by insertion.

Yes, exactly.  I think the tricky part is that if we support preserving order, we also want to support:
- User can control whether it is ordered by one of the existing columns (numeric id, or a segment property) or ordered by the "insertion/saved order"
- User can drag and drop to reorder, but this only really makes sense when viewing by insertion/saved order.

> 
> I like the idea of using !, seems like it would be hard to find a better solution.

sdorkenw:
Hi. Great discussion; I love the ordering idea - that is an often requested feature!

I wanted to way in on the representation for visible segments. I see the issue with duplication when storing visible segment IDs in a separate list. However, the introduction of a "!" would introduce compatibility issues (with old links) and affects other code bases that are directly reading neuroglancer's state. In some way it is violating what one expects to receive when reading out neuroglancers state. Being able to interact with the state easily is a great feature and breaking changes there are rough. 

Instead, I would suggest a separate list with the indices of the visible segments (to avoid storing uint64 strings which take up much more space) or a separate boolean list (which should compress well).

jbms:
It is indeed partially a breaking change, in that if someone tries to use a state with starred non-visible segments (which may easily be added even without a specific intention to use the feature) with existing software not expecting the new syntax, it won't work.  On the other hand, if starred non-visible segments are not used, the state remains identical to before.

However, I'm not sure what a better solution is.  The best backward compatibility would be provided by having a separate list of starred non-visible segments, but then we lose the ability to represent an interleaved order of visible and non-visible starred segments.  If we require that visible segments are always listed first, then that is non-issue.

I see some issues with the two other proposed representations: both are also backwards incompatible, since old software would interpret the starred non-visible segments as also visible.  Additionally, both representations would make the state significantly less human readable and human editable.  Also there is currently no compression of the state in the URL.

ceesem:
I don't see why visibility is significantly different than storing other per-segment properties that we will absolutely want, for example, user-defined colors. I could see two similar options. We could have a property list with keys as segment ids and values as key-value pairs with property name (e.g. `visibility`) and property value (e.g. `false`), potentially with default values elided for efficiency. The segment list could then be kept as-is. 

Alternatively, this id:property list could replace the segment list entirely since it contains segment ids and is potentially redundant. However, this would be both breaking and would throw away information when a segment id is removed (which might or might not be desired)

jbms:
There already is support in the state for user-defined colors (segmentColors), they just can't be changed in the ui yet.

While visibility could be seen as just another property, backwards compatibility and size of the state representation needs to be taken into account.

ceesem:
Colors are one of a number of per-segment properties I can imagine being extremely useful not only for visualization features, but also other mechanisms for interacting with the data. For example, in a graphene layer it might be convenient to store a supervoxel corresponding to how a segment was selected in order to update the root id while maintaining intended segment.

However, I feel like I'm not understanding something about your two main concerns guiding the design decisions here, and maybe it would help us if you could be a bit more explicit about where you are coming from.

1) I don't understand why backwards compatibility is a critical concern. As a nice-to-have, of course I agree, but I struggle to see a significant use case in people bringing Neuroglancer states from one deployment to another, older one. Do you have a different experience here?

2) I understand the general goal of managing state length, but in my experience the size of the state representation is dominated by annotations when doing most types of work. It is hard to imagine a useful state having more than several tens to maybe hundreds of segments selected at once, but as someone who uses Neuroglancer almost every day, I am frequently putting hundreds to thousands of annotations in a state. As often as not, these annotations are driven in real time by analysis and it would not be trivial to replace even with the sort of annotation query tools we are hoping to build in soon. With that aspect of the state baked into, in my mind, an approach that is only marginally less efficient than an existing method (e.g. segmentColor, which already duplicates segment ids) has a very modest cost in terms of state length and has the virtue of being both more human-readable and more flexible going forward. I would like to understand better what you define what an acceptable state length to be, because I think there's a mismatch in our mental models here.

jbms:
> Colors are one of a number of per-segment properties I can imagine being extremely useful not only for visualization features, but also other mechanisms for interacting with the data. For example, in a graphene layer it might be convenient to store a supervoxel corresponding to how a segment was selected in order to update the root id while maintaining the intended segment.
> 
> However, I feel like I'm not understanding something about your two main concerns guiding the design decisions here, and maybe it would help us if you could be a bit more explicit about where you are coming from.
> 
> 1) I don't understand why backwards compatibility is a critical concern. As a nice-to-have, of course I agree, but I struggle to see a significant use case in people bringing Neuroglancer states from one deployment to another, older one. Do you have a different experience here?

The json state is the only stable api provided by neuroglancer.  I am not concerned with allowing a new state to be used with an older neuroglancer version, but the states are also read and written by numerous other tools, sometimes using the Neuroglancer python api, and sometimes directly.  While complete backward compatibility cannot be guaranteed, I'd like to avoid breakage without a compelling reason.
> 
> 2) I understand the general goal of managing state length, but in my experience the size of the state representation is dominated by annotations when doing most types of work. It is hard to imagine a useful state having more than several tens to maybe hundreds of segments selected at once, but as someone who uses Neuroglancer almost every day, I am frequently putting hundreds to thousands of annotations in a state. As often as not, these annotations are driven in real time by analysis and it would not be trivial to replace even with the sort of annotation query tools we are hoping to build in soon. With that aspect of the state baked into, in my mind, an approach that is only marginally less efficient than an existing method (e.g. segmentColor, which already duplicates segment ids) has a very modest cost in terms of state length and has the virtue of being both more human-readable and more flexible going forward. I would like to understand better what you define what an acceptable state length to be, because I think there's a mismatch in our mental models here.

There are lots of different use cases of neuroglancer.  One common one is to pseudo-agglomerate segments making up a cell by simply selecting them in a neuroglancer state.  For this use case, we may end up hundreds or thousands of segments.   With multi-resolution meshes, it is possible to view hundreds of even thousands of full cells simultaneously.  Sometimes it is desired to view all cells meeting certain criteria at once, e.g. all cells of a given type.

While there isn't a specific length limit, longer url lengths create various challenges:
- length may exceed maximum message/URL size in various systems other than the browser, used to share links
- length may exceed browser limit on display and copy/paste
- eventually length may exceed browser limit on URL length for history and navigation

Some functionality may inherently require additional state (e.g. annotations stored in the state) and then the state size increase is unavoidable.  But I'd like to avoid increasing the state size unnecessarily.

For example, segmentColors increases the state size, but you only pay that cost when using the feature.  In contrast, since visible segments automatically become starred segments, starred segments is not an optional feature.

Also, while visibility and starred status have some similarity to other segment properties, there is also the key difference that these are inherently just part of the local state, while other properties may come from a persistent data source.

In any case, though, if you can provide an example of the state representation you are thinking of, that would surely help the discussion.

ceesem:
The state representation I was thinking of would have something like:
```
{
    ....
    "segments": ["12345", "67890", "101112"],
    "segmentProperties": {
        "12345": {"color": "#ffffff", "visible": false},
        "67890": {"visible": false}
       // No entry for "101112" because it has all default values.
    }
   ...
}
```
where default values for properties (visible:true or color based on seed) would be omitted for state size. It would also be a common pattern where layers could add useful layer-specific features. It is a bit unwieldy to parse, though, and I appreciate your point that it is worth handling it differently as a common property across all segments.

chrisj:
Hi Jeremy, I haven't touched the JSON representation yet or the segmentProperties concept, but I created a Uint64OrderedSet class which preserves order, and I removed the sorting from segment_list.

None of the backend threads make use of selectedSegments but it is in the VisibleSegmentsState so the `withSegmentationLayerBackendState` expects it. I get the error `Uncaught TypeError: constructorFunction is not a constructor` which I temporarily fixed by adding `import {Uint64OrderedSet} from 'neuroglancer/uint64_ordered_set';` to `graphene/backend.ts`. I'm not sure why the same import in `segmentation_display_state/backend.ts` isn't enough.

Also, have you considered using BigInt in neuroglancer?

jbms:
> Hi Jeremy, I haven't touched the JSON representation yet or the segmentProperties concept, but I created a Uint64OrderedSet class which preserves order, and I removed the sorting from segment_list.

Thanks, that's great.  One concern I have with the current implementation is that adding a new segment takes linear time since it searches the list to see if it is already present.  Adding thousands of segments may, therefore, become rather slow.  It would be great if we can avoid that somehow, e.g. by keeping an additional Set or something.
> 
> None of the backend threads make use of selectedSegments but it is in the VisibleSegmentsState so the `withSegmentationLayerBackendState` expects it. I get the error `Uncaught TypeError: constructorFunction is not a constructor` which I temporarily fixed by adding `import {Uint64OrderedSet} from 'neuroglancer/uint64_ordered_set';` to `graphene/backend.ts`. I'm not sure why the same import in `segmentation_display_state/backend.ts` isn't enough.
> 
> Also, have you considered using BigInt in neuroglancer?

I looked into a little bit.  The biggest advantage it offers is that they can be used as keys in Map and Set.  I haven't done benchmarks myself recently, but when I looked at it some time ago, it appeared that no browser implementation had actually added any optimizations, meaning it was significantly slower than using a hand-rolled Uint64 type.  But the performance difference might be insignificant for Neuroglancer.

chrisj:
@jbms just pushed a change to continue using the hashstable. I could go further and make ordered a config option when creating a Uint64Set. `Uint64Set({ordered: true})`.

chrisj:
@jbms pinging in case you missed my last update

jbms:
Thanks Chris --- this is looking really great.  It seems the one thing missing is the separate "query results" display.

I also noticed one minor issue: in Chrome, the star icon seems to slightly change size when starred vs unstarred, which makes the list of segments shift a little bit.   It would be nice to avoid that.

See for example (localhost URL):

http://127.0.0.1:8080/#!%7B%22dimensions%22:%7B%22x%22:%5B4e-9%2C%22m%22%5D%2C%22y%22:%5B4e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%2C%22position%22:%5B515412.78125%2C356707.4375%2C2646.5%5D%2C%22crossSectionScale%22:6.049647464412947%2C%22projectionScale%22:1048576%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22precomputed://gs://h01-release/data/20210601/4nm_raw%22%2C%22tab%22:%22source%22%2C%22name%22:%224nm_raw%22%7D%2C%7B%22type%22:%22segmentation%22%2C%22source%22:%22precomputed://gs://h01-release/data/20210601/c3%22%2C%22tab%22:%22segments%22%2C%22segments%22:%5B%5D%2C%22name%22:%22c3%22%7D%5D%2C%22showSlices%22:false%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22c3%22%7D%2C%22layout%22:%224panel%22%7D

chrisj:
I looked into the the shifting segment list. It seems to happen at certain zoom levels and I think I can reproduce what you are talking about in the master branch by toggling a checkbox. It doesn't appear to me that the star icon changes size based on the fill.

https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B4e-9%2C%22m%22%5D%2C%22y%22:%5B4e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%2C%22position%22:%5B630784%2C102400%2C1024%5D%2C%22crossSectionScale%22:12.099294928825895%2C%22crossSectionDepth%22:-20%2C%22projectionScale%22:1048576%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22precomputed://gs://h01-release/data/20210601/4nm_raw%22%2C%22tab%22:%22source%22%2C%22name%22:%224nm_raw%22%7D%2C%7B%22type%22:%22segmentation%22%2C%22source%22:%22precomputed://gs://h01-release/data/20210601/c3%22%2C%22tab%22:%22segments%22%2C%22segments%22:%5B%22488974691%22%5D%2C%22name%22:%22c3%22%7D%5D%2C%22showSlices%22:false%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22c3%22%7D%2C%22layout%22:%224panel%22%7D

I was able to stop it shifting by removing the filter by label and copy columns as well as setting a fixed height for checkbox. I can keep playing around with it.

jbms:
> I looked into the the shifting segment list. It seems to happen at certain zoom levels and I think I can reproduce what you are talking about in the master branch by toggling a checkbox. It doesn't appear to me that the star icon changes size based on the fill.
> 
> https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B4e-9%2C%22m%22%5D%2C%22y%22:%5B4e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%2C%22position%22:%5B630784%2C102400%2C1024%5D%2C%22crossSectionScale%22:12.099294928825895%2C%22crossSectionDepth%22:-20%2C%22projectionScale%22:1048576%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22precomputed://gs://h01-release/data/20210601/4nm_raw%22%2C%22tab%22:%22source%22%2C%22name%22:%224nm_raw%22%7D%2C%7B%22type%22:%22segmentation%22%2C%22source%22:%22precomputed://gs://h01-release/data/20210601/c3%22%2C%22tab%22:%22segments%22%2C%22segments%22:%5B%22488974691%22%5D%2C%22name%22:%22c3%22%7D%5D%2C%22showSlices%22:false%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22c3%22%7D%2C%22layout%22:%224panel%22%7D
> 
> I was able to stop it shifting by removing the filter by label and copy columns as well as setting a fixed height for checkbox. I can keep playing around with it.

Thanks for looking into this --- you are right that it is a pre-existing bug and therefore not relevant to this PR.

chrisj:
Hi Jeremy, been a while since the last update. It is still not completely ready but I want to see if you agree with the direction it is going.

I added classes to represent the list and all of its associated UI elements. There is a base class and then one class with specific query related features, and another simple class that only represents the selected segments. Currently this feature can't easily be disabled but I can add that if we want it.

There are a few minor features I am still working on, some confirmation related code `selectSegmentConfirmationThreshold`. I'm also still trying to figure out if there is a way to get the flex box to neatly partition itself between the 1-3 scrolling elements that might be visible (query properties, query list, selected list), without resorting to javascript but I've been stuck on that for a while.

jbms:
> Hi Jeremy, been a while since the last update. It is still not completely ready but I want to see if you agree with the direction it is going.
> 
> I added classes to represent the list and all of its associated UI elements. There is a base class and then one class with specific query related features, and another simple class that only represents the selected segments. Currently this feature can't easily be disabled but I can add that if we want it.

Do you mean have an option to disable the "starred segments" feature entirely?  I don't think that is necessary --- happy to have this feature available all the time.

> 
> There are a few minor features I am still working on, some confirmation related code `selectSegmentConfirmationThreshold`. I'm also still trying to figure out if there is a way to get the flex box to neatly partition itself between the 1-3 scrolling elements that might be visible (query properties, query list, selected list), without resorting to javascript but I've been stuck on that for a while.

Normally to get flex children to evenly divide the available space you can set on the children: `height: 0px; flex-grow: 1;`.  However, this situation is a bit trickier because we ideally don't want the `neuroglancer-segment-query-result-statistics` part to take up more space than it needs, i.e. if its natural height is less than 1/3 of the remaining space then it should just have its natural height, not 1/3 of the available space.  Currently that is implemented by setting `max-height: 40%; flex-shrink: 0;`.

Ideally the two segment lists also wouldn't take up more space than they need, but given how the virtual list works, that is probably challenging and may cause issues, so I think it would be sufficient to just hide each list entirely if it is empty.

I think it may be tricky to do this purely in CSS, but I think we can accomplish most of it with CSS, and just set a CSS class or CSS variable or `data-` attribute from javascript to indicate a little bit about the state, e.g. whether there is a query result list.  Then the CSS can use that to perhaps set the statistics element max-height to either 40% or say 20%.

jbms:
One thing I noticed that using the checkbox at the top of the lower pane to toggle the visibility of all starred segments is somewhat slow now (there is a noticeable delay):

http://127.0.0.1:8080/#!%7B%22dimensions%22:%7B%22x%22:%5B4e-9%2C%22m%22%5D%2C%22y%22:%5B4e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%2C%22position%22:%5B401533.84375%2C360436.40625%2C3061.5%5D%2C%22crossSectionScale%22:29.759463449745702%2C%22projectionOrientation%22:%5B-0.07988395541906357%2C-0.08492271602153778%2C-0.19145002961158752%2C0.9745529890060425%5D%2C%22projectionScale%22:599625.4731252204%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22precomputed://gs://h01-release/data/20210601/4nm_raw%22%2C%22tab%22:%22source%22%2C%22name%22:%224nm%20EM%22%2C%22visible%22:false%7D%2C%7B%22type%22:%22segmentation%22%2C%22source%22:%5B%7B%22url%22:%22precomputed://gs://h01-release/data/20210601/c3%22%2C%22subsources%22:%7B%22default%22:true%2C%22bounds%22:true%2C%22properties%22:true%2C%22mesh%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%2C%7B%22url%22:%22precomputed://gs://lichtman-h01-49eee972005c8846803ef58fbd36e049/goog14r0s5c3_new_props/segment_properties%22%2C%22subsources%22:%7B%22default%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%5D%2C%22panels%22:%5B%7B%22row%22:1%2C%22flex%22:1.55%2C%22size%22:479%2C%22tab%22:%22segments%22%7D%5D%2C%22segments%22:%5B%221100054524%22%2C%221115430292%22%2C%2212237931142%22%2C%221333290325%22%2C%221538274151%22%2C%221539076840%22%2C%221594648509%22%2C%221638188509%22%2C%221828951844%22%2C%221915887451%22%2C%221988993337%22%2C%2220070214646%22%2C%222090806103%22%2C%222134549398%22%2C%222178704414%22%2C%222294780853%22%2C%222339328448%22%2C%222499107339%22%2C%222499384877%22%2C%222557789796%22%2C%222673254402%22%2C%2227622860459%22%2C%2227651872764%22%2C%2227870683066%22%2C%222791142865%22%2C%2228000894735%22%2C%2228045909614%22%2C%2228378958224%22%2C%2228409678489%22%2C%2228452985548%22%2C%2228525770719%22%2C%2228643309290%22%2C%2228672203772%22%2C%2228802547903%22%2C%2228803598117%22%2C%2228918216929%22%2C%2229021270843%22%2C%2229182786959%22%2C%2229238417446%22%2C%2229298236361%22%2C%222935896346%22%2C%2229459096371%22%2C%2229618159554%22%2C%2229765396306%22%2C%2229925423252%22%2C%2229938695074%22%2C%2229969547791%22%2C%2230101233972%22%2C%223023633866%22%2C%2230376944711%22%2C%2230406000355%22%2C%223052908678%22%2C%2230668773752%22%2C%2230767987832%22%2C%2230871567933%22%2C%2230974109956%22%2C%2231031989065%22%2C%2231032951251%22%2C%2231061717348%22%2C%2231133932587%22%2C%2231149133165%22%2C%223125564306%22%2C%2231658722916%22%2C%2232444986357%22%2C%223430740099%22%2C%223504577656%22%2C%223518943222%22%2C%223573728110%22%2C%2236123853342%22%2C%2236153829678%22%2C%2236241229806%22%2C%2236284756264%22%2C%2236445761568%22%2C%2236590837294%22%2C%2236619747807%22%2C%2236620463168%22%2C%2236633005581%22%2C%223664545249%22%2C%2236662790455%22%2C%2236678107375%22%2C%2236763714196%22%2C%2236809473223%22%2C%2236822073774%22%2C%2236822262086%22%2C%2236852896228%22%2C%2237173591638%22%2C%2237218168369%22%2C%2237318668413%22%2C%2237319791473%22%2C%2237420743412%22%2C%2237421824682%22%2C%2237463700916%22%2C%2237536734693%22%2C%2237581252436%22%2C%2237594612366%22%2C%2237668755920%22%2C%2237770758892%22%2C%2237801070911%22%2C%2237930756265%22%2C%2237958718097%22%2C%2238018479939%22%2C%2238092682277%22%2C%2238106873604%22%2C%2238208029710%22%2C%2238222309348%22%2C%2238339219897%22%2C%2238368158152%22%2C%2238383402139%22%2C%223839697992%22%2C%2238413115427%22%2C%2238543488421%22%2C%2238573083867%22%2C%2238586823171%22%2C%2238602358951%22%2C%2238717924646%22%2C%2238863556782%22%2C%2238863862489%22%2C%223897678996%22%2C%2238994907574%22%2C%2239023934118%22%2C%2239038491013%22%2C%2239052800624%22%2C%2239081564265%22%2C%2239271712983%22%2C%2239271916939%22%2C%2239359991385%22%2C%223941569746%22%2C%2239505037109%22%2C%2239694250521%22%2C%2240043667040%22%2C%2240102203154%22%2C%2240247497000%22%2C%2240378483178%22%2C%2240407875435%22%2C%2240464862282%22%2C%2241354296446%22%2C%2241630488832%22%2C%224217863883%22%2C%224318903980%22%2C%2245701832369%22%2C%224580553370%22%2C%224698412002%22%2C%224784821941%22%2C%2248079386659%22%2C%224917456198%22%2C%224961186471%22%2C%225062824784%22%2C%225499847526%22%2C%2258866575372%22%2C%2259536235974%22%2C%225965326176%22%2C%226024781859%22%2C%22664433854%22%2C%22707068981%22%2C%22794630620%22%2C%22823395680%22%2C%22853035674%22%2C%22910562342%22%2C%22105686375323%22%5D%2C%22segmentQuery%22:%22%3Cid%22%2C%22colorSeed%22:4270253886%2C%22name%22:%22c3%20segmentation%22%7D%5D%2C%22layout%22:%223d%22%2C%22selection%22:%7B%22row%22:2%2C%22visible%22:false%7D%7D

chrisj:
I fixed the performance issue and a few bugs. I introduced `forEachQueryResultSegmentIdGenerator`, maybe I can get replace `forEachQueryResultSegmentId` with the generator version. I struggled to return an IterableIterator using `forEachQueryResultSegmentId` because of it's use of a callback.

I also was looking closer at `SegmentListSource` and I think my changes there are not complete. I noticed for example that toggling visibility for a segment does not trigger `SegmentListSource.changed`.

src/neuroglancer/ui/segment_list.ts
```diff
@@ -82,34 +87,38 @@ class SegmentListSource extends RefCounted implements VirtualListSource {
...
         } else {
           splices.push(
-              ...getMergeSplices(explicitSegments, newSortedVisibleSegments, Uint64.compare));
+              ...getMergeSplices(explicitSegments, newSelectedSegments, Uint64.compare));
         }
-        this.explicitSegments = newSortedVisibleSegments;
+        this.explicitSegments = newSelectedSegments;
         changed = true;
       } else {
         splices.push({retainCount: this.explicitSegments.length, deleteCount: 0, insertCount: 0});
```
chrisj, line 116: What is the condition and reason for this line?

src/neuroglancer/ui/segment_list.ts
```diff
@@ -82,34 +87,38 @@ class SegmentListSource extends RefCounted implements VirtualListSource {
...
       }
       this.explicitSegmentsVisible = true;
```
chrisj, line 118: github won't let me tag line 132 but what is the reason for `this.explicitSegments = explicitIds;` from the query result since it seems explicitSegmentsVisible is always false if there is a query.

chrisj (commented):
I added a few comments so I can better understand SegmentListSource

src/neuroglancer/ui/segment_list.ts
jbms: The way it is currently implemented, is that there are two parts of the list, the `explicitSegments` part and the matches part.  The `explicitSegments`  lists the segment ids directly, while the matches list the segments by their index into the segment property map.  The `explicitSegments` part is used for two purposes:

1. If the query is unconstrained, then for showing the visible segments.
2. If the user specifies explicit segment ids as the query, i.e. `12345, 6789`.  These have to be in the `explicitSegments` portion because there is not necessarily an entry in the segment property map for these ids.

chrisj:
One thing I just realized is that starred segments (previously visible before this PR) are no longer shown at the top of segment property list but maybe that is fine because they have their place in the starred segment list?

Added this commit which refactors SegmentListSource. https://github.com/google/neuroglancer/pull/425/commits/34d7abf31b8fa9f546a5055a62794460470f5734

Not sure if it is the right way to go but it felt like most of the complicated logic in the list source was related to the query. The starred segment list really only needed the rendering logic. There is probably some further simplification that can be done.

jbms:
Yes, agreed that it is no longer necessary/helpful to show the starred segments at the top of the query results list --- they now have a dedicated list.

chrisj:
I code related to the pinned segments and the visible list and fixed a bug with the updateStatus

src/neuroglancer/datasource/graphene/backend.ts
```diff
@@ -45,6 +45,9 @@ import { DisplayDimensionRenderInfo } from 'neuroglancer/navigation_state';
 import { forEachVisibleSegment } from 'neuroglancer/segmentation_display_state/base';
 import { computeChunkBounds } from 'neuroglancer/sliceview/volume/backend';
 
+import {Uint64OrderedSet} from 'neuroglancer/uint64_ordered_set';
```
jbms, line 48: It seems to me that we don't need the starred segments on the worker thread at all.

If for some reason that is not possible, then the correct way to add it is to add it to config/bundle-config.js in the `chunkWorkerModules` array.

jbms:
Thanks --- this is looking great.  One remaining issue: if there are no segment properties, and the query is empty, currently the `neuroglancer-preview-list` still takes up half of the height.  Instead it should be hidden.

chrisj:
> Thanks --- this is looking great. One remaining issue: if there are no segment properties, and the query is empty, currently the `neuroglancer-preview-list` still takes up half of the height. Instead it should be hidden.

I changed it so it no longer occupies space when empty, though it still takes up an equal portion of space if there is at least one element.

jbms:
Thanks.  I think one last change is to the Python API/wrappers --- instead of exposing `segments` as a `set`, it should now be a class that inherits from `dict`, where the value associated with each key (segment id) indicates whether it is visible or not.  `dict` preserves order in Python.  We would just need to support assignment/conversion from a list/set, in which case we would just add all elements as visible segments.

src/neuroglancer/datasource/graphene/backend.ts
chrisj: I just pushed [3197e91](https://github.com/google/neuroglancer/pull/425/commits/3197e9135c749d3eaa18a07ab185054e2de9d58e)

selectedSegments is in VisibleSegmentsState, it seems tricky to remove. forEachVisibleSegment is used by the chunk workers, at least by graphene, and expects a VisibleSegmentsState.

As long as I don't call `Uint64OrderedSet.makeWithCounterpart`, there is no error though it does mean selectedSegments is undefined if someone tried to access it from a chunk worker.

chrisj:
[986b4d3](https://github.com/google/neuroglancer/pull/425/commits/986b4d3106846a506b27b092f62d0d9ed40fcfe0)

Extending typed_map seemed to work well. I needed to override __init__  and to_json to do the conversion from ! array to a dict. Also, __setitem__ wasn't using the value with a placeholder comment next to None. I'm confused how typed_map was being used with the value only being passed to set_wrapped.

I also added a comment above:
https://github.com/google/neuroglancer/pull/425#discussion_r1163367745

jbms:
Thanks --- sorry for the delay in reviewing.  If you can rebase this to address the merge conflict due to the batch merge PR, it is ready to go in.

jbms:
This has now been merged: https://github.com/google/neuroglancer/commit/e431ad048421879ed28759a51923ae51770456e1

PR #450, opened 2023-03-16
feat(datasource/graphene) added refresh meshes tool

chrisj:
It would be nice to clear out GrapheneMeshSource.chunksNotFound when hiding a mesh to prevent memory growth. Any suggestion on how you would do that since MeshSource doesn't have access to visibleSegments?


jbms:
Rather than force the user to manually refresh after waiting some amount of time, how about just automatically retrying a couple times, e.g. after 2 seconds, 4 seconds, 8 seconds, 16 seconds?  Ideally the server could just wait until the mesh is finished when it receives a request, but I imagine that may be more challenging to implement.

Issue #446, opened 2023-03-02
Placeholder Precomputed Scales


william-silversmith:
Hi Jeremy,

My lab has been using "placeholder" mips to keep different versions of datasets numerically aligned in "mip" sequence. Recently, an improvement to neuroglancer made it so that these placeholders with no data display black.

https://github.com/google/neuroglancer/commit/796cc7dbeb9610e3b1203493d185d87c15aed680

Would it be possible to support either toggling off the fill value or alternatively, supporting a `"hidden": true` attribute on scales to avoid displaying them in Neuroglancer? @chrisj brought this to my attention.

Thank you so much!

jbms:
How about setting their shape to [0,0,0]?

Although if it is the first level that would be problematic.

jbms:
The first level is used by Neuroglancer to determine the clip bounds, so if it has a shape of `[0,0,0]` then that won't work.

I think a `hidden` attribute would be fine.

Do you want it to completely ignore the scale, e.g. basically behave as if the scale did not exist, or still use the first scale (even if hidden) to determine clip bounds and units to use for the global coordinate space by default?

jbms:
I think it would be most natural in Neuroglancer to just have "hidden" mean completely ignored, in which case the clip bounds would instead come from the first visible scale.  Presumably that would only differ by a small amount.  Do you think that would work for you?

william-silversmith:
I think one thing we want from this is for Neuroglancer to display the coordinates as hidden mip 0 coordinates, but display the image from mip 1. This way we can easily convert between different datasets. For example, sometimes an aligned dataset is rendered at mip 2 but we want to be able to cross reference between layers easily.

jbms:
Okay, sounds reasonable.  Would you care to create a pull request to implement this?

chrisj:
This issue is discussing the same problem https://github.com/google/neuroglancer/issues/444

I have a commit to fix this https://github.com/seung-lab/neuroglancer/commit/a7edaddb0ba0f296f797463a6aaefdc06c850177#diff-048ab67a423ce920e92e4cff52db026f6361491c6b7189eb558f3a7c42e14759R208

Not sure if it is an appropriate place to put it but seems to work. Also needs to be updated if we swap to using the hidden attribute.

unidesigner:
@chrisj It seems like there is agreement between @jbms and @william-silversmith about the hidden attribute. Would you care about submitting a PR to implement this as you are very familiar with the neuroglancer codebase? Thank you!

PR #451, opened 2023-03-24
feat: adds hidden attribute to ScaleInfo for placeholder resolutions

dxenes1:
Hi Jeremy, 

I've been following the thread between you and Chris on #446 and wanted to offer this PR as a potential solution to the placeholder issue we're having with the FAFB dataset. Feel free to disregard it if you have a better solution in mind!

This PR adds an optional boolean `hidden` attribute to the ScaleInfo. If a scale is hidden, it is filtered out when creating the nested arrays for `getSources`. However, the base resolution is still set on the first scale, regardless if it's hidden. It's similar to Chris's implementation of 'placeholder' on the seung-lab fork. 

Thank you for your time and consideration!

### DIFF
--- a/src/neuroglancer/datasource/precomputed/frontend.ts
+++ b/src/neuroglancer/datasource/precomputed/frontend.ts
@@ -37,7 +37,7 @@ import {Borrowed} from 'neuroglancer/util/disposable';
 import {mat4, vec3} from 'neuroglancer/util/geom';
 import {completeHttpPath} from 'neuroglancer/util/http_path_completion';
 import {isNotFoundError, responseJson} from 'neuroglancer/util/http_request';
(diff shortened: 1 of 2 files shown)


src/neuroglancer/datasource/precomputed/frontend.ts
```diff
@@ -124,6 +125,7 @@ class ScaleInfo {
           x => parseFixedLengthArray(vec3.create(), x, verifyPositiveInt));
     }
     this.key = verifyObjectProperty(obj, 'key', verifyString);
+    this.hidden =  verifyObjectProperty(obj, 'hidden', verifyOptionalBoolean);
```
jbms, line 128: add `?? false` to the end so that `this.hidden` has type `boolean` rather than `boolean | undefined`

jbms:
This change looks good, but can you also add it to the documentation in `volume.md`?

PR #453, opened 2023-04-02
Deep Zoom support

Tevemadar:
For whatever reasons, some folks want to view Deep Zoom images (high resolution images stored in a tiled gaussian pyramid) in Neuroglancer, so the 2D counter part of `precomputed`. This data source (`deepzoom://`) enables them doing that.

https://github.com/Tevemadar/neuroglancer/blob/deepzoom-support/src/neuroglancer/datasource/deepzoom/log.md contains the story, the most important points are related to `backend.ts`:

* PNG tile support is stable for sure (I'm very familiar with the internals of PNG)
* JPEG tile support is somewhat hairy (I'm not very familiar with the internals of JPEG)
* there's an alternative implementation commented out, which is stable and actually does work on every current browser, but doesn't work on Safari from more than a week ago or Firefox from more than half year ago.

I have no idea of the IPR stuff related to Deep Zoom image format. There are many software supporting it (OpenSeadragon being the most widely used, but there's OpenZoom from the past, etc.), but I couldn't locate anything related to the format.


src/neuroglancer/datasource/deepzoom/README.md
```diff
@@ -0,0 +1,11 @@
+The `"deepzoom"` data source format is based on static collections of files served directly over
+HTTP; it therefore can be used without any special serving infrastructure.  In particular, it can be
+used with data hosted by a cloud storage provider like Google Cloud Storage or Amazon S3.  Note that
+it is necessary, however, to either host the Neuroglancer client from the same server or enable CORS
+access to the data.
```
jbms, line 5: Is there a linkj to a description of the deepzoom metadata format that could be added?

config/bundle-config.js
```diff
@@ -34,6 +34,15 @@ const DEFAULT_DATA_SOURCES = exports.DEFAULT_DATA_SOURCES = [
       'neuroglancer/async_computation/decode_gzip',
     ],
   },
+  {
+    source: 'neuroglancer/datasource/deepzoom',
+    asyncComputation: [
+      'neuroglancer/async_computation/decode_jpeg',
+      'neuroglancer/async_computation/decode_gzip',
+      'neuroglancer/async_computation/decode_compresso',
```
jbms, line 42: compresso can be removed, that is not relevant to deepzoom

config/bundle-config.js
```diff
@@ -34,6 +34,15 @@ const DEFAULT_DATA_SOURCES = exports.DEFAULT_DATA_SOURCES = [
       'neuroglancer/async_computation/decode_gzip',
     ],
   },
+  {
+    source: 'neuroglancer/datasource/deepzoom',
+    asyncComputation: [
+      'neuroglancer/async_computation/decode_jpeg',
+      'neuroglancer/async_computation/decode_gzip',
```
jbms, line 41: I think gzip can also be removed, shouldn't be relevant to deepzoom.

src/neuroglancer/datasource/deepzoom/backend.ts
```diff
@@ -0,0 +1,132 @@
...
+          tiledata = await requestAsyncComputation(
+              decodePng, cancellationToken, [responseBuffer],
+              new Uint8Array(responseBuffer), tilewidth, tileheight, 3, 1, false
+          );
+          tiledata = transposeArray2d(tiledata, tilewidth*tileheight, 3);
+          break;
+
+        case ImageTileEncoding.JPG:
+        case ImageTileEncoding.JPEG:
+          /* Fragile heuristic with 10-minute read into the standard */
```
jbms, line 100: We definitely don't want to have to parse the jpeg/png headers here to determine the sizes.  Instead, we should modify the `decodeJpeg` and `decodePng` async computation interfaces to return the actual dimensions, and allow it to be unspecified.  There could also be an option to skip the jpeg transpose if that is helpful.  That will presumably require some minor changes to existing callsites.

src/neuroglancer/datasource/deepzoom/base.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export enum ImageTileEncoding {
+  JPG,
```
jbms, line 18: There should be just one entry for jpeg.

src/neuroglancer/datasource/deepzoom/frontend.ts
```diff
@@ -0,0 +1,304 @@
...
+    verifyObject(obj);
+    const rank = (numChannels === 1) ? 3 : 4;
+    const resolution = this.resolution = new Float64Array(rank);
+    const voxelOffset = this.voxelOffset = new Float32Array(rank);
+    const size = this.size = new Float32Array(rank);
+    if (rank === 4) {
+      resolution[3] = 1;
+      size[3] = numChannels;
+    }
+    verifyObjectProperty(
```
jbms, line 55: This json parsing should be removed.

src/neuroglancer/datasource/deepzoom/frontend.ts
```diff
@@ -0,0 +1,304 @@
...
+  } while (w > 1 || h > 1);
+
+  if (scaleInfos.length === 0) throw new Error('Expected at least one scale');
+  const baseScale = scaleInfos[0];
+  const rank = 4;
+  const scales = new Float64Array(rank);
+  const lowerBounds = new Float64Array(rank);
+  const upperBounds = new Float64Array(rank);
+  const names = ['x', 'y', 'z'];
+  const units = ['m', 'm', 'm'];
```
jbms, line 118: The unit can be the empty string `''` if the unit is unknown.

src/neuroglancer/datasource/deepzoom/frontend.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  for (let i = 0; i < 3; ++i) {
+    scales[i] = baseScale.resolution[i] / 1e9;
+    lowerBounds[i] = baseScale.voxelOffset[i];
+    upperBounds[i] = lowerBounds[i] + baseScale.size[i];
+  }
+  if (rank === 4) {
```
jbms, line 125: This extra logic is not applicable here, right?

src/neuroglancer/datasource/deepzoom/frontend.ts
```diff
@@ -0,0 +1,304 @@
...
+    h = Math.ceil(h / 2);
+  } while (w > 1 || h > 1);
+
+  if (scaleInfos.length === 0) throw new Error('Expected at least one scale');
+  const baseScale = scaleInfos[0];
+  const rank = 4;
+  const scales = new Float64Array(rank);
+  const lowerBounds = new Float64Array(rank);
+  const upperBounds = new Float64Array(rank);
+  const names = ['x', 'y', 'z'];
```
jbms, line 117: Should just be `x` and `y` --- the `z` dimension should be removed, rank should be 2.

src/neuroglancer/datasource/deepzoom/frontend.ts
```diff
@@ -0,0 +1,304 @@
...
+  tilesize: number;
+  overlap: number;
+  format: string;
+}
+
+function getDZIMetadata(
+    chunkManager: ChunkManager, credentialsProvider: SpecialProtocolCredentialsProvider,
+    url: string): Promise<DZIMetaData> {
+  if (url.endsWith('.json') || url.includes('.json?'))
+    throw new Error('DZI-JSON: OpenSeadragon hack not supported yet.');
```
jbms, line 230: add curly braces, also please add link to format specification for the dzi-json format

src/neuroglancer/datasource/deepzoom/log.md
```diff
@@ -0,0 +1,73 @@
...
+
+`frontend.ts`:
+
+* I think it's fully gulp-accepted now (moved a single comment around after the change)
+* 3 utility functions are removed and `imported` from `precomupted` instead. One of them was not `export`ed, now it is. These may be good candidates to be collected at some completely different location anyway, with other utility functions.
+
+Random things:
+
+* Noticed the latest feature addition, the "hidden placeholder resolution" thing. It's not in use by this datasource yet, but definitely something to look into
+* While I needed the name distinctions to have a better track of what's local and what's imported, on a longer run it may be possible the inherit from `precomputed` (instead of what's happened here, so making a copy of it)
```
jbms, line 71: Ultimately there should be very little in common, so better to have it independent as you've done.

src/neuroglancer/datasource/deepzoom/log.md
```diff
@@ -0,0 +1,73 @@
...
+
+This one works on older Firefox and Safari versions.  
+
+The built-in `png` and `jpeg` decoders are brought into action. As there are known Deep Zoom tile generators putting an extra `overlay` pixel row on right/bottom edge tiles, the specify-image-dimensions-beforehand requirement is solved with actually reading into the files (`PNG` has its dimensions at fixed positions, and in case of `JPEG` the first occurrence of a `SIFn` chunk is assumed to tell the full dimensions).  
+Extra quirk is that the `JPEG` decoder produces planar image (there is an explicit `transposeArray2d()` call for that in `decode_jpeg.ts`), while `PNG` returns packed (like `getImageData()` in the previous variant).
+
+5. Formatting, and an `export`
+
+Gulp is somewhat happy now.  
+Side note: it reformatted 178 files, 2 of them being mine (`base.ts` and `register_defaults.ts` were untouched).
```
jbms, line 55: Yeah, it has diverged a lot --- for now just clang-format the modified files and leave the rest alone.

jbms:
Thanks --- this will be a nice addition!

src/neuroglancer/datasource/deepzoom/README.md
Tevemadar: Well, this is what Microsoft has from 2011: https://learn.microsoft.com/en-us/previous-versions/windows/silverlight/dotnet-windows-silverlight/cc645077(v=vs.95) - there's an XML Schema to download, and a page describing its elements, rather terrible for a documentation.

Personally I learned the format from https://www.openzoom.org/ ("Behind the scenes" at the bottom of the page), the [second part](https://www.gasi.ch/blog/inside-deep-zoom-2) in particular.

config/bundle-config.js
Tevemadar: Yes. I haven't re-visited this file besides the intial addition and commenting-uncommenting stuff en masse.

config/bundle-config.js
Tevemadar: Same. I will apply these changes.

src/neuroglancer/datasource/deepzoom/base.ts
Tevemadar: Right now both are needed. Inside the XML descriptor a text field comes in specifying the actual extension of files. Later in `backend.js` this extension is reconstructed. `.jpg` and `.jpeg` are both valid extensions.

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: I think I've done that at a point and it triggered something about the 1e9 factor, so target dimensions appeared as `<width>`, `<height>`, `1000000000` (or `1e9`). For this one I will check the new hidden dimension feature anyway.
Side note: Deep Zoom images have no physical dimensions specified inside them. I just left the `m`s in place. Tried with `px`, but the target units still remained `m`, and upon scaling it introduced some fancy units, like nanopixels (`npx`).

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: That's a hack by OpenSeadragon guys, http://openseadragon.github.io/examples/tilesource-dzi/ - they just rewrote the XML as JSON, even keeping the `xmlns` ("XML namespace") part. While it's not standard, it's trivial to support, just I don't have ready-made test data for it (theirs is not CORS enabled).
I think for now I'll add the curly braces only.

src/neuroglancer/datasource/deepzoom/README.md
Tevemadar: Readme is extended now.

config/bundle-config.js
Tevemadar: Done.

config/bundle-config.js
Tevemadar: Done.

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: Done (-ish). So it's not implemented, but link is provided with a comment, and curly braces are there.

src/neuroglancer/datasource/deepzoom/backend.ts
Tevemadar: Related: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/@@species says `TypedArray`s can be subclassed, so it may be viable to stick `width`/`height` on `Uint8Array`, and existing code will ignore/slice. Assuming this really survives a `postMessage()`.

jbms: unfortunately `postMessage` will not transfer extra properties like that on `TypedArray`

src/neuroglancer/datasource/deepzoom/base.ts
jbms: Oh okay.

src/neuroglancer/datasource/deepzoom/frontend.ts
jbms: I think the issue you observed was simply due to using an old neuroglancer json state with the updated code, e.g. due to using `dev-server`.  The global coordinate space units and scale are part of the json state.

jbms: Note: The "hidden" scale property added recently for precomputed is something entirely different, not related to this.  That simply allows some of the scale levels that are specified in the precomputed metadata to be ignored.

src/neuroglancer/datasource/deepzoom/backend.ts
Tevemadar: I can add the other set of changes, just there may be a better idea for naming what's `BasicBitmap` above, and its `uint8Array` field (`width` and `height` could stay, I guess).

src/neuroglancer/datasource/deepzoom/frontend.ts
```diff
@@ -0,0 +1,310 @@
...
+  const dataType = DataType.UINT8;
+  const numChannels = 3;
+  const volumeType = VolumeType.IMAGE;
+  const scaleInfos = new Array<ScaleInfo>();
+  let w = width, h = height;
+  let maxlevel = Math.ceil(Math.log2(Math.max(w, h)));
+  do {
+    const lvl = scaleInfos.length;
+    const res = 1 << lvl;
+    scaleInfos.push(new ScaleInfo(
```
jbms, line 98: Instead of converting to json here and then parsing it in the `ScaleInfo` constructor, just make `ScaleInfo` an interface.

Tevemadar: Thanks for the suggestion, I will do that. Just haven't got to these parts yet.

Tevemadar: Actually I burnt myself with this multiple times last week, but now I succeeded with it in much smaller steps.
Done in https://github.com/google/neuroglancer/commit/ceca06e04a75b5eb39eb435dbd3058decefb5d85

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: Done in https://github.com/google/neuroglancer/commit/ceca06e04a75b5eb39eb435dbd3058decefb5d85

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: Done in https://github.com/google/neuroglancer/commit/ceca06e04a75b5eb39eb435dbd3058decefb5d85

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: Yes, this works after all, but has extreme side effects on the other views (makes the plane infinitely thick).

As of https://github.com/google/neuroglancer/commit/ceca06e04a75b5eb39eb435dbd3058decefb5d85 the block starts from line 64, and

    const rank = 4;
    const scales = Float64Array.of(1 / 1e9, 1 / 1e9, 1, 1);
    const lowerBounds = new Float64Array(rank);
    const upperBounds = Float64Array.of(width, height, 1, 3);
    const names = ['x', 'y', 'z', 'c^'];
    const units = ['m', 'm', '', ''];

would be the "strange" variant of it I'm talking about.

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: Apparently I managed to try this too, but it has the same z-extent-becomes-infinite side effect, and the bottom-right view becomes a constant shade of gray (also in the 3D view).

src/neuroglancer/datasource/deepzoom/frontend.ts
jbms: Yes, by default a 2-d volume will be "infinitely" thick in the third dimension.  But why do you care about other views?  Surely for a 2d dataset you would just want the xy view?  You can through the layer's coordinate transform project the 2d dataset to a 2d linear subspace of a higher-dimensional space.  That is done in the UI via the "+V" button at the bottom of the coordinate transform matrix.

PolarBean: Perhaps this could be an issue if overlaying a 2d image onto a 3-D volume.

jbms: Yes, it depends on what you want to do --- the default behavior is to "broadcast" the volume along any other dimensions, i.e. as if it is one infinitely thick voxel in that dimension, which is often the desired result when mixing 2d/3d/4d datasets.  But as noted, you can also project to a 2d subspace of a 3d coordinate space, so that it will display as a single "slice" in the 3d volume.

src/neuroglancer/async_computation/decode_jpeg.ts
```diff
@@ -22,21 +22,26 @@ import {transposeArray2d} from 'neuroglancer/util/array';
...
     decodeJpeg,
     async function(
-        data: Uint8Array, width: number, height: number, numComponents: number,
+        data: Uint8Array, width: number|undefined, height: number|undefined, numComponents: number|undefined,
         convertToGrayscale: boolean) {
       let parser = new JpegDecoder();
       parser.parse(data);
       // Just check that the total number pixels matches the expected value.
-      if (parser.width * parser.height !== width * height) {
+      if (width && height && parser.width * parser.height !== width * height) {
```
jbms, line 30: While it is unlikely to be a problem in this case, it would be better to explicitly check `width !== undefined && height !== undefined` --- just relying on the "truthy" conversion is error prone, since for example if the user specified a width of `0` that would not get checked, when instead it should lead to an error.

src/neuroglancer/async_computation/decode_jpeg.ts
```diff
@@ -22,21 +22,26 @@ import {transposeArray2d} from 'neuroglancer/util/array';
...
         throw new Error(
             `JPEG data does not have the expected dimensions: ` +
             `width=${parser.width}, height=${parser.height}, ` +
             `expected width=${width}, expected height=${height}`);
+      } else {
+        width = parser.width;
+        height = parser.height;
       }
-      if (parser.numComponents !== numComponents) {
+      if (numComponents && parser.numComponents !== numComponents) {
```
jbms, line 39: Same as above, explicitly check `numComponents !== undefined`

src/neuroglancer/datasource/deepzoom/backend.ts
```diff
@@ -0,0 +1,114 @@
...
+ * - Safari before 16.4 (OffscreenCanvas, 2023-03-27)
+ */
+// declare var OffscreenCanvas: any; // shutting up some outdated compiler(?)
+
+@registerSharedObject() export class DeepzoomImageTileSource extends
+(WithParameters(WithSharedCredentialsProviderCounterpart<SpecialProtocolCredentials>()(VolumeChunkSource), ImageTileSourceParameters)) {
+  gridShape = (() => {
+    const gridShape = new Uint32Array(3);
+    const {upperVoxelBound, chunkDataSize} = this.spec;
+    for (let i = 0; i < 3; ++i) {
```
jbms, line 41: Should go to `2` rather than `3`.

src/neuroglancer/datasource/deepzoom/frontend.ts
```diff
@@ -0,0 +1,240 @@
...
+     * JSON variant is a bit of a hack, it's not known how much it is in use for real.
+     * The actual reason for not implementing it right now is the lack of CORS-enabled
+     * test data.
+     */
+    throw new Error('DZI-JSON: OpenSeadragon hack not supported yet.');
+  }
+  return chunkManager.memoize.getUncounted(
+      {'type': 'deepzoom:metadata', url, credentialsProvider: getObjectId(credentialsProvider)},
+      async () => {
+        return await cancellableFetchSpecialOk(credentialsProvider, url, {}, responseText)
```
jbms, line 171: You can simplify this to use javascript async functions rather than `Promise.then`.

```
const text = await cancellableFetchSpecialOk(credentialsProvider, url, {}, responseText);
const xml = ...
```

src/neuroglancer/datasource/deepzoom/log.md
```diff
@@ -0,0 +1,98 @@
+# Deep Zoom support (2D pyramidal image, don't ask why people want it in Neuroglancer)
```
jbms, line 1: These comments are fine to include in the PR description but shouldn't be included as a file.

src/neuroglancer/sliceview/png/index.ts
```diff
@@ -160,9 +161,9 @@ export async function decompressPng(
   }
 
   if (
-    sx !== width 
-    || sy !== height 
-    || numComponents !== numChannels
+    width && sx !== width
```
jbms, line 164: As elsewhere, explicitly check for `!== undefined`

src/neuroglancer/datasource/deepzoom/frontend.ts
jbms: Can you say a bit more what you find disturbing?  The precomputed volume format is limited to 3d multichannel, just as deepzoom is limited to 2d multichannel.  But e.g. the zarr and n5 data sources allow an arbitrary number of dimensions.

jbms:
Can you include this data source in the top-level README, and if there is a public example dataset you can link to, perhaps include that in the examples as well?

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: Yes, of course, that's just stuck this way after the initial copy-paste. `return await` without a `try-catch` around is suspicious anyway, there's another one in the same file.
However I'm leaving for a week right now, so the changes will have to wait.

src/neuroglancer/datasource/deepzoom/backend.ts
Tevemadar: Done in https://github.com/google/neuroglancer/pull/453/commits/a5737feb51489a7b612d8bee5a376612ef25b608
Also reduced the size of the array itself, it still seems to work.

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: Done in https://github.com/google/neuroglancer/pull/453/commits/a5737feb51489a7b612d8bee5a376612ef25b608

src/neuroglancer/sliceview/png/index.ts
Tevemadar: Done in https://github.com/google/neuroglancer/pull/453/commits/a5737feb51489a7b612d8bee5a376612ef25b608

src/neuroglancer/datasource/deepzoom/frontend.ts
Tevemadar: > Can you say a bit more what you find disturbing?
1. I made changes which were systematic, but I don't actually know what they interface with
2. *and* the `xz` and `yz` views started to behave differently. I understand they don't matter, but I don't understand why they became different. I'd expect `yz` to produce the same striped pattern, just perhaps horizontally (though in fact that would be `zy`).

And it does, after adding a fake "axis": [strange](https://www.nesys.uio.no/NG+DZI/#!%7B%22dimensions%22:%7B%22x%22:%5B1e-9%2C%22m%22%5D%2C%22y%22:%5B1e-9%2C%22m%22%5D%7D%2C%22position%22:%5B26327.98828125%2C18448.6015625%5D%2C%22crossSectionScale%22:141.17496392147683%2C%22projectionScale%22:65536%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22deepzoom://https://data-proxy.ebrains.eu/api/v1/buckets/localizoom/14122_mPPC_BDA_s186.tif/14122_mPPC_BDA_s186.dzi%22%2C%22tab%22:%22source%22%2C%22shader%22:%22void%20main%20%28%29%20%7B%5Cn%20%20emitRGB%28vec3%28toNormalized%28getDataValue%280%29%29%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20toNormalized%28getDataValue%281%29%29%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20toNormalized%28getDataValue%282%29%29%29%29%3B%5Cn%7D%22%2C%22channelDimensions%22:%7B%22c%5E%22:%5B1%2C%22%22%5D%7D%2C%22name%22:%2214122_mPPC_BDA_s186.dzi%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%2214122_mPPC_BDA_s186.dzi%22%7D%2C%22layout%22:%224panel%22%7D) vs. [expected](https://www.nesys.uio.no/NG+DZI/#!%7B%22dimensions%22:%7B%22x%22:%5B1e-9%2C%22m%22%5D%2C%22y%22:%5B1e-9%2C%22m%22%5D%2C%22axis%22:%5B1%2C%22%22%5D%7D%2C%22position%22:%5B26327.98828125%2C18448.6015625%2C0.5%5D%2C%22crossSectionScale%22:141.17496392147683%2C%22projectionScale%22:65536%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22deepzoom://https://data-proxy.ebrains.eu/api/v1/buckets/localizoom/14122_mPPC_BDA_s186.tif/14122_mPPC_BDA_s186.dzi%22%2C%22transform%22:%7B%22sourceRank%22:3%2C%22outputDimensions%22:%7B%22x%22:%5B1e-9%2C%22m%22%5D%2C%22y%22:%5B1e-9%2C%22m%22%5D%2C%22c%5E%22:%5B1%2C%22%22%5D%2C%22axis%22:%5B1%2C%22%22%5D%7D%2C%22inputDimensions%22:%7B%22x%22:%5B1e-9%2C%22m%22%5D%2C%22y%22:%5B1e-9%2C%22m%22%5D%2C%22c%5E%22:%5B1%2C%22%22%5D%2C%22%22:%5B1%2C%22%22%5D%7D%7D%7D%2C%22tab%22:%22source%22%2C%22shader%22:%22void%20main%20%28%29%20%7B%5Cn%20%20emitRGB%28vec3%28toNormalized%28getDataValue%280%29%29%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20toNormalized%28getDataValue%281%29%29%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20toNormalized%28getDataValue%282%29%29%29%29%3B%5Cn%7D%22%2C%22channelDimensions%22:%7B%22c%5E%22:%5B1%2C%22%22%5D%7D%2C%22name%22:%2214122_mPPC_BDA_s186.dzi%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%2214122_mPPC_BDA_s186.dzi%22%7D%2C%22layout%22:%224panel%22%7D)

src/neuroglancer/datasource/deepzoom/log.md
Tevemadar: `log.md` is removed now.
Added `deepzoom` in the datasources section, and linked an example image. Of course it doesn't bring anywhere now, but it's the same view as [this](https://www.nesys.uio.no/NG+DZI/#!%7B%22dimensions%22:%7B%22x%22:%5B1e-9%2C%22m%22%5D%2C%22y%22:%5B1e-9%2C%22m%22%5D%7D%2C%22position%22:%5B10387071%2C5347131%5D%2C%22crossSectionScale%22:263.74955563693914%2C%22projectionScale%22:65536%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22deepzoom://https://data-proxy.ebrains.eu/api/v1/buckets/localizoom/14122_mPPC_BDA_s186.tif/14122_mPPC_BDA_s186.dzi%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B1e-9%2C%22m%22%5D%2C%22y%22:%5B1e-9%2C%22m%22%5D%2C%22c%5E%22:%5B1%2C%22%22%5D%7D%2C%22inputDimensions%22:%7B%22x%22:%5B3.25e-7%2C%22m%22%5D%2C%22y%22:%5B3.25e-7%2C%22m%22%5D%2C%22c%5E%22:%5B1%2C%22%22%5D%7D%7D%7D%2C%22tab%22:%22rendering%22%2C%22shader%22:%22void%20main%28%29%7BemitRGB%28vec3%28toNormalized%28getDataValue%280%29%29%2CtoNormalized%28getDataValue%281%29%29%2CtoNormalized%28getDataValue%282%29%29%29%29%3B%7D%22%2C%22channelDimensions%22:%7B%22c%5E%22:%5B1%2C%22%22%5D%7D%2C%22name%22:%2214122_mPPC_BDA_s186.dzi%22%7D%5D%2C%22selectedLayer%22:%7B%22layer%22:%2214122_mPPC_BDA_s186.dzi%22%7D%2C%22layout%22:%22xy%22%7D). Even the scalebar is real now.

Side note: I used `<kbd>` for mentioning <kbd>Ctrl</kbd>+<kbd>MouseWheel</kbd> in particular, but later I saw that keyboard commands are simple backticks at other places. It seems to [look okay](https://github.com/Tevemadar/neuroglancer/blob/deepzoom-support/README.md#examples) in Chrome/Edge/FireFox, but I admit I haven't tried in Safari or mobile browsers.

PR #460, opened 2023-05-03
feat(graphene): multiscale meshes

akhileshh:


.github/workflows/build.yml
```diff
@@ -3,272 +3,67 @@ name: Build
 on: [push, pull_request]
 
 jobs:
-  client:
```
jbms, line 6: I don't think you intended to include the changes in this file.

appengine/frontend/app.yaml
```diff
@@ -0,0 +1,32 @@
+runtime: python27
```
jbms, line 1: Probably not intended to be included in this PR.

config/state_servers.json
```diff
@@ -0,0 +1,6 @@
+{
```
jbms, line 1: Not intended for this PR.

src/neuroglancer/mesh/draco/index.ts
```diff
@@ -68,7 +68,7 @@ export async function decodeDracoPartitioned(
   heap.set(buffer, offset);
   numPartitions = partition ? 8 : 1;
   const code = (m.instance.exports.neuroglancer_draco_decode as Function)(
-      offset, buffer.byteLength, partition, vertexQuantizationBits, true);
+      offset, buffer.byteLength, partition, vertexQuantizationBits, false);
```
jbms, line 71: This will need to be an option, since we don't want to change this option for the neuroglancer precomputed datasource.

src/neuroglancer/datasource/graphene/backend.ts
```diff
@@ -44,14 +44,15 @@ import { SharedWatchableValue } from 'neuroglancer/shared_watchable_value';
 import { DisplayDimensionRenderInfo } from 'neuroglancer/navigation_state';
 import { forEachVisibleSegment } from 'neuroglancer/segmentation_display_state/base';
 import { computeChunkBounds } from 'neuroglancer/sliceview/volume/backend';
+import { verifyObject } from 'src/neuroglancer/util/json';
```
jbms, line 47: Should be 'neuroglancer/util/json' -- no 'src/' prefix.

src/neuroglancer/mesh/draco/index.ts
akhileshh: Yes I will do that when the PR is ready to merge. This was just to get the meshes to appear during development.

src/neuroglancer/datasource/graphene/backend.ts
```diff
@@ -127,6 +128,82 @@ async function decodeDracoFragmentChunk(
...
+
+function decodeMultiscaleManifestChunk(chunk: GrapheneMultiscaleManifestChunk, response: any) {
+  verifyObject(response);
+  chunk.manifest = {
+    chunkShape: vec3.clone(response.chunkShape),
+    chunkGridSpatialOrigin: vec3.create(),
+    lodScales: new Float32Array(response.lodScales),
+    octree: new Uint32Array(response.octree),
+    vertexOffsets: new Float32Array(response.lodScales.length * 3),
+    clipLowerBound: vec3.create(),
```
jbms, line 149: This is why the meshes aren't displaying properly --- the mesh rendering code calculates bounds for each octree node by calculating the bounds of the node according to the grid, and then restricting them by these overall clip bounds.  If the resultant bounding box does not intersect the view frustum, then the octree node won't be rendered.  With these values for the clip bounds, all of the calculated bounds end up being (0, 0, 0)---(0,0,0), which both defeats the normal logic of selectively rendering only visible chunks, and also means if the view frustum does not contain (0, 0, 0) then nothing will be rendered.

jbms: If you don't have the real bounds available, but you can just calculate the bounds that include all of the fragments present.

akhileshh: Thanks! I have pushed changes and updated the chunkedgraph deployment to return clip bounds.

Realized that zoom-out disappearance is being caused by resolution, changing it fixes that. But this means higher resolution meshes are loaded earlier than expected.

Currently chunkedgraph is passing bounding box of the dataset for clip lower and upper bounds. Should that fix the above issue? It doesn't look like it does though, I'm not familiar with webGL.

PR #464, opened 2023-05-12
feat(tools): custom keybinds and layer independent tool keybinds

chrisj:
----
🔥 Preview 🔥: https://neuroglancer--pr464-jqtohdii.web.app
Updated at Thu Feb 22 2024 17:27:49 GMT+0000 (Coordinated Universal Time) for commit bc4f166a9f852f529781d72ce2925afd2da6a7dd
Expires: Invalid Date

### DIFF
--- a/src/ui/default_viewer_setup.ts
+++ b/src/ui/default_viewer_setup.ts
@@ -56,7 +56,7 @@ type CustomBinding = {
 };
 
 type CustomBindings = {
-  [identifier: EventIdentifier]: CustomBinding;
+  [identifier: EventIdentifier]: CustomBinding | CustomBinding[];
 };
 
 declare const NEUROGLANCER_CUSTOM_INPUT_BINDINGS: CustomBindings | undefined;
@@ -128,20 +128,22 @@ function setCustomInputEventBindings(viewer: Viewer, bindings: CustomBindings) {
   };
 
   for (const [key, val] of Object.entries(bindings)) {
-    const { action, context = "global" } = val;
-    const actionMap = viewer.inputEventBindings[context];
-    if (actionMap === undefined) {
-      throw new Error(`invalid action map context: ${context}`);
-    }
-    if (typeof action === "boolean") {
-      if (action === false) {
-        deleteKey(actionMap, key);
+    const bindings = Array.isArray(val) ? val : [val];
+    for (const { action, context = "global" } of bindings) {
+      const actionMap = viewer.inputEventBindings[context];
+      if (actionMap === undefined) {
+        throw new Error(`invalid action map context: ${context}`);
       }
-    } else if (typeof action === "string" || "action" in action) {
-      actionMap.set(key, action);
-    } else {
-      const toolAction = bindNonLayerSpecificTool(key, action);
-      actionMap.set(key, toolAction);
+      if (typeof action === "boolean") {
+        if (action === false) {
+          deleteKey(actionMap, key);
+        }
+      } else if (typeof action === "string" || "action" in action) {
+        actionMap.set(key, action);
+      } else {
+        const toolAction = bindNonLayerSpecificTool(key, action);
+        actionMap.set(key, toolAction);
+      }
     }
   }
 }
--- a/src/ui/default_viewer_setup.ts
+++ b/src/ui/default_viewer_setup.ts
@@ -32,7 +32,6 @@ import { UrlHashBinding } from "#src/ui/url_hash_binding.js";
 import type {
   ActionIdentifier,
   EventAction,
-  EventActionMap,
   EventIdentifier,
 } from "#src/util/event_action_map.js";
 import {
@@ -65,7 +64,10 @@ export const hasCustomBindings =
   typeof NEUROGLANCER_CUSTOM_INPUT_BINDINGS !== "undefined" &&
   Object.keys(NEUROGLANCER_CUSTOM_INPUT_BINDINGS).length > 0;
 
-function setCustomInputEventBindings(viewer: Viewer, bindings: CustomBindings) {
+export function setCustomInputEventBindings(
+  viewer: Viewer,
+  bindings: CustomBindings,
+) {
   const bindNonLayerSpecificTool = (
     key: string,
     customBinding: CustomToolBinding,
@@ -120,13 +122,6 @@ function setCustomInputEventBindings(viewer: Viewer, bindings: CustomBindings) {
     return action;
   };
 
-  const deleteKey = (map: EventActionMap, key: string) => {
-    map.delete(key);
-    for (const pMap of map.parents) {
-      deleteKey(pMap, key);
-    }
-  };
-
   for (const [key, val] of Object.entries(bindings)) {
     const bindings = Array.isArray(val) ? val : [val];
     for (const { action, context = "global" } of bindings) {
@@ -136,7 +131,7 @@ function setCustomInputEventBindings(viewer: Viewer, bindings: CustomBindings) {
       }
       if (typeof action === "boolean") {
         if (action === false) {
-          deleteKey(actionMap, key);
+          actionMap.suppress(key);
         }
       } else if (typeof action === "string" || "action" in action) {
         actionMap.set(key, action);
--- a/src/ui/default_viewer_setup.ts
+++ b/src/ui/default_viewer_setup.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import type { UserLayer, UserLayerConstructor } from "#src/layer/index.js";
+import { layerTypes } from "#src/layer/index.js";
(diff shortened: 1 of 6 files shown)


src/neuroglancer/ui/default_viewer_setup.ts
```diff
@@ -20,16 +20,84 @@ import {setDefaultInputEventBindings} from 'neuroglancer/ui/default_input_event_
...
 import {bindTitle} from 'neuroglancer/ui/title';
 import {UrlHashBinding} from 'neuroglancer/ui/url_hash_binding';
+import {UserLayer, UserLayerConstructor} from 'neuroglancer/layer';
+import {SegmentationUserLayer} from 'neuroglancer/segmentation_user_layer';
+import {Tool, restoreTool} from 'neuroglancer/ui/tool';
 
 declare var NEUROGLANCER_DEFAULT_STATE_FRAGMENT: string|undefined;
 
+type CustomBinding = {
+  layer: string, tool: string, protocol?: string,
```
jbms, line 30: `tool` should be a json object, not just a string, as some tools support options

src/neuroglancer/ui/default_viewer_setup.ts
```diff
@@ -20,16 +20,84 @@ import {setDefaultInputEventBindings} from 'neuroglancer/ui/default_input_event_
...
+          }
+          if (previousTool) {
+            viewer.activateTool(toolKey, previousTool);
+          }
+        }
+      }
+    });
+  }
+
+  const nameToLayer: {[key: string]: UserLayerConstructor|undefined} = {};
```
jbms, line 80: This already exists as `layerTypes` in `neuroglancer/layer.ts`

jakobtroidl:
@jbms, is it possible to get this merged? We need this feature in ng for proofreading H01 with CAVE.

chrisj:
Updated this pull request with the ability to unbind existing keybinds with a false value

Example config

```json
{
  "keym": {
    "layer": "segmentation",
    "tool": "grapheneMergeSegments",
    "provider": "graphene"
  },
  "keyc": {
    "layer": "segmentation",
    "tool": "grapheneMulticutSegments",
    "provider": "graphene"
  },
  "keyf": {
    "layer": "segmentation",
    "tool": "grapheneFindPath",
    "provider": "graphene"
  },
  "keyx": false
}
```

src/ui/default_viewer_setup.ts
```diff
@@ -14,26 +14,124 @@
...
+  layer: string;
+  tool: unknown;
+  provider?: string;
+};
+
+type CustomBindings = {
+  [key: string]: CustomToolBinding | string | boolean;
+};
+
+declare const CUSTOM_BINDINGS: CustomBindings | undefined;
```
jbms, line 48: Please rename to NEUROGLANCER_CUSTOM_INPUT_BINDINGS

src/ui/default_viewer_setup.ts
```diff
@@ -14,26 +14,124 @@
...
+          }
+        }
+      }
+    });
+  };
+
+  if (hasCustomBindings) {
+    for (const [key, val] of Object.entries(CUSTOM_BINDINGS!)) {
+      if (typeof val === "string") {
+        viewer.inputEventBindings.global.set(key, val);
```
jbms, line 111: Note that in addition to the global map there are several others that are relevant in default_input_event_bindings.ts --- in particular the default slice view and perspective view bindings.  Perhaps this should account for that?

src/ui/default_viewer_setup.ts
```diff
@@ -14,26 +14,124 @@
...
+      const acceptableLayers = viewer.layerManager.managedLayers.filter(
+        (managedLayer) => {
+          const correctLayerType =
+            managedLayer.layer instanceof desiredLayerType;
+          if (desiredProvider && correctLayerType) {
+            for (const dataSource of managedLayer.layer?.dataSources || []) {
+              const protocol = viewer.dataSourceProvider.getProvider(
+                dataSource.spec.url,
+              )[2];
+              if (protocol === desiredProvider) {
```
jbms, line 83: With the kvstore changes getProvider is not exactly the right method anymore.  Perhaps just have a regexp that matches on the URL?

src/ui/tool.ts
```diff
@@ -318,15 +318,15 @@ export class GlobalToolBinder extends RefCounted {
     this.changed.dispatch();
   }
 
-  activate(key: string): Borrowed<Tool> | undefined {
-    const tool = this.get(key);
+  activate(key: string, tool?: Tool<object>): Borrowed<Tool> | undefined {
+    tool = tool || this.get(key);
```
jbms, line 322: Please use `tool ?? this.get(key)` instead --- not necessary here but avoids thinking about javascript truthyness complications

src/ui/tool.ts
```diff
@@ -318,15 +318,15 @@ export class GlobalToolBinder extends RefCounted {
...
     if (tool === undefined) {
       this.deactivate_();
       return;
     }
     this.debounceDeactivate.cancel();
     const activeTool = this.activeTool_;
-    if (tool === activeTool?.tool) {
+    if (tool.toJSON() === activeTool?.tool.toJSON()) {
```
jbms, line 329: We can't compare the json like this because it can be an object rather than a plain string.  Instead we need to compare the serialized json.  However, it would be better if we can avoid having to insert the json comparison here, and instead just ensure that the same `tool` object is passed.  Is your logic involving `previousTool` not sufficient?

src/ui/default_viewer_setup.ts
```diff
@@ -14,26 +14,124 @@
...
+import {
+  verifyObject,
+  verifyObjectProperty,
+  verifyString,
+} from "#src/util/json.js";
 
 declare let NEUROGLANCER_DEFAULT_STATE_FRAGMENT: string | undefined;
 
+type CustomToolBinding = {
+  layer: string;
```
jbms, line 39: rename to `layerType` for consistency with the tool palette queries.

src/ui/default_viewer_setup.ts
chrisj: I added a "context" property, one of `"global" | "perspectiveView" | "sliceView"`. I could also call it map or actionMap. Restructured the json object to support it.


```json
{
  "keym": {
    "action": {
      "layerType": "segmentation",
      "tool": "grapheneMergeSegments",
      "provider": "graphene"
    }
  },
  "keyx": {
    "action": false
  },
  "control+shift+keyx": {
    "action": "clear-segments",
    "context": "sliceView"
  }
}
```

src/ui/default_viewer_setup.ts
chrisj: I switched it to use the schemePattern from `"#src/kvstore/url.js";` does that seem fine?

src/ui/tool.ts
chrisj: I think originally my previousTool was for deactivation but then I realized if you have a custom binding and an user specified binding for the same tool+layer. It is a bit awkward. It is not a big deal because very few users would swap between both so this code can be removed but I made it so they can deactivate each other. For example if `c` is bound to multicut tool and the user binds shift+a to multicut, they can activate it with either bind and deactivate it with the other.

```
activeTool.tool.constructor === tool.constructor &&
activeTool.tool.context === tool.context
```

I think this logic is cleaner, let me know. If you prefer, I'm fine with removing it. Also because of this logic, previousTool has less value, really just a cache for `restoreTool`.   I could remove it and call `restoreTool` with every keypress.

chrisj:
Example config:
```json
{
  "keyh": {
    "action": "help"
  },
  "backslash": {
    "context": "global",
    "action": false
  },
  "at:control+wheel": {
    "context": "sliceView",
    "action": {
      "action": "zoom-via-wheel",
      "preventDefault": true
    }
  },
  "at:shift+mousedown0": [
    {
      "context": "sliceView",
      "action": {
        "action": "rotate-via-mouse-drag",
        "stopPropagation": true
      }
    },
    {
      "context": "perspectiveView",
      "action": {
        "action": "translate-via-mouse-drag",
        "stopPropagation": true
      }
    }
  ],
  "keyq": {
    "action": {
      "layerType": "segmentation",
      "tool": "selectSegments"
    }
  },
  "keyw": {
    "action": {
      "layerType": "segmentation",
      "provider": "precomputed",
      "tool": {
        "type": "selectSegments"
      }
    }
  }
}
```

chrisj:
I added documentation here: https://neuroglancer-docs--pr464-wl93s40n.web.app/json/api/index.html#configuration

I'm not sure if these json config files should go in the JSON API tab. I find it is a little bit harder to follow than just looking at an example as above but it is important to understand the event_action_map and the EventIdentifier schema.

I fixed a few edge cases and locked down tool keybinds to single letter for now as that is what ToolBinder will currently support.

PR #465, opened 2023-05-17
feat(datasource/graphene) auto reloading graphene manifest after an edit

chrisj:
Replaces https://github.com/google/neuroglancer/pull/450

With our current infrastructure, the only way for us to get mesh updates is to continuously poll the manifest endpoint to see if it changes, we can't rely on 404 responses as previously thought. Eventually we want to get push messages on the client to replace this logic.

I was asked to try to reduce the flicker that occurs when we change the manifest of an existing cell. I have partial "fix" with this commit 80d95e103c3a03bc04e338371455939dfb233cab. The issue is that the ManifestChunk arrives before the first FragmentChunk.

The easiest solution that mostly works is to prevent ManifestChunks from triggering redraws. I was surprised that ManfestChunk gets a GPU_MEMORY state. It still can cause issues.

Do you have any suggestions on how we could make it smoother. I was asked to go even further by keeping existing fragments from previous cells around until the new cell is at least partially meshed. For example, with a merge, it is very easy to just combine the fragments from each segment.

### DIFF
--- a/src/neuroglancer/chunk_manager/frontend.ts
+++ b/src/neuroglancer/chunk_manager/frontend.ts
@@ -224,9 +224,10 @@ export class ChunkQueueManager extends SharedObject {
         if (newState !== oldState) {
           switch (newState) {
             case ChunkState.GPU_MEMORY:
-              // console.log("Copying to GPU", chunk);
               chunk.copyToGPU(this.gl);
-              visibleChunksChanged = true;
+              if (chunk.constructor.name !== "ManifestChunk") {
+                visibleChunksChanged = true;
+              }
               break;
             case ChunkState.SYSTEM_MEMORY:
               chunk.freeGPUMemory(this.gl);
--- a/src/neuroglancer/chunk_manager/frontend.ts
+++ b/src/neuroglancer/chunk_manager/frontend.ts
@@ -224,10 +224,9 @@ export class ChunkQueueManager extends SharedObject {
         if (newState !== oldState) {
           switch (newState) {
             case ChunkState.GPU_MEMORY:
+              // console.log("Copying to GPU", chunk);
               chunk.copyToGPU(this.gl);
-              if (chunk.constructor.name !== "ManifestChunk") {
-                visibleChunksChanged = true;
-              }
+              visibleChunksChanged = true;
               break;
             case ChunkState.SYSTEM_MEMORY:
               chunk.freeGPUMemory(this.gl);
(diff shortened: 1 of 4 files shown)


fcollman:
@jbms any reason we can't get this one merged?

src/neuroglancer/datasource/graphene/frontend.ts
```diff
@@ -949,6 +949,27 @@ class GraphConnection extends SegmentationGraphSourceConnection {
     return undefined;
   }
 
+  getMeshSource() {
```
jbms, line 952: This should somehow ensure that it is associated with this graph --- can have more than one datasource in a UserLayer.

src/neuroglancer/chunk_manager/frontend.ts
```diff
@@ -224,9 +224,10 @@ export class ChunkQueueManager extends SharedObject {
         if (newState !== oldState) {
           switch (newState) {
             case ChunkState.GPU_MEMORY:
-              // console.log("Copying to GPU", chunk);
               chunk.copyToGPU(this.gl);
-              visibleChunksChanged = true;
+              if (chunk.constructor.name !== "ManifestChunk") {
+                visibleChunksChanged = true;
```
jbms, line 229: This shouldn't be done by name, and also skipping the notification doesn't really fix the problem in general, because any other chunk update would also trigger the notification.  Additionally it will still redraw on the first fragment that is received so there will still be flicker if there is more than one fragment.  I think a different strategy is needed to avoid flicker.

Will some of the existing fragments be referenced by the new manifest, if they haven't changed, or does the fragment key depend on the root id?

jbms:
Sorry for the delay in reviewing, I added some comments.

src/neuroglancer/datasource/graphene/frontend.ts
chrisj: Would this solve that? It looks like I could either do `loadState.dataSource.subsources` or `loadState.subsources`
 
```ts
getMeshSource() {
    const {layer} = this;
    for (const dataSource of layer.dataSources) {
      const {loadState} = dataSource;
      if (loadState instanceof LoadedLayerDataSource) {
        const {subsources} = loadState.dataSource;
        const graphSubsource = subsources.filter(subsource => subsource.id === 'graph')[0];
        if (graphSubsource && graphSubsource.subsource.segmentationGraph) {
          if (graphSubsource.subsource.segmentationGraph !== this.graph) {
            continue;
          }
        }
        const meshSubsource = subsources.filter(subsource => subsource.id === 'mesh')[0];
        if (meshSubsource) {
          return meshSubsource.subsource.mesh;
        }
      }
    }
    return undefined;
  }
```

chrisj:
@jbms I added a response about `getMeshSource`

I will remove the attempt at fixing flicker until we have a good solution

chrisj:
@jbms I pushed the change I talked about and removed the flicker reduction code so it should be ready to be merged.

PR #475, opened 2023-07-28
annotation: add missing overload for setPointMarkerColor(vec3)

stuarteberg:
It's listed in [the docs](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/annotation/rendering.md#point-annotations), but was missing from the overload set.


jbms:
Thanks!

PR #477, opened 2023-08-02
N5 datasource docs JSON fix

clbarnes:
This README is the best documentation of the BDV / N5viewer metadata I keep referring to, so I thought I'd help get rid of some red squigglies.


jbms:
Thanks!

PR #493, opened 2023-09-29
Annotation features mostly related to MultiscaleAnnotationSource

chrisj:
All of these were developed for our cave annotation datasource. I could add the cave datasource here or keep it a separate PR, whichever you prefer.

Each of these features still needs discussion.

53821ff246731b2a26cea19340ec5c5c7ec17f62

Mostly happy with this but I left out sorting. We don't always want to sort annotations by distance, we often want the order from the server response. We also are only using 1 chunk the size of the dataset for annotations.

cee014674ee86dfb547b12f3e90d84f520c45229

This feature makes it easier to quickly glance at the annotations, more useful when the segment list is small. Because there can be multiple annotation lists visible at a time, I'm currently relying on tabIndex to make it focusable so I can listen to key events directly on that element. It isn't ideal but it's robust.

8398ac52e85f5f7bb36d0a7c542f55ab135589d9
Screen Shot 2023-09-28 at 6 52 12 PM

A list of enum values colored by a default hash function (exposed in the shader as `hashColor`) appears when hovering over `enum` in the `neuroglancer-annotation-shader-property-list`.

The hashColor function is really nice for quickly making a useful annotation shader. I could add a seed argument. The list element would no longer match unless we started to parse the shader for that which is probably something to avoid.

If I change the shader prop code so that it supports `prop_cell_type == prop_cell_type("b")` instead as you suggested https://github.com/google/neuroglancer/discussions/481#discussioncomment-6795461 Does glsl allow prop_cell_type to be both a value and a function or would it be a pre-compilation step?

### DIFF
--- a/src/neuroglancer/annotation/annotation_layer_state.ts
+++ b/src/neuroglancer/annotation/annotation_layer_state.ts
@@ -55,14 +55,7 @@ export class WatchableAnnotationRelationshipStates extends
         nestedContext.registerDisposer(segmentationGroupState.changed.add(this.changed.dispatch));
         nestedContext.registerDisposer(registerNested((groupContext, groupState) => {
           const {visibleSegments} = groupState;
-          let wasEmpty = visibleSegments.size === 0;
-          groupContext.registerDisposer(visibleSegments.changed.add(() => {
-            const isEmpty = visibleSegments.size === 0;
-            if (isEmpty !== wasEmpty) {
-              wasEmpty = isEmpty;
-              this.changed.dispatch();
-            }
-          }));
+          groupContext.registerDisposer(visibleSegments.changed.add(this.changed.dispatch));
         }, segmentationGroupState));
       }, segmentationState));
     });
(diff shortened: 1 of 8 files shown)


src/neuroglancer/segment_color.ts
```diff
@@ -28,6 +28,33 @@ import {Trackable} from './util/trackable';
 
 const NUM_COMPONENTS = 2;
 
+export class BasicHashColorShaderManager {
```
chrisj, line 31: I left this here for now in case we don't want to go ahead with this approach

src/neuroglancer/annotation/annotation_layer_state.ts
```diff
@@ -55,14 +55,7 @@ export class WatchableAnnotationRelationshipStates extends
...
           const {visibleSegments} = groupState;
-          let wasEmpty = visibleSegments.size === 0;
-          groupContext.registerDisposer(visibleSegments.changed.add(() => {
-            const isEmpty = visibleSegments.size === 0;
-            if (isEmpty !== wasEmpty) {
-              wasEmpty = isEmpty;
-              this.changed.dispatch();
-            }
-          }));
+          groupContext.registerDisposer(visibleSegments.changed.add(this.changed.dispatch));
```
chrisj, line 58: this was simplified because I need the annotation list to update anytime the segment list changes

chrisj:
Closed in favor of separate pull requests.

PR #494, opened 2023-10-10
feat: add a UI parameter to control the selected volume rendering shader

seankmartin:
# Goal

Enable the use of different fragment shaders in the volume rendering mode. Enable the user to change between these modes in the UI, and store the selected mode in JSON state.

## Primary changes:
1. Introduced a volume rendering shader mode enum which is controllable from UI.
2. The shader tracks this set value, and triggers a shader rebuild and a redraw then the value is changed.
3. Moved most volume rendering vertex and fragment shader code into a `volume_rendering/glsl.ts` file with variables named `glsl_NAME`. Additional shaders can be added here.
4. Enable control of this value from Python via a tool.
5. Added a normalised current chunk number uniform to the shader that is used in the chunk visualisation debug mode to visualise the current chunk.

## Uncertainties
1. Line 199 in `image_user_layer.ts`, do we need to call `verfiyOptionalProperty` to restore the state for the volume rendering mode, which is a `TrackableEnum`? I've done this because the blend mode is also a `TrackableEnum` and the blend mode state is restored using `verifyOptionalProperty`.


src/neuroglancer/image_user_layer.ts
```diff
@@ -84,6 +86,7 @@ export class ImageUserLayer extends Base {
       channelCoordinateSpace => makeValueOrError(() => getChannelSpace(channelCoordinateSpace)),
       this.channelCoordinateSpace));
   volumeRendering = new TrackableBoolean(false, false);
```
jbms, line 88: I think it would make more sense to just use the existing `volumeRendering` variable but change its type from `boolean` to enum.  There can be a NONE value or similar to indicate it is disabled.

src/neuroglancer/image_user_layer.ts
```diff
@@ -168,6 +172,7 @@ export class ImageUserLayer extends Base {
           renderScaleHistogram: this.volumeRenderingRenderScaleHistogram,
           localPosition: this.localPosition,
           channelCoordinateSpace: this.channelCoordinateSpace,
+          shaderSelection: this.volumeRenderingShaderSelection,
```
jbms, line 175: I think I would call this `mode` rather than `shaderSelection` --- while perhaps currently only one shader snippet is swapped, as the feature is refined there will surely be many more differences.

src/neuroglancer/volume_rendering/trackable_shader_mode.ts
```diff
@@ -0,0 +1,20 @@
...
+export enum SHADER_MODES {
+  DEFAULT = 0,
+  MAX_PROJECTION = 1,
+  CHUNK_VISUALIZATION = 2,
+}
+
+export const SHADER_FUNCTIONS = new Map([
+  [SHADER_MODES.DEFAULT, glsl_USER_DEFINED_RAY_TRAVERSAL],
+  [SHADER_MODES.MAX_PROJECTION, glsl_MAX_PROJECTION_SHADER],
+  [SHADER_MODES.CHUNK_VISUALIZATION, glsl_CHUNK_NUMBER_SHADER],
```
jbms, line 13: There is already a "wire frame" mode option in the general "settings" panel that is intended for the same use case of debugging rendering.  I think we could just use that option to enable this mode rather than having a separate volume rendering option.

src/neuroglancer/volume_rendering/glsl.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+    intersectStart = max(intersectStart, startFraction);
+    intersectEnd = min(intersectEnd, endFraction);
+  }
+  float stepSize = (uFarLimitFraction - uNearLimitFraction) / float(uMaxSteps - 1);
+  int startStep = int(floor((intersectStart - uNearLimitFraction) / stepSize));
+  int endStep = min(uMaxSteps, int(floor((intersectEnd - uNearLimitFraction) / stepSize)) + 1);
+`;
+
+const glsl_TRAVERSE_RAYS = `
```
jbms, line 61: I'm not sure how helpful it is to move these snippets to a separate file, since then it makes it less clear how they fit into the overall generated shader. 

As discussed in the meeting, with further refinement, max projection would likely be structured rather differently from the regular volume rendering, for example.

jbms:
Yes, every property does need to be separately handled in restoreState.

src/neuroglancer/volume_rendering/trackable_shader_mode.ts
seankmartin: I wasn't aware of the option, thanks. That sounds like a better solution. If it works, for this round, I'll convert the existing chunk visualisation mode to a wireframe of the chunks that are rendered instead of the ray casting.
In the future, we could maybe get that to be an overlay of the chunk wireframe on top of the ray casting instead.

seankmartin:
Thanks @jbms for all the comments and suggestions, appreciate that!

I've made changes that hopefully move in the right direction for addressing these. Let me know what you think. I can start moving the chunk debug over to the wireframe rendering global option as well.

seankmartin:
Ah I see, thanks. What might be the best way to go about accessing the value of the `viewer.wireFrame` boolean option from within a `VolumeRenderingRenderLayer` object?

jbms:
You can access it via `renderContext.wireFrame`, and then pass it as a parameter to `shaderGetter` --- see e.g. https://github.com/google/neuroglancer/blob/master/src/neuroglancer/sliceview/volume/renderlayer.ts which does that for cross-section views

src/neuroglancer/image_user_layer.ts
```diff
@@ -42,23 +41,23 @@ import {makeCopyButton} from 'neuroglancer/widget/copy_button';
...
 import {addLayerControlToOptionsTab, LayerControlDefinition, registerLayerControl} from 'neuroglancer/widget/layer_control';
-import {checkboxLayerControl} from 'neuroglancer/widget/layer_control_checkbox';
 import {enumLayerControl} from 'neuroglancer/widget/layer_control_enum';
 import {rangeLayerControl} from 'neuroglancer/widget/layer_control_range';
 import {makeMaximizeButton} from 'neuroglancer/widget/maximize_button';
 import {renderScaleLayerControl} from 'neuroglancer/widget/render_scale_widget';
 import {ShaderCodeWidget} from 'neuroglancer/widget/shader_code_widget';
 import {LegendShaderOptions, registerLayerShaderControlsTool, ShaderControls} from 'neuroglancer/widget/shader_controls';
 import {Tab} from 'neuroglancer/widget/tab_view';
+import {trackableShaderModeValue, VOLUME_RENDERING_MODES} from 'src/neuroglancer/volume_rendering/trackable_volume_rendering_mode';
```
jbms, line 51: change from src/neuroglancer/ to neuroglancer/

src/neuroglancer/volume_rendering/volume_render_layer.ts
```diff
@@ -396,6 +460,9 @@ void main() {
               chunkTransform: {channelToChunkDimensionIndices}
             } = transformedSource;
             const {} = transformedSource;
+            const normChunkNumber = chunkNumber / chunks.size;
+            gl.uniform1f(shader.uniform('uChunkNumber'), normChunkNumber);
```
jbms, line 464: could condition calling this on wireFrame mode

src/neuroglancer/volume_rendering/volume_render_layer.ts
```diff
@@ -165,7 +185,42 @@ void emitTransparent() {
...
+                intensityCalculation: `
+    userMain();
+`,
+                beforeColorEmission: ``
+              };
+              break;
+            case VOLUME_RENDERING_MODES.MAX:
+              glslSnippets = {
+                intensityCalculation: `
+    float normChunkValue = toNormalized(getInterpolatedDataValue(0));
```
jbms, line 209: I guess this is fine for experimentation purposes, but it is a bit unfortunate to hardcode this here --- it would be better if the user can somehow determine the input to the `max` function.  For example if the input is inverted the user would need to be able to negate it first.

src/neuroglancer/volume_rendering/volume_render_layer.ts
seankmartin: Great spot, sorry I missed that one - will fix!

PR #495, opened 2023-10-10
feat: change the number of depth samples in volume rendering to a controllable UI parameter

seankmartin:
# Goal

Change the number of depth samples from a fixed 64 to a user-controlled variable. Named "samples per ray".

## Primary changes

1. Create a UI controllable int, default 64, min 1, max 8096 that defines the samples per ray.
2. On update of samples per ray we trigger a redraw and a recompute of the chunk priority, in case this change switches the desired data resolution level. The data resolution level is currently based on the depth of the view frustum divided by the number of samples.
3. Pass the samples per ray into both the chunk worker and the shader to accommodate this.
4. Control the samples per ray in Python.


jbms:
This definitely seems like a good option, but what do you think about using the existing "Resolution" control (that currently does nothing) but modifying it to display the different resolutions in units of "voxel size / view frustum depth"?  That way this "Resolution" control would behave similarly to the other Resolution controls and would no longer be a "display only" control.

seankmartin:
I really like that idea. My hope was that the implementation here would be a stepping stone to something more robust, and that is a good suggestion.

We'd essentially be changing up the logic from the resolution being determined by a user setting of the near and far planes and samples per ray, to the number of samples per ray being determined by a user setting of the near and far planes and resolution level. Am I understanding that right?

jbms:
To be clear, I was imagining that the actual user setting would still be number of samples, as you have it now, it would just be controlled using the (modified) "Resolution (3d)" control rather than a separate control.  In order for that to make sense, the "Resolution (3d)" control would need to display the number of samples in log scale, and plot each resolution according to the number of samples that would be optimal for it, so that it displays in the correct position relative to the vertical bar that indicates the current "number of samples" selection.

seankmartin:
Sure, I understand better now - I still like the idea!

I'll get to making that change in the coming days

python/neuroglancer/viewer_state.py
```diff
@@ -164,6 +164,12 @@ class OpacityTool(Tool):
     TOOL_TYPE = "opacity"
 
 
+@export_tool
+class VolumeRenderingScaleTool(Tool):
```
jbms, line 168: Since this specifies number of depth samples (which is essentially 1/scale) a different name might be better, like volumeRenderingDepthSamples or similar.  Or we could change how it is represented in the JSON state, so that it actually is a "scale" parameter, but that adds other complications since the real scale depends on other things like the view frustum depth.

src/neuroglancer/image_user_layer.ts
```diff
@@ -73,9 +74,10 @@ export class ImageUserLayer extends Base {
   dataType = new WatchableValue<DataType|undefined>(undefined);
   sliceViewRenderScaleHistogram = new RenderScaleHistogram();
   sliceViewRenderScaleTarget = trackableRenderScaleTarget(1);
-  volumeRenderingRenderScaleHistogram = new RenderScaleHistogram();
-  // unused
-  volumeRenderingRenderScaleTarget = trackableRenderScaleTarget(1);
+  volumeRenderingRenderScaleHistogram =
+      new RenderScaleHistogram(VOLUME_RENDERING_RESOLUTION_LOG_SCALE_ORIGIN);
+  volumeRenderingRenderScaleTarget =
+      trackableRenderScaleTarget(2 ** VOLUME_RENDERING_STARTING_LOG_SCALE);
```
jbms, line 80: it would probably be better to add the validation directly in a modified `trackableRenderScaleTarget` rather than just in the widget --- that way the parameter remains within the valid range even if set via the JSON state/Python.

python/neuroglancer/viewer_state.py
seankmartin: I think changing the name is a better option, especially with the complications around the view frustum depth as you mention - thanks Jeremy!

jbms:
Please resolve merge conflicts and then should be ready.

seankmartin:
@jbms merge conflicts fixed!

PR #496, opened 2023-10-10
feat: change image user layer UI labels

seankmartin:
# Goal

To try and make it clearer that opacity and blending don't impact volume rendering, only the cross-section slice view


src/neuroglancer/image_user_layer.ts
```diff
@@ -291,37 +291,37 @@ function makeShaderCodeWidget(layer: ImageUserLayer) {
...
+    ...rangeLayerControl(layer => ({value: layer.opacity})),
+  },
   {
     label: 'Volume rendering (experimental)',
     toolJson: VOLUME_RENDERING_JSON_KEY,
     ...checkboxLayerControl(layer => layer.volumeRendering),
   },
   {
-    label: 'Resolution (3d)',
+    label: 'Resolution indicator (3D)',
```
jbms, line 317: Per the comment in the other PR, the change to this label may be unnecessary if it is changed to control the number of samples.

python/neuroglancer/viewer_state.py
```diff
@@ -158,6 +158,12 @@ class BlendTool(Tool):
     TOOL_TYPE = "blend"
 
 
+@export_tool
+class VolumeRenderingTool(Tool):
+    __slots__ = ()
+    TOOL_TYPE = "volumeRendering"
```
jbms, line 164: Per the comment in the other PR, I think there could be one option that controls the volume rendering mode, with one mode being "off" --- then there wouldn't be a separate tool and python option.

seankmartin: Agreed, I'll change the other PR to have this new mode controllable from Python and remove the new tool from this PR

PR #500, opened 2023-11-02
fix(state_share): new method to include credentials for state sharing

chrisj:
due to chrome introducing restrictions on setting protocols using the URL API

```ts
const x = new URL('http://google.com');
x.protocol = 'middleauth+https:';
```

The second line is silently rejected in chrome 118.

### DIFF
--- a/src/neuroglancer/datasource/state_share.ts
+++ b/src/neuroglancer/datasource/state_share.ts
@@ -72,10 +72,9 @@ export class StateShare extends RefCounted { // call it a widget? no because it 
               body: JSON.stringify(viewer.state.toJSON())
             }, responseJson)
           .then((res) => {
-            const stateUrl = new URL(res);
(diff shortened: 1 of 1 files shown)


src/neuroglancer/datasource/state_share.ts
```diff
@@ -73,8 +73,9 @@ export class StateShare extends RefCounted { // call it a widget? no because it
             }, responseJson)
           .then((res) => {
             const stateUrl = new URL(res);
-            stateUrl.protocol = protocol; // copy protocol in case it contains authentication type
-            const link = `${window.location.origin}/#!${stateUrl}`;
+            const stateUrlWithoutProtocol = stateUrl.toString().split(stateUrl.protocol)[1];
```
jbms, line 76: Split is kind of an odd way to handle this -- instead could just use substring(stateUrl.protocol.length)

chrisj: Yes, made this change

PR #503, opened 2023-11-23
Volume rendering compositing changes and wireframe rendering to vis chunks in volume rendering mode

seankmartin:
Performs front-to-back compositing in three steps:

1. The opacity from the current data point is corrected based on the ratio between the optimal number of depth samples to use for the current data resolution, and the actual number of chosen depth samples. This avoids the output image changing opacity if the volume is over/undersampled.
2. The corrected opacity is composited with the already accumulated opacity along the ray.
3. The resulting composited opacity is multiplied by the already accumulated color, and then the accumulated color and opacity information is updated with the opacity multiplied color and the composited opacity.

Also adds a chunk visualiser to VR if wireframe mode is enabled in settings


src/neuroglancer/volume_rendering/volume_render_layer.ts
```diff
@@ -168,22 +171,32 @@ vec4 outputColor;
 void userMain();
 `);
         defineChunkDataShaderAccess(builder, chunkFormat, numChannelDimensions, `curChunkPosition`);
+        // See Real-Time Volume Graphics, page 12
         builder.addFragmentCode(`
 void emitRGBA(vec4 rgba) {
-  float alpha = rgba.a * uBrightnessFactor;
-  outputColor += vec4(rgba.rgb * alpha, alpha);
+  float opacityCorrectedAlpha = 1.0 - (pow(clamp(1.0 - rgba.a, 0.0, 1.0), uSamplingRatio));
+  float compositedAlpha = (1.0 - outputColor.a) * opacityCorrectedAlpha;
```
jbms, line 178: This is indeed a more accurate handling of transparency within the chunk, but the problem is that this only works if the samples can be depth ordered.  If there is more than one chunk, we can't achieve that easily.

However, currently this shader just adds up the alpha-weighted samples without regard to depth.  Instead it could do the same thing that the order-independent transparency  is doing: https://github.com/google/neuroglancer/blob/7cee81c8a60baec4263627f5394172eed0751b22/src/neuroglancer/perspective_view/panel.ts#L87

That would fix the artifacts at chunk boundaries, and would probably result in at least somewhat better normalization.

Fundamentally I think some type of gain factor may be unavoidable, but perhaps there could be a mechanism to adjust it automatically, e.g. based on the actual brightness distribution of the rendered result.  While the gain could effectively just be included in the line-segment-based transfer function, I expect users will want to control the overall brightness as they change other view parameters, and having to move multiple points in the transfer function would be inconvenient.

seankmartin:
This pull request implements four main features:
1. OIT is performed along rays during volume rendering ray marching. This is to try and remove chunk border artifacts.
2. During ray marching, the value stored in the offscreen Z buffer is compared against the depth computed at the current ray position. If the ray has been deemed to pass behind an opaque object, the compositing loop for that ray ends.
3. Wireframe rendering mode shows volume rendering chunks by a different color.
4. A gain parameter has been included in volume rendering mode, to heighten or lessen the opacity value by scaling with the gain.

Additionally, the volume rendering flag has been bound to a Python controllable tool.

I think it should be good for another round of review if you get the time @jbms. Thanks again for all the thoughtful input on this one so far, really helpful and much appreciated!

src/image_user_layer.ts
```diff
@@ -98,8 +98,10 @@ import {
   ShaderControls,
 } from "#/widget/shader_controls";
 import { Tab } from "#/widget/tab_view";
+import { trackableFiniteFloat } from "#/trackable_finite_float";
 
 const OPACITY_JSON_KEY = "opacity";
+const GAIN_JSON_KEY = "gain";
```
jbms, line 104: maybe rename to volumeRenderingGain since it only applies to volume rendering?

Also when I've created this control manually as a UI control, I've typically found it convenient to give it an exponential scale so that it can more easily cover a larger range, i.e. the slider ranges from -10 to 10 and the actual gain is `exp(gain)`.  But that could also be handled purely in the slider control.  Maybe you've also found the linear gain to work okay.

src/perspective_view/render_layer.ts
```diff
@@ -50,6 +50,11 @@ export interface PerspectiveViewRenderContext
    * Specifies whether there was a previous pick ID pass.
    */
   alreadyEmittedPickID: boolean;
+
+  /**
+   * Specifies the ID of the depth frame buffer texture to query during rendering.
+   */
+  depthBufferTextureID?: WebGLTexture | null;
```
jbms, line 57: Since `WebGLTexture` is a garbage-collected object rather than a numeric identifier, it may make sense to just call this `depthBufferTexture` instead.

jbms:
Thanks!  This looks good --- I left just a few minor comments.

src/image_user_layer.ts
seankmartin: Great, thanks! Linear worked pretty well, but I prefer the exponential scale after trying it out, thanks for the suggestion on that.
Not sure if we should hint at the scale in the UI somehow now, or if the name Gain (3D) is clear enough. Happy either way, but let me know if you have thoughts on that

src/image_user_layer.ts
```diff
@@ -125,6 +127,7 @@ export class ImageUserLayer extends Base {
   dataType = new WatchableValue<DataType | undefined>(undefined);
   sliceViewRenderScaleHistogram = new RenderScaleHistogram();
   sliceViewRenderScaleTarget = trackableRenderScaleTarget(1);
+  volumeRenderingGain = trackableFiniteFloat(1);
```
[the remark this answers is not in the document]
seankmartin: Yes, good catch, sorry about that!

PR #552, opened 2024-03-19
Max projection volume rendering with picking

seankmartin:
# Min and max projection VR modes with picking

## Adds:
1. Min and Max projection modes to the volume rendering. Both of these work by computing the min/max over rays within single chunks in the shader, and then combining across chunks via depth testing.
2. Experimental picking works in Min and Max mode because we know the depth of the max/min sample.
3. First pass documentation about volume rendering.

## Experimental change, open to thoughts:
Moved the axis helper visualisation to draw after volume rendering, so that it draws on top of it. Previously the axis helper was getting occluded in high-opacity volume rendering.

## Possible future changes
* Picking in volume rendering mode = "on", possibly on CPU
* Picking across a small window of data in min/max mode


seankmartin:
Hi @fcollman,

There is a bit of an admittedly unfortunate pattern needed for RGB and RGBA emitters. There needs to be an `emitIntensity(float_in_0_1_range)` call before the color emission. For example, in the shader:

```glsl
#uicontrol invlerp normalized
void main() {
  float val = normalized();
  emitIntensity(val);
  emitRGB(colormapJet(val));
}
```

It's possible we could do something like having a default intensity in RGB mode which is the max(color) or a normalized sum of the color, and in RGBA, the opacity is used. In that case, `emitIntensity` would override the default intensity. In that way, the RGB and RGBA modes would still work without needing to call `emitIntensity` - but could still be overwritten.

For the picking in non-max mode, that's a little trickier. We could provide a dropdown in the UI, with modes "off", "max" and "min", something like "picking in non-max VR" (probably a better name). This dropdown would enable performing a second rendering pass if `volume_rendering="on"` which does a max/min projection in the background for picking purposes.

What do you think of these? (also for @jbms)

fcollman:
I think as long as there is a route to rendering that has picking and supports RGB and RGBA that would be good… I would support an automatic fallback for RGB and RGBA intensities.

src/perspective_view/panel.ts
```diff
@@ -646,6 +705,59 @@ export class PerspectivePanel extends RenderedDataPanel {
...
+  private get maxProjectionConfiguration() {
+    let maxProjectionConfiguration = this.maxProjectionConfiguration_;
+    if (maxProjectionConfiguration === undefined) {
+      maxProjectionConfiguration = this.maxProjectionConfiguration_ =
+        this.registerDisposer(
+          new FramebufferConfiguration(this.gl, {
+            colorBuffers: [
+              new TextureBuffer(
+                this.gl,
+                WebGL2RenderingContext.RGBA32F,
```
jbms, line 717: since we aren't accumulating things in the color buffer we can just use regular 8-bit per channel representation

src/perspective_view/panel.ts
```diff
@@ -820,10 +932,6 @@ export class PerspectivePanel extends RenderedDataPanel {
       gl.disable(WebGL2RenderingContext.BLEND);
     }
 
-    if (this.viewer.showAxisLines.value) {
```
jbms, line 823: I'm inclined to keep the axis lines in their original (depth-correct) rendering position for now --- users could reduce the alpha value used by the max projection if they want to see them

src/perspective_view/panel.ts
```diff
@@ -856,10 +987,70 @@ export class PerspectivePanel extends RenderedDataPanel {
         if (renderLayer.isTransparent) {
           renderContext.depthBufferTexture =
             this.offscreenFramebuffer.colorBuffers[OffscreenTextures.Z].texture;
+        }
+        const isVolumeRenderingLayer = "mode" in renderLayer;
```
jbms, line 991: I guess this may be to avoid a circular dependency but I think a less hacky approach would be better here.

src/volume_rendering/trackable_volume_rendering_mode.ts
```diff
@@ -0,0 +1,23 @@
+import { TrackableEnum } from "#src/util/trackable_enum.js";
+
+export enum VOLUME_RENDERING_MODES {
```
jbms, line 3: This should be VolumeRenderingModes

src/layer/image/index.ts
```diff
@@ -106,7 +108,7 @@ const SHADER_JSON_KEY = "shader";
 const SHADER_CONTROLS_JSON_KEY = "shaderControls";
 const CROSS_SECTION_RENDER_SCALE_JSON_KEY = "crossSectionRenderScale";
 const CHANNEL_DIMENSIONS_JSON_KEY = "channelDimensions";
-const VOLUME_RENDERING_JSON_KEY = "volumeRendering";
+const VOLUME_RENDERING_MODE_JSON_KEY = "volumeRenderingMode";
```
jbms, line 111: While the existing volume rendering support was marked experimental, we should still retain the existing json key (and handle true/false) to avoid breaking compatibility with existing states (and urls).

src/perspective_view/panel.ts
seankmartin: Thanks Jeremy, I did have this to avoid a circular dependency, but I agree that should be changed. I've used the `isVolumeRendering` getter on the layer itself after review, let me know if that's better. Happy to look at it again if not!

src/volume_rendering/volume_render_layer.ts
```diff
@@ -242,31 +242,59 @@ export class VolumeRenderingRenderLayer extends PerspectiveViewRenderLayer {
...
+  userIntensity = -1.0;
+  defaultIntensity = -1.0;
+}
+`;
+            glsl_emitIntensity = `
+float convertIntensity(float value) {
+  return clamp(value, 0.0, 1.0);
+}
+${emitFunctions}
+`;
```
seankmartin, line 279: This is to handle a default intensity which would be defaulted to in min/max mode, which can be overwritten if the user calls `emitIntensity`.

seankmartin:
Hi @jbms,

I implemented something along the lines of the heuristic we discussed for the default max projection intensity in [this commit](https://github.com/google/neuroglancer/pull/552/commits/3725d11705e2030a3d98bc965fe1ac826aa8380a). The value of the first invlerp called is used as the default intensity, if there is an invlerp call.

It should work with transfer functions too, because they call an invlerp.

Please let me know if you had something else mind, either in the heuristic used, or the implementation.

src/layer/image/index.ts
```diff
@@ -106,7 +108,9 @@ const SHADER_JSON_KEY = "shader";
 const SHADER_CONTROLS_JSON_KEY = "shaderControls";
 const CROSS_SECTION_RENDER_SCALE_JSON_KEY = "crossSectionRenderScale";
 const CHANNEL_DIMENSIONS_JSON_KEY = "channelDimensions";
-const VOLUME_RENDERING_JSON_KEY = "volumeRendering";
+// The volume rendering mode key is newer, but we still need to support the old one for backwards compatibility.
+const VOLUME_RENDERING_COMPAT_JSON_KEY = "volumeRendering";
```
jbms, line 112: I think it would be simpler to just keep using the old key, and simply extend the allowed set of values.

src/perspective_view/panel.ts
```diff
@@ -832,17 +944,40 @@ export class PerspectivePanel extends RenderedDataPanel {
     );
 
     if (hasTransparent) {
-      // Draw transparent objects.
-      gl.depthMask(false);
-      gl.enable(WebGL2RenderingContext.BLEND);
+      //Draw transparent objects.
+      const { maxProjectionConfiguration } = this;
+      const bindMaxProjectionBuffer = () => {
```
jbms, line 949: We could add a `hasMaxProjection` variable so that we don't create/clear the max projection framebuffers if not needed.

src/volume_rendering/volume_render_layer.ts
jbms: Can you comment on the reason for mixing and these various epsilon values?

src/volume_rendering/volume_render_layer.ts
```diff
@@ -220,7 +236,63 @@ export class VolumeRenderingRenderLayer extends PerspectiveViewRenderLayer {
...
+            }
+            glsl_rgbaEmit = `
+void emitRGBA(vec4 rgba) {
+  float newIntensity = getIntensity();
+  float intensityChanged = step(savedIntensity, newIntensity - 0.00001);
+  float alpha = clamp(rgba.a, 0.0, 1.0);
+  outputColor = mix(outputColor, vec4(rgba.rgb * alpha, alpha), intensityChanged);
+  savedIntensity = mix(savedIntensity, newIntensity, intensityChanged); 
+  savedDepth = mix(savedDepth, depthAtRayPosition, intensityChanged);
+  userIntensity = -1.0;
```
jbms, line 286: I think it might be clearer and simpler to check the intensity in the outer loop that calls userMain -- that way the user can call emitIntensity before or after calling emitRGBA

src/volume_rendering/volume_render_layer.ts
seankmartin: I've now removed most of these checks with the new heuristic we are using, where the value of the invlerp is the default intensity. Previously, the checks were to handle this.

One check remains, which I think is needed and can't be avoided. That is to compare the intensity value from the current ray position against the max intensity encountered along a ray in this chunk and pick the strictly bigger one. Then, if the intensity changed (i.e the new intensity was bigger), update the saved Color and Depth values.

There are two ways I can think of to implement this, and if you are aware of something clearer or more efficient @jbms that would be great!

Option 1: Step and Mix pattern (see [this branch](https://github.com/MetaCell/neuroglancer/tree/feature/max-with-step-and-mix))

```glsl
float intensityChanged = step(savedIntensity, newIntensity - 0.00001);
savedIntensity = mix(savedIntensity, newIntensity, intensityChanged); 
savedDepth = mix(savedDepth, depthAtRayPosition, intensityChanged);
outputColor = mix(outputColor, newColor, intensityChanged);
```

Option 2: Ternary operator (on this branch):

```glsl
bool intensityChanged = newIntensity > savedIntensity;
savedIntensity = intensityChanged ? newIntensity : savedIntensity; 
savedDepth = intensityChanged ? depthAtRayPosition : savedDepth;
outputColor = intensityChanged ? newColor : outputColor;
```

If you have a preference based on performance/code clarity etc. or have a suggestion just let me know.

Thanks!

seankmartin:
For consistency with other emission modes, I've removed the call to `emitIntensity(value)` for `emitGrayscale(value)`.  Now only for `emitTransparent()`, emitIntensity(0.0) is called.

PR #575, opened 2024-04-12
Update status.ts DOM text reinterpreted as HTML {Patch} 

Shivam7-1:
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML. Always be cautious when dealing with user input or dynamic content to prevent security risks

### DIFF
--- a/src/status.ts
+++ b/src/status.ts
@@ -65,7 +65,7 @@ export class StatusMessage {
     }
   }
   setHTML(text: string, makeVisible?: boolean) {
-    this.element.innerHTML = text;
+    this.element.innerText = text;
     if (makeVisible) {
       this.setVisible(true);
     }


Shivam7-1:
Hi @jbms Could You Please Review Above PR
Regards

Shivam7-1:
Hi @jbms
I noticed my PR was closed. Could you please provide some feedback on why it was closed? I'm eager to learn and improve.
Thank you

NeilFraser:
@Shivam7-1 If you look at the code, you'll see that there are two functions, `setText` and `setHTML`:
```
  setText(text: string, makeVisible?: boolean) {
    this.element.textContent = text;
    if (makeVisible) {
      this.setVisible(true);
    }
  }
  setHTML(text: string, makeVisible?: boolean) {
    this.element.innerHTML = text;
    if (makeVisible) {
      this.setVisible(true);
    }
  }
```
The `setText` function sets `.textContent` which is a safe way to assign text to a node, whereas the `setHTML` function sets `.innerHTML` specifically because it intends to creat HTML tags.  Your proposed change would make `setHTML` behave the same as `setText` with any HTML tags printed as raw sources on the screen for the user to see.

In general it's not a good idea to submit code changes to a product you don't use, without testing it, and without reading and understanding the surrounding code.  Blindly submitting changes breaks things.

PR #576, opened 2024-04-15
chore: output declaration files for neuroglancer packages

chrisj:
esbuild does not support generating declaration files https://github.com/evanw/esbuild/issues/95

### DIFF
--- a/build_tools/build-package.ts
+++ b/build_tools/build-package.ts
@@ -14,12 +14,31 @@ import fs from "node:fs/promises";
 import path from "node:path";
 import esbuild from "esbuild";
 import { glob } from "glob";
+import ts from "typescript";
 import yargs from "yargs";
 
 const rootDir = path.resolve(import.meta.dirname, "..");
 
-async function buildPackage(options: { inplace?: boolean }) {
-  const { inplace = false } = options;
+function buildDeclarationFiles(
+  fileNames: string[],
+  options: ts.CompilerOptions,
+): void {
+  options = {
+    allowJs: true,
+    declaration: true,
+    emitDeclarationOnly: true,
+    ...options,
+  };
+  const host = ts.createCompilerHost(options);
+  const program = ts.createProgram(fileNames, options, host);
+  program.emit();
+}
+
+async function buildPackage(options: {
+  inplace?: boolean;
+  declaration?: boolean;
+}) {
+  const { inplace = false, declaration = false } = options;
 
   const srcDir = path.resolve(rootDir, "src");
   const outDir = inplace ? rootDir : path.resolve(rootDir, "dist", "package");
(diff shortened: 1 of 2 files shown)


build_tools/build-package.ts
```diff
@@ -45,13 +64,26 @@ async function buildPackage(options: { inplace?: boolean }) {
...
-    entryPoints: typescriptSources.map((name) => path.resolve(srcDir, name)),
+    entryPoints,
     outbase: srcDir,
     bundle: false,
     outdir: libDir,
   });
 
+  if (declaration) {
+    buildDeclarationFiles(entryPoints, {
+      allowJs: true,
```
chrisj, line 80: need to remove these superfluous arguments

chrisj:
I improved the quality of the declarations by using the tsconfig file. The bulk of the change comes from `"moduleResolution": "bundler",`

It does lose a little bit of type information for some reason but it adds a lot more than it removes
```
-    spatiallyIndexedSources: Set<Borrowed<AnnotationGeometryChunkSource>>;
+    spatiallyIndexedSources: Set<AnnotationGeometryChunkSource>;
```

```
-    references: Map<AnnotationId, Borrowed<AnnotationReference>>;
-    localUpdates: Map<AnnotationId, LocalUpdateUndoState>;
+    references: Map<string, AnnotationReference>;
+    localUpdates: Map<string, LocalUpdateUndoState>;
```

jbms:
Thanks --- we could probably just emit the declarations all the time --- I don't think there is a time when we don't want the declarations.

jbms:
Having the type aliases resolved in some cases is fine, I think, and I agree that some annotations are better than none.

When I tried to generate declarations previously I ran into some errors due to some exported things having types depending on non-exported types, I think.  But I guess it works now for some reason.

chrisj:
Removed the flag making it optional. Added that in case the performance benefit of leaving it off was useful.

I agree imperfect types are far better than none. I was fortunate that this solution https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#getting-the-dts-from-a-javascript-file was very easy to drop in

chrisj:
I just realized `const host = ts.createCompilerHost(options);` is probably unnecessary but it does change the declaration files. Removes ".ts" from the imports
```
-    get changed(): import("../util/signal.ts").Signal<() => void>;
+    get changed(): import("../util/signal").Signal<() => void>;
```
Checking to make sure that isn't a problem.

chrisj:
`check(new Uint64(1423829346, 3183191017), 28);` in `toString parseString round trip`is a real error. I'm trying to debug it

`check(new Uint64(0, 79741775), 29);` also fails, and it seems to fail with any low value.

79741775 seems to be the smallest high that causes this test to fail with base 29.

PR #583, opened 2024-04-25
perf(datasource/graphene): group together segment id requests within the same bounding box (leaves_many)

chrisj:
Implemented using a proxy object with a setTimeout of 0ms.

The downside to this approach is that neuroglancer still thinks there are many outgoing requests when it applies "concurent chunk requests" requiring users to set that value to a high numbers. 10k is good but even higher leads to quicker loading from my tests.


chrisj:
@jbms I'm now using `get downloadSlots(): number` to make the chunk requests "free" when I can bundle them with a previous request.

Also cleaned up/updated the code.

chrisj:
@jbms if you have time, could you take another look at this PR? I think it is in a good state.

PR #585, opened 2024-04-30
Feature/niftizarr

MikeSchutzman:
Created Nifti Zarr file

Used Volume Logic from Zarr but modelTransform logic from Nifti


jbms:
Can you explain a bit more what the purpose of this is?

jbms:
It sounds like your goal here is to just specify an affine transform that gets applied?

I think a more reasonable way to achieve that would be via a change to the OME metadata --- possibly the already-proposed OME coordinate transforms stuff would already achieve what you want once supported in Neuroglancer.

PR #589, opened 2024-05-20
feat(volume-rendering): picking, depth testing, and blending improvements

seankmartin:
Allows picking via a second pass through max projection in volume rendering. The second pass only runs when the camera is still currently.

Also includes a fix for cases where a non-transparent object was in the middle of a volume rendering chunk. Sometimes the early depth testing would prevent some volume rendering which should have been performed. Now, the volume rendering always performs depth testing during the fragment shader, by computing the depth along rays, and comparing that to a depth value that is sampled from a texture buffer storing the Z value for non-transparent objects.

One other improvement is included, which is to blend together the volume rendering buffer and regular transparent buffer using OIT. Before, the VR buffer rendered to the main color buffer, and then the transparent buffer was rendered to the main color buffer. Now the VR buffer and transparent buffer are combined, then the combination rendered to the main color buffer.

image


jbms:
This looks good, can you please rebase it so that it can be merged?

Thanks

PR #591, opened 2024-05-21
Feature - adaptive downsampling of volume rendering based on framerate

seankmartin:
The overall idea is that a setting (currently on by default) would allow a global downsampling in XY resolution for 3D panels of transparent layers during camera movement if the framerate drops below a certain threshold.

All of this is open to change, but currently:

Target FPS is 60
The mean framerate over a running 6 frames after a camera move is initiated is calculated
If this median framerate drops below 60, a downsampling rate of 2, 4, or 8 is chosen based on how far from 60 the framerate is.
If a downsample is detected, then during this camera movement, the downsampling rate will never go back above the highest downsample rate detected for about the last second
When the camera stops moving, a fresh full resolution drawing is requested after 300ms

To update:
1. Only DS for VR, new buffer
2. Make timing easy to change so that GL query can slot in as needs be
3. More global setting for camera in constant movement. Discrete movements won't trigger DS

### DIFF
--- a/src/perspective_view/panel.ts
+++ b/src/perspective_view/panel.ts
@@ -288,7 +288,7 @@ export class PerspectivePanel extends RenderedDataPanel {
   // if a high downsample rate is applied, it persists for a few frames
   // to avoid flickering when the camera is moving
   private frameRateCalculator = new DownsamplingBasedOnFrameRateCalculator(
-    10 /* numberOfStoredFrameDeltas */,
+    6 /* numberOfStoredFrameDeltas */,
     8 /* maxDownsamplingFactor */,
     16 /* desiredFrameTimingMs */,
     60 /* downsamplingPersistenceDurationInFrames */,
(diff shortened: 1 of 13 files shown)


src/viewer.ts
```diff
@@ -295,6 +295,7 @@ class TrackableViewerState extends CompoundTrackable {
     this.add("layers", viewer.layerSpecification);
     this.add("showAxisLines", viewer.showAxisLines);
     this.add("wireFrame", viewer.wireFrame);
+    this.add("enableAdaptiveDownsampling", viewer.enableAdaptiveDownsampling);
```
jbms, line 298: Add corresponding option to Python viewer_state.py

seankmartin: Thanks @jbms ! I added it in https://github.com/google/neuroglancer/pull/591/commits/60ca39c0389d02e623c937d39c05350d7d1a026e

It looks like the wireframe setting was also missing, so I added it too. But let me know if that's not right https://github.com/google/neuroglancer/pull/591/commits/c9ef6d925deff20bf99dad8e6973be07aa53900e

src/perspective_view/panel.ts
```diff
@@ -102,6 +105,12 @@ export enum OffscreenTextures {
   NUM_TEXTURES = 3,
 }
 
+enum TransparentRenderingState {
+  Transparent = 0,
```
jbms, line 109: consistent casing

PR #592, opened 2024-05-21
Feature - CDF in invlerp if only 3D view shown of image layer

seankmartin:
Allows invlerps to show CDFs even if only 3D data is being rendered.

Adds to the volume rendering code, with the overall idea as:

Use the same buffers that were created for the 2d panels histograms
Defines a vertex shader that samples a random part of a chunk for each input point, and based on the bin of the histogram it would fall in, increments the count for that bin by one
Run this vertex shader over each visible chunk of the volume renderer.
At the moment, the 3d histogram calculations are not run when the camera is moving, when there are slice views present, in wireframe mode, or if all histograms are currently hidden.

Sample rate in each chunk is chosen based on the number of visible chunks and the overall desired number of total samples.


src/perspective_view/panel.ts
```diff
@@ -251,6 +253,7 @@ export class PerspectivePanel extends RenderedDataPanel {
   protected visibleLayerTracker: Owned<
     VisibleRenderLayerTracker<PerspectivePanel, PerspectiveViewRenderLayer>
   >;
+  private redrawAfterMoveTimeOutId: number = -1;
```
jbms, line 256: can use lodash debounce for this

src/rendered_data_panel.ts
```diff
@@ -616,7 +617,9 @@ export abstract class RenderedDataPanel extends RenderedPanel {
     registerActionListener(element, "move-to-mouse-position", () => {
       const { mouseState } = this.viewer;
       if (mouseState.updateUnconditionally()) {
+        this.isMovingToMousePosition = true;
```
jbms, line 620: as discussed in the meeting, it would be better for this to be more global than a single RenderedDataPanel

src/volume_rendering/volume_render_layer.ts
```diff
@@ -817,6 +1026,148 @@ void main() {
...
+            1,
+            WebGL2RenderingContext.RGBA,
+            WebGL2RenderingContext.FLOAT,
+            tempBuffer,
+          );
+          const tempBuffer2 = new Float32Array(256);
+          for (let j = 0; j < 256; ++j) {
+            tempBuffer2[j] = tempBuffer[j * 4];
+          }
+          console.log("histogram%d", i, tempBuffer2.join(" "));
```
jbms, line 1165: Note: console.log doesn't support printf-style format syntax

PR #601, opened 2024-06-20
Feature - continous camera movement tracking

seankmartin:
The context can be used to mark the camera continuously moving. The context will mark the camera movement as completed after 300ms using lodash debounce, and emit a signal when this happens. At the moment, only the perspective panel listens for this signal, and uses it to trigger a redraw request if volume rendering is enabled on that panel.

The actions that flag the start of a continuous camera movement are:
* Zooming via mouse, touch, or keys
* Rotating via mouse, touch or keys
* Changing the depth planes

### DIFF
--- a/src/display_context.ts
+++ b/src/display_context.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { debounce } from "lodash-es";
+
 import type { FrameNumberCounter } from "#src/chunk_manager/frontend.js";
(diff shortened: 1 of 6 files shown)


seankmartin:
The changes here have been integrated into https://github.com/google/neuroglancer/pull/589, but 589 can be updated again based on changes from this PR

seankmartin:
Closed as #591 included these changes

PR #611, opened 2024-07-09
fix(datasource/zarr) `HEAD` to get content length

d-v-b:
Some static file servers (e.g., the nodejs program [`serve`](https://github.com/vercel/serve)) will handle a successful HTTP range request (i.e., status code 206) _without_ including a `content-range` header in that response. Perhaps `content-range` is omitted because it's not on the list of [CORS safelisted response headers](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header), but that's just a guess. This impairs neuroglancer's ability to read sharded zarr v3 chunks when the shard index is configured to reside at the end of a chunk, as in this case neuroglancer makes a GET with a suffixLength byte range request but if the response from the server does not include the `content-length` header, it's not possible to resolve an offset + length representation of the body of that request.

This PR addresses this problem by falling back to a HEAD request against the shard to figure out its length. I am re-using a routine already employed for hosts that don't support suffixLength byte range requests. 

I can confirm from local testing that this patch enables reading sharded zarr v3 arrays from the nodejs `serve` application, which are failing on main, but I would appreciate any pointers on places inside the neuroglancer codebase to modify / add a test.

### DIFF
--- a/src/kvstore/special/index.ts
+++ b/src/kvstore/special/index.ts
@@ -51,6 +51,33 @@ class SpecialProtocolKvStore implements ReadableKvStore {
     public credentialsProvider: SpecialProtocolCredentialsProvider,
     public baseUrl: string,
   ) {}
+
+  async getObjectLength(url: string, options: ReadOptions){
+    // Use a HEAD request to get the length of an object
+    const { cancellationToken = uncancelableToken } = options;
+    const headResponse = await cancellableFetchSpecialOk(
+      this.credentialsProvider,
+      url,
+      { method: "HEAD" },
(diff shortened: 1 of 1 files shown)


davidackerman:
@jbms I also am running into a cors error when trying to load zarrv3

jbms:
@davidackerman Does this patch fix the issue for you?

davidackerman:
seems to, yes!
neuroglancer-demo.appspot:
image

this patch:
image

jbms:
Thanks!

PR #616, opened 2024-07-17
ome.ts: update with dev2 version from the ngff challenge

joshmoore:
The current proposal for updating OME-Zarr to support Zarr v3 wraps the NGFF metadata in the Zarr metadata within an "ome" block. This checks for that block and if it exists unwraps the necessary metadata.

This format is **not** finalized.

see:
 - https://forum.image.sc/t/ome2024-ngff-challenge/97363
 - https://ngff.openmicroscopy.org/rfc/2/

### DIFF
--- a/src/datasource/zarr/ome.ts
+++ b/src/datasource/zarr/ome.ts
@@ -39,7 +39,7 @@ export interface OmeMultiscaleMetadata {
   coordinateSpace: CoordinateSpace;
 }
 
-const SUPPORTED_OME_MULTISCALE_VERSIONS = new Set(["0.4", "0.5-dev"]);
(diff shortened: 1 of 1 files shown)


joshmoore:
Think it should be disabled. @normanrz?

Note: now deployed this branch on the challenge S3 to test a workaround for https://github.com/will-moore/ome2024-ngff-challenge/pull/1

normanrz:
Yes, the OME-Zarr 0.5 spec is strict about not supporting Zarr v2.

jbms:
Can you complete the CLA, and modify this to only look for an `ome` attribute if zarr version is 3?  You could modify `parseOmeMetadata` to take in the zarr version.

joshmoore:
Closing in favor of #670. (Opening the second PR finally fixed a 400 error from the CLA refresh code)

PR #628, opened 2024-07-31
Fix: resolves browser test failures on Mac

allison-truhlar:
Addresses issue #627. 

- Adds `playwright` as a dependency; removes `webdriverio`. 
- Configures Vitest to use `playwright` as the browser provider rather than `webdriverio`.
- Adds `npx playwright install` to the `test` and `test:watch` commands in `package.json`. This is to ensure `playwright` updates after the initial install, to avoid getting the error `Error: browserType.launch`, followed the by the instructions: "Looks like Playwright Test or Playwright was just installed or updated. Please run the following command to download new browsers: `npx playwright install`.

All tests pass in macOS 14.5 using this configuration.

I haven't completed the Google CLA yet - I am opening this as a draft as I wait for instructions on how to complete it at Janelia.

Issue #627, opened 2024-07-31
Browser unit test failures on Mac


allison-truhlar:
When using `npm run test` on a Mac running Sonoma 14.5, 61/269 tests fail. I receive the same error message for all: `Error: WebGL not supported.` I have confirmed through visiting `chrome://gpu/` in the Chrome browser that WebGL is enabled; in addition, the local server started through `npm dev-server` works to visualize files. I am working on resolving the issue and plan to open a pull request.

hubbardp:
I get the same failure on macOS 12.1 and 14.0 beta, with error messages like this:

 FAIL  |browser| src/single_mesh/frontend.browser_test.ts [ src/single_mesh/frontend.browser_test.ts ]
Error: WebGL not supported.
 ❯ initializeWebGL src/webgl/context.ts:41:10
     39|   const gl = <GL>canvas.getContext("webgl2", options);
     40|   if (gl == null) {
     41|     throw new Error("WebGL not supported.");
       |          ^
     42|   }
     43|   gl.memoize = new Memoize<any, RefCounted>();
 ❯ webglTest src/webgl/testing.ts:25:9
 ❯ fragmentShaderTest src/webgl/shader_testing.ts:367:2
 ❯ src/single_mesh/frontend.browser_test.ts:43:4

Interestingly, I do not get failures on Windows 11 Enterprise.

PR #630, opened 2024-08-01
Fix browser test errs on mac

allison-truhlar:
Addresses issue https://github.com/google/neuroglancer/issues/627.

- Adds `playwright` as a dependency; removes `webdriverio`.
- Configures Vitest to use `playwright` as the browser provider rather than `webdriverio`.
- Adds `npx playwright install` to the `test` and `test:watch` commands in package.json. This is to ensure playwright updates after the initial install, to avoid getting the error Error: browserType.launch, followed the by the instructions: "Looks like Playwright Test or Playwright was just installed or updated. Please run the following command to download new browsers: npx playwright install.

All tests pass in macOS 14.5 using this configuration.

Replaces PR #628. These commits use the correct email address associated with the Janelia CLA.


jbms:
Thanks!

Does this allow the WebGL tests to pass on macOS?  (See https://github.com/google/neuroglancer/commit/33f5eb5e4ceef81d221436764bbdc438724abbcd#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721)

allison-truhlar:
I'm not sure - I just removed the if statement in the GitHub action to allow the tests including WebGL to run.

jbms:
It looks like this still does not allow the tests to pass on macOS on github actions, but hopefully it will work in the future after https://github.com/microsoft/playwright/issues/30585 is fixed.

Per the description here (https://playwright.dev/docs/library#key-differences) does it work to just add @playwright/browser-chromium as a dev dependency rather than having to do `npx playwright install`?

allison-truhlar:
In the latest commit, I tested adding @playwright/browser-chromim as a dev dependency and removing `npx playwright install` from the test commands - I can confirm this works.

mkitti:
```
Error: browserType.launch: Executable doesn't exist at /Users/runner/Library/Caches/ms-playwright/chromium-1124/chrome-mac/Chromium.app/Contents/MacOS/Chromium
```

Would adding [`npx playwright install-deps chromium`](https://github.com/microsoft/playwright-github-action?tab=readme-ov-file#to-install-dependencies-with-a-cli) help resolve the client tests here?

allison-truhlar:
@mkitti Which version of the PR did you get this error from? Is this after I added @playwright/browser-chromium as a dev dependency?

mkitti:
Yes. You can see it in the Github Actions tests for client (macos-latest):
https://github.com/google/neuroglancer/actions/runs/10289555738/job/28497954784?pr=630#step:14:18

allison-truhlar:
Added `npx playwright install-deps chromium` to the workflow, per @mkitti's suggestion, to see if this resolves the test failure in GitHub actions.

jbms:
Thanks for your help with this, and I'm sorry for the very long delay in responding.  In the meantime I did end up incorporating this PR into this massive commit 9d3a6180f4652560fd12a2761e32144afb3bd7f3.  I apologize that I forgot to mention you as an acknowledgement in that commit message.

PR #636, opened 2024-08-13
feat(segmentation_user_layer): add individual segment color picker tool

chrisj:
replaces https://github.com/google/neuroglancer/pull/378

has a white border when widget has a set value, right click unsets the value

Screenshot 2024-08-13 at 2 31 39 PM

### DIFF
--- a/src/segmentation_display_state/frontend.ts
+++ b/src/segmentation_display_state/frontend.ts
@@ -47,14 +47,19 @@ import { observeWatchable, registerNestedSync } from "#src/trackable_value.js";
 import { isWithinSelectionPanel } from "#src/ui/selection_details.js";
 import type { Uint64Map } from "#src/uint64_map.js";
 import { setClipboard } from "#src/util/clipboard.js";
-import { useWhiteBackground } from "#src/util/color.js";
+import {
+  packColor,
+  serializeColor,
+  TrackableRGB,
+  useWhiteBackground,
+} from "#src/util/color.js";
 import { RefCounted } from "#src/util/disposable.js";
 import { measureElementClone } from "#src/util/dom.js";
-import type { vec3 } from "#src/util/geom.js";
-import { kOneVec, vec4 } from "#src/util/geom.js";
+import { kOneVec, vec3, vec4 } from "#src/util/geom.js";
 import { NullarySignal } from "#src/util/signal.js";
 import { Uint64 } from "#src/util/uint64.js";
 import { withSharedVisibility } from "#src/visibility_priority/frontend.js";
+import { ColorWidget } from "#src/widget/color.js";
 import { makeCopyButton } from "#src/widget/copy_button.js";
 import { makeEyeButton } from "#src/widget/eye_button.js";
 import { makeFilterButton } from "#src/widget/filter_button.js";
@@ -347,6 +352,8 @@ const segmentWidgetTemplate = (() => {
   filterElement.classList.add("neuroglancer-segment-list-entry-filter");
   const filterIndex = template.childElementCount;
   template.appendChild(filterElement);
+  const colorWidgetIndex = template.childElementCount;
+  template.appendChild(ColorWidget.template());
   return {
     template,
     copyContainerIndex,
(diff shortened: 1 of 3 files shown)


chrisj:
I still need to add the build time option to enable this feature

stuarteberg:
>I still need to add the build time option to enable this feature

This seems like a generally useful feature... can it be enabled by default?

fcollman:
@jbms can we take a look at this PR? I do think having this enabled by default for neuromancer-demo.appspot.com would be a good idea, but I'll leave that up to you.

seankmartin:
I can try look in more detail, but this is a really nice feature, thanks! The only comments I have at a high level is that the build flag removing this feature from the frontend seems unfortunate since I think it would be widely helpful. I read the old PR for context. Could it instead be the case that the build flag just changes the entry point into the color changer? So with the flag for the separate widget you get what you have now. And without the flag you just get that alt-click on the ID element opens a color picker?

Also would be great if the color picker had a tooltip, especially to incidate the right click to clear set color functionality

src/segmentation_display_state/frontend.ts
```diff
@@ -528,6 +546,34 @@ function makeRegisterSegmentWidgetEventHandlers(
       const { selectedSegments } = displayState.segmentationGroupState.value;
       selectedSegments.set(id, !selectedSegments.has(id));
     });
+    const trackableRGB = new TrackableRGB(vec3.fromValues(0, 0, 0));
```
jbms: I think it would be better to just add the `<input type=color>` element to the template and handle mouse events in the `onMoueDown` handler defined above.

The `ColorWidget` abstraction doesn't seem to help here.

src/segmentation_display_state/frontend.ts
```diff
@@ -653,14 +699,24 @@ export class SegmentWidgetFactory<Template extends SegmentWidgetTemplate> {
...
     setSegmentIdElementStyle(
       idContainer.children[template.idIndex] as HTMLElement,
-      getBaseObjectColor(this.displayState, mapped) as vec3,
+      color,
+      !SEGMENT_LIST_COLOR_WIDGET_ENABLED &&
+        !!this.displayState?.segmentStatedColors.value.has(mapped),
     );
+    children[template.colorWidgetIndex] as HTMLInputElement;
+    if (SEGMENT_LIST_COLOR_WIDGET_ENABLED) {
+      setColorWidgetColor(
```
jbms: Currently if `SEGMENT_LIST_COLOR_WIDGET_ENABLED` is false, then when you open the widget via alt+click the initial color does not match the current color.  That could be fixed by calling `setColorWidgetColor` as part of the alt+click handler.

src/segmentation_display_state/frontend.ts
```diff
@@ -48,19 +48,29 @@ import { isWithinSelectionPanel } from "#src/ui/selection_details.js";
...
 import { parseUint64 } from "#src/util/json.js";
 import { NullarySignal } from "#src/util/signal.js";
 import { withSharedVisibility } from "#src/visibility_priority/frontend.js";
+import { ColorWidget } from "#src/widget/color.js";
 import { makeCopyButton } from "#src/widget/copy_button.js";
 import { makeEyeButton } from "#src/widget/eye_button.js";
 import { makeFilterButton } from "#src/widget/filter_button.js";
 import { makeStarButton } from "#src/widget/star_button.js";
 
+declare const SEGMENT_LIST_COLOR_WIDGET: boolean | undefined;
```
jbms: This should be prefixed with NEUROGLANCER like all of the other build-time options to avoid (theoretical) name collisions, especially if neuroglancer is being embedded as a library.

src/widget/color.ts
```diff
@@ -23,22 +23,35 @@ import { vec3 } from "#src/util/geom.js";
 export class ColorWidget<
   Color extends vec3 | undefined = vec3,
 > extends RefCounted {
-  element = document.createElement("input");
+  static template() {
```
jbms: I think the changes in this file can be reverted now.

chrisj: oops... done!

PR #637, opened 2024-08-14
fix(http_requests): allow blob: to prepend URL string

xgui3783:
This is my first PR in this repo. Apologies if I have made any mistakes.

Currently, if the data source originates from the web (thus have the URL of `https?://`), the `parseUrl` function correctly parses the protcol, host and path. 

In our usecase, we often need to download the entire (nifti) image, preprocess them, and generate a blob URL. According to [mdn doc](https://developer.mozilla.org/en-US/docs/Web/API/URL#static_methods) on `URL.createObjectURL` , it creates a URL with `blob:` prepended. This fails the check as it is implemented currently. 

This PR addresses this issue.

### DIFF
--- a/src/util/http_request.spec.ts
+++ b/src/util/http_request.spec.ts
@@ -1,0 +1,57 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { describe, it, expect } from "vitest"
+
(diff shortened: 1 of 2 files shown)


xgui3783:
I signed it.

(Looks like I have initiated the CLA, but not signed it until a few seconds ago)

jbms:
Does this actually work, though?  It looks like it just discards the "blob:" prefix when parsing.

xgui3783:
The main issue is, without `^(blob:)?` the function `parseUrl` throw an error, and does not get passed on. 

Otherwise a nifti layer source such as `nifti://blob:https://localhost:8080/{blob-id}`

jbms:
The issue is that some code in Neuroglancer attempts to reassemble the URL from the result of calling `parseUrl` --- but you have modified `parseUrl` to just strip `blob:`, which means the URL would get reassembled as just `http://...` without the `blob:` prefix.  The nifti datasource probably doesn't do that, but in any case I think it would be better to include `blob:` in the protocol, i.e. make `blob:http` the protocol.  That should be possible with just a small change to your regular expression.  If you can also add a test, e.g. a Python test that uses webdriver apis to construct a blob URL and then loads it, that would be great.

xgui3783:
Apologies. I am a little overwhelmed at the moment. 

Seeing how the url parse mechanism changed in a big way lately, I feel this PR might no longer be needed. 

(PS very excited to see the `|` operation in data source. This opens the door to so much data sources, like unsharded precomputed chunks in zip)

PR #638, opened 2024-08-15
Rlbellaire patch 1

rlbellaire:
Mike - we're hoping you can accept/merge this pull request into Snapshot2.0 of NeuroTrALE-ui


fcollman:
@rlbellaire I think you chose the wrong target for this PR

PR #645, opened 2024-09-09
fix(virtual_lists) use existing debounceUpdateView with resize observer

pedson:
Use the existing `debounceUpdateView()` in place of just `updateView()` for the `resizeObserver()`, similar to other updates in the class.

When embedding Neuroglancer in an React app, the lack of debounce causes an infinite update loop and it will not load.


jbms:
Thanks for this fix.

Can you sign the CLA so that I can merge this?

pedson:
Is showing the green check mark for me now - let me know if it is still missing.

PR #650, opened 2024-10-01
feat(annotation) added annotation tag type properties and annotation iteration keybinds

chrisj:
Tags properties are currently represented as:
```json
"annotationProperties": [
  {
    "id": "tag0",
    "type": "uint8",
    "tag": "a"
  }
]
```

Annotation layers now have a tag tab which displays the current tag properties and allows additional ones to be added. A tool is created for each tag which can be bound to a keybind like any other tool. The tools toggle the associated tag's status for the current selected annotation.

This feature also bundles keybinds to iterating through the currently selected layer's annotations. To do this, I added the concept of layer events and sendEventToSelectedLayer.

Screenshot 2024-10-01 at 4 04 45 PM


chrisj:
@jbms any chance you could take a look at this while we re-think the timestamp PR?

jbms:
A few comments:

- why do tag properties have both an id and a tag, and why do they have a data type? 
- it may make sense to pack tags as a bitvector
- in general I think the "selected layer" concept in neuroglancer is kind of a mistake, since it inherently limits you to operating on a single layer at a time, and I've been trying to reduce its significance with the introduction of tool bindings and the ability to display more than one layer side panel at a time.  Can we instead make cycling through annotations a tool, e.g. shift+letter and then mouse wheel to cycle?

ceesem:
I know Chris will address these issues more technically, but from the feature perspective part of the goal was to separate the tag itself from the string value, in case you need to correct or update the string. I think that is the key reason they have both an id and a tag.

I like the idea of tool bindings that can be in a single side panel, but I think it's very important for usability that annotation iteration is a simple key binding alone.

chrisj:
>     * why do tag properties have both an id and a tag, and why do they have a data type?
I made the tag property type a valid numeric type as well so that is the only reason that tags have a data type with the only valid value being "uint8". I think this was purely done just with the order in which I developed the feature, introducing the separate type near the end but it might require some additional code break that association.

Like @ceesem One reason for having both id and tag is to simplify renaming tags, I'm trying to think if there was other reason but it avoids having to iterate through every annotation when a rename occurs.

>     * it may make sense to pack tags as a bitvector
Would it be possible to use bigint for the data type so we could have essentially an unlimited number of tags packed into one property?
>     * in general I think the "selected layer" concept in neuroglancer is kind of a mistake, since it inherently limits you to operating on a single layer at a time, and I've been trying to reduce its significance with the introduction of tool bindings and the ability to display more than one layer side panel at a time.  Can we instead make cycling through annotations a tool, e.g. shift+letter and then mouse wheel to cycle?
We have some expectations coming from our old fork of neuroglancer that was tuned for our proofreading style for certain actions to be optimized. I need to collect more feedback if something like that would be acceptable.

PR #653, opened 2024-10-21
feat: JPEG-XL Support (encoding: "jxl")

william-silversmith:
Hi Jeremy,

My lab has been exploring JPEG-XL for some time, and I finally got it working in my neuroglancer branch. I think it would be a good contribution to the ecosystem.

CloudVolume and Igneous both currently support JPEG-XL.

The implementation here uses the jxl-oxide implemention in Rust which I found easier to integrate than the C++ reference implementation.

Let me know what you think!

This was tested with both Grayscale and RGB color.


jbms:
Hi, can you add support for an `area` constraint in addition to the `width` and `height` constraints, as in: https://github.com/google/neuroglancer/commit/7ae9f648b73112045a0838103cf3cdf5afe4b3ac

william-silversmith:
sure, will do.

william-silversmith:
Ok! I think I set it up correctly.

jbms:
Can you fix the formatting (see updated CONTRIBUTING.md) and also fix the wasm reproducibility issue?

william-silversmith:
Seems like even though generating on MacOS (on the same machine) is consistent, ubuntu is doing something different. I'll have to try again on an Ubuntu machine.

william-silversmith:
@jbms I rebuilt the wasm on an Ubuntu machine. Please see if the reproducibility issue is fixed by running the check again.

william-silversmith:
Hi just wanted to check in and see if this is good to merge. <3 Thanks for all your hard work Jeremy. Once this merges, we'll start using JXL (lossless at the very least) on an industrial scale.

jbms:
@william-silversmith There seems to be a reproducibility issue with the build -- do you think you can take a look?  I tried pinning the docker container hash, which should be done anyway, but that does not seem to be sufficient.

william-silversmith:
It might be that I didn't save a Cargo.lock file so if a dependency changes, it could change the binary. What I can do is open a new PR with a lock file generated.

jbms:
> It might be that I didn't save a Cargo.lock file so if a dependency changes, it could change the binary. What I can do is open a new PR with a lock file generated.

You did include a Cargo.lock file, so I think it may be something else.

william-silversmith:
I see a Cargo.toml on github, but not a lock file in the repo. Am I missing it? Sorry!

jbms:
Oh, good point --- I thought I checked for the Cargo.lock myself but I must have misread it --- thanks.

I applied the fix myself in this PR:

PR #663, opened 2024-11-14
feat: option to hide cross section background in 3D slice views

seankmartin:
# Summary

This PR introduces a new global setting to optionally hide the background of 3D cross-sections in Neuroglancer. When activated, this setting allows users to view 3D slice projections without the background color, offering a less obstructed view of the data while still keeping the slice views in 3D on. Open to thoughts and suggestions on the functionality and implementation!

### Default (background visible):
image

### New Option (background hidden):
image

## Motivation

When displaying slice views in 3D, the background of 2D cross-sections can sometimes appear prominently, especially when the user is zoomed out. This view can be useful but may also limit visibility of the 3D content itself. This feature provides the flexibility to hide the background in 3D, to improve visibility without turning off the section rendering completely. 2D cross-section views remain unaffected.

## Implementation

The primary change is to the `SliceViewRenderHelper` class, and specifically the constructor and shader. The shader becomes a context and parameter dependent shader instead of a fixed shader. This allows the shader to change based on whether the user controlled setting is toggled on or off (the parameter to the memoized shader getter), and whether the render helper is for a cross-section view or projection view (the context).

When the hide setting is on, and the rendering is for a projection view, if the alpha of the sampled cross-section texture is 0 in the fragment shader, instead of picking up the background color, the fragment is discarded.

## Oblique slicing example

For clarity, here is an example with oblique slicing and a red projection background to illustrate the effects of this new setting.

image


jbms:
I put together this example showing how the shader can be additionally modified to make zero values (which in the fib25 dataset always indicate out-of-bounds) transparent, which in combination with this PR leads to a nicer visual effect:

[Example](https://neuroglancer--pr663-dlok77cl.web.app/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B3913.223876953125%2C3687.290771484375%2C4275.19775390625%5D%2C%22crossSectionScale%22:6.049647464412947%2C%22projectionOrientation%22:%5B-0.5151371359825134%2C0.2763257324695587%2C0.16025310754776%2C-0.7953594923019409%5D%2C%22projectionScale%22:14926.797212799056%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22precomputed://gs://neuroglancer-public-data/flyem_fib-25/image%22%2C%22tab%22:%22rendering%22%2C%22shader%22:%22#uicontrol%20invlerp%20normalized%5Cnvoid%20main%28%29%20%7B%5Cn%20%20if%20%28getDataValue%28%29%20==%20uint8_t%280u%29%29%20%7B%5Cn%20%20%20%20emitTransparent%28%29%3B%5Cn%20%20%20%20return%3B%5Cn%20%20%7D%5Cn%20%20emitGrayscale%28normalized%28%29%29%3B%5Cn%7D%5Cn%22%2C%22shaderControls%22:%7B%22normalized%22:%7B%22range%22:%5B79%2C180%5D%2C%22window%22:%5B54%2C205%5D%7D%7D%2C%22name%22:%22image%22%7D%5D%2C%22hideCrossSectionBackgroundIn3D%22:true%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22image%22%7D%2C%22crossSectionBackgroundColor%22:%22#000000%22%2C%22layout%22:%224panel%22%2C%22settingsPanel%22:%7B%22visible%22:true%7D%7D)

It may be reasonable to just enable this behavior all the time, rather than add an option for it.  I can't really think of a situation where displaying the out-of-bounds background is useful.

As a caveat, though, I will say that in general I find it not to be particularly useful to display the cross sections in the 3d views in actual usage --- mostly I have found the cross section display in the 3d view to be useful for screenshots, videos, and for demonstrating oblique slicing.  Others that actually rely on cross section display in the 3d view for interactive use may have a better idea as to whether this change would be undesirable in some cases.

seankmartin:
I've reversed the flag in this PR. So this means the default would be to hide areas of the cross section that are transparent in 3D.

It means that previous neuroglancer links would have this new behaviour to hide transparent pieces of the cross sections in 3D by default, but that the previous behaviour can be restored and saved in the state if desired

seankmartin:
Setting back to ready for review after fixing some issues in this PR with the most recent merge of changes from master branch

src/ui/default_input_event_bindings.ts
```diff
@@ -29,6 +29,7 @@ export function getDefaultGlobalBindings() {
     map.set("keyv", "toggle-default-annotations");
     map.set("keya", "toggle-axis-lines");
     map.set("keyo", "toggle-orthographic-projection");
+    map.set("keyt", "toggle-hide-cross-section-background-3d");
```
jbms, line 32: I think for now we could avoid reserving a key for this behavior and just leave it in the settings.

src/viewer.ts
```diff
@@ -469,6 +473,7 @@ export class Viewer extends RefCounted implements ViewerState {
   enableAdaptiveDownsampling = new TrackableBoolean(true, true);
   showScaleBar = new TrackableBoolean(true, true);
   showPerspectiveSliceViews = new TrackableBoolean(true, true);
+  hideCrossSectionBackgroundIn3D = new TrackableBoolean(true, true);
```
jbms, line 476: Upon further consideration, I think it is probably unwise to change the default value for existing states.  Instead we should figure out a general mechanism for changing the defaults in new states only, but for now I think it is better to keep it as you had it, and leave the default as false.

seankmartin: Yeah that makes a lot of sense, I think this is a good idea (the general mechanism for changing new states). I'll set this back to default false for now though, thanks!

seankmartin:
Hey @jbms, thank you for all the thoughtful input on this one,

I was wondering if the current version that keeps the default as showing the cross-section opaque background in 3-D views is ok, and then there is a setting to hide the background. Or do you have ideas on how we should we iterate upon this further?

jbms:
Thanks, this looks good --- as a final step can you add the option to python/neuroglancer/viewer_state.py and docs/json_schema/viewer_state.yml ?

PR #664, opened 2024-11-14
fix(layer): layer reference now reacts to layersChanged signal to unset reference when a layer is deleted

chrisj:
While testing the timestamp/cave annotation functionality, I came across this "issue" where deleting a segmentation layer that is linked from an annotation layer triggers this exception `TypeError: Cannot read properties of undefined (reading 'visibleSegmentEquivalencePolicy')`
Screenshot 2024-11-14 at 11 00 28 AM

Having LayerReference listen `layersChanged` instead of `specificationChanged` seems to make sense and works for both deleting a layer through the UI and through the state editor.

### DIFF
--- a/src/layer/index.ts
+++ b/src/layer/index.ts
@@ -1792,7 +1792,7 @@ export class LayerReference extends RefCounted implements Trackable {
     super();
     this.registerDisposer(layerManager);
     this.registerDisposer(
-      layerManager.specificationChanged.add(() => {
+      layerManager.layersChanged.add(() => {
         const { layer_ } = this;
         if (layer_ !== undefined) {
           if (!this.layerManager.layerSet.has(layer_) || !this.filter(layer_)) {


chrisj:
Also, this wasn't causing any obvious bugs in neuroglancer other than the error showing up in the console. It was only noticeable to me since I had "pause on uncaught exceptions" enabled.

Here is a link where you can reproduce it by deleting the first layer

https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B4e-9%2C%22m%22%5D%2C%22y%22:%5B4e-9%2C%22m%22%5D%2C%22z%22:%5B4e-8%2C%22m%22%5D%7D%2C%22position%22:%5B109421.8984375%2C41044.6796875%2C5417%5D%2C%22crossSectionScale%22:2.1875%2C%22projectionOrientation%22:%5B-0.08939177542924881%2C-0.9848012924194336%2C-0.07470247149467468%2C0.12882165610790253%5D%2C%22projectionScale%22:27773.019357116023%2C%22layers%22:%5B%7B%22type%22:%22segmentation%22%2C%22source%22:%22precomputed://gs://fafb-ffn1-20190805/segmentation%22%2C%22tab%22:%22rendering%22%2C%22segments%22:%5B%22710435991%22%5D%2C%22name%22:%22fafb-ffn1-20190805%22%7D%2C%7B%22type%22:%22annotation%22%2C%22source%22:%22precomputed://gs://neuroglancer-20191211_fafbv14_buhmann2019_li20190805%22%2C%22tab%22:%22rendering%22%2C%22annotationColor%22:%22#cecd11%22%2C%22shader%22:%22#uicontrol%20vec3%20preColor%20color%28default=%5C%22blue%5C%22%29%5Cn#uicontrol%20vec3%20postColor%20color%28default=%5C%22red%5C%22%29%5Cn#uicontrol%20float%20scorethr%20slider%28min=0%2C%20max=1000%29%5Cn#uicontrol%20int%20showautapse%20slider%28min=0%2C%20max=1%29%5Cn%5Cnvoid%20main%28%29%20%7B%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%20%20setEndpointMarkerColor%28%5Cn%20%20%20%20vec4%28preColor%2C%200.5%29%2C%5Cn%20%20%20%20vec4%28postColor%2C%200.5%29%29%3B%5Cn%20%20setEndpointMarkerSize%285.0%2C%205.0%29%3B%5Cn%20%20setLineWidth%282.0%29%3B%5Cn%20%20if%20%28int%28prop_autapse%28%29%29%20%3E%20showautapse%29%20discard%3B%5Cn%20%20if%20%28prop_score%28%29%3Cscorethr%29%20discard%3B%5Cn%7D%5Cn%5Cn%22%2C%22shaderControls%22:%7B%22scorethr%22:80%7D%2C%22linkedSegmentationLayer%22:%7B%22pre_segment%22:%22fafb-ffn1-20190805%22%2C%22post_segment%22:%22fafb-ffn1-20190805%22%7D%2C%22filterBySegmentation%22:%5B%22post_segment%22%2C%22pre_segment%22%5D%2C%22name%22:%22synapses_buhmann2019%22%7D%5D%2C%22showAxisLines%22:false%2C%22showSlices%22:false%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22synapses_buhmann2019%22%7D%2C%22layout%22:%22xy-3d%22%7D

jbms:
I think `specificationChanged` is correctly being triggered, it is just that it may be triggered later.  I think a better fix would be to simply delete this entire function (https://github.com/google/neuroglancer/blob/39272cdd398459bcbc03bf2f18affa9ebc2e340d/src/shared_disjoint_sets.ts#L70) so that accessing the object after its reference count becomes zero does not trigger exceptions.

I don't think "clearing" those fields had any significant impact on garbage collection anyway.

chrisj:
@jbms I forgot to say that from my testing at least `specificationChanged` does not get called when deleting a layer, I can work on some changes to make that happen. It looks like layerManager.specificationChanged is an aggregate of each individual `layer.specificationChanged` signal so probably that doesn't get fired when/before a layer is deleted.

Deleting disposed does fix this issue as well. I know I've run into this same type of error with disjoint sets in other places, probably a timing thing so any solution would be useful.

jbms:
Ensuring that `specificationChanged` is triggered when a layer is deleted would be a helpful change --- it was intended to trigger any time the JSON representation changes.

I think we should combine that fix with the `disposed` fix.

PR #666, opened 2024-11-16
wip:feat: adds Globus GCS-sourced assets as a datasource

jbottigliero:
This pull request adds integration with [Globus](https://www.globus.org/). Specifically, adding the ability to source data via HTTPS on a Globus Connect Server instance, authenticated using Globus Auth.

- The `credential_provider` was created based off of the `#src/util/google_oauth2.ts` implementation and other OAuth-like credential providers (e.g. `middleauth` and `ngauth`).

The Globus credential provider uses [PKCE](https://oauth.net/2/pkce/) for the authorization flow, which I've added a few basic utilities around [^1].

[^1]: These utilities could be replaced with an external library (e.g. [`pkce-challenge`](https://www.npmjs.com/package/pkce-challenge)) if that is preferred.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,8 +96,8 @@
   "type": "module",
   "exports": {
     ".": "./src/main_module.ts",
-    "./unstable/*.js": "./src/*.ts",
-    "./unstable/*": "./src/*"
+    "./*.js": "./src/*.ts",
+    "./*": "./src/*"
   },
   "imports": {
     "#src/third_party/jpgjs/jpg.js": "./src/third_party/jpgjs/jpg.js",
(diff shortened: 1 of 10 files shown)


joshmoore:
@jbottigliero: this sounds really interesting. Do you see coming back to it, or too much hassle?

jbottigliero:
Hey @joshmoore, the initial PR was a bit premature, but in a much better spot now.

It looks like I am unable to re-open this branch, so I've created a new one with the latest code: https://github.com/google/neuroglancer/pull/675 

(...which gives us a new number too! 😂)

PR #681, opened 2024-12-09
perf: faster jxl decoding

william-silversmith:
Hi Jeremy,

The current JXL decoding speed is kind of slow. It looks like rust jxl-oxide doesn't support simd or threads currently, but I was able to optimize the binary with `wasm-opt -O3` which cut the size of the wasm file from 2.0 MB to 1.7 MB and seems to speed up decoding by 19% on some test images.

It's still slow, but the speed up makes it somewhat more tolerable (~2.1s to ~1.7s). By comparison JPEG decodes in about 0.2s on these images.

As an aside, I heard from internet forums that Google Research might be producing a JXL rust decoder implementation for Firefox, but I don't know what the status is of that. I presume it'll be pretty fast when it's ready.

Will


william-silversmith:
I believe the test failures are unrelated to this PR (mypy is showing type errors in other areas of the codebase).

william-silversmith:
Merged in newest changes and it seems to pass CI now.

PR #688, opened 2024-12-19
Add option to hide/show code block

aranega:
## Summary
This PR introduces a new control button in the layer configuration panel that allows users to toggle the visibility of the shader code editor if one is present (depending on the layer's type). This button shows or hides the code editor within the layer configuration panel, allowing users to focus on other parameters present in the layer configuration panel. Open to feedback and suggestions on the implementation and user interaction!

## Current version (no code viewer button):
screenwithout

## New Option (code viewer visible with button, and hidden):
close
open

## Motivation
The ability to show or hide the code viewer was requested to help the user to not be intimidated by the shader code editor which is not manipulated often directly by user visualizing data only. Users often need the option to focus on the 3D view or data without the distraction of the shader code.

## User Interaction
In Neuroglancer, the new toggle button is available in the layer configuration panel, placed on top of the shader code editor, next to the "show large editor view" and the "documentation" button. This button allows users to show or hide the code viewer as needed. The control is simple, enabling users to easily manage the visibility of the code viewer without requiring additional keybinds. Very open to suggestions on improving this functionality and placement in the UI.

## Implementation
The primary change occurs within the tab rendering components for each impacted layers (currently, annotation layer, image layer, and single mesh layer). A button icon is dynamically updated based on the visibility state, ensuring users can easily toggle between viewing and hiding the code block. The state of the viewer is saved in the state related to each layer.

## Example of the new functionality:

https://github.com/user-attachments/assets/e2f09740-28b6-4ea3-88e3-f314c7f35c12


src/layer/index.ts
```diff
@@ -710,6 +710,7 @@ export class ManagedUserLayer extends RefCounted {
   }
 
   visible = true;
+  codeVisible = new TrackableBoolean(true, true);
```
jbms, line 713: I don't think there is a reason to make this a ManagedUserLayer property --- instead it should be a property of the specific UserLayer types that have a shader widget.

src/layer/image/index.ts
```diff
@@ -541,6 +552,22 @@ class RenderingOptionsTab extends Tab {
     topRow.className = "neuroglancer-image-dropdown-top-row";
     topRow.appendChild(document.createTextNode("Shader"));
     topRow.appendChild(spacer);
+
```
jbms, line 555: Can we factor out this topRow logic now that it has become more complicated to reduce duplication across layers?

jbms:
Note: The segmentation layer also has a shader code control, used when rendering skeletons only.  It would be good to add the option to hide that also.

src/layer/image/index.ts
aranega: I tried to factorize the top row creation, but at the end, it's tightly coupled to each reification of the UserLayer as the `codeVisible` property is located inside each `UserLayer` that requires it. Also, the top row is created by being associated at some point to the `ShaderOverlay` class that is defined in each module that defines the UserLayer. We could pass it as parameter, but we woud need also to have something for the help text, as well as the css class for the div. Also, having this as a separate function somewhere would imply that "layer" is passed as parameter and we couldn't access the "codeVisible" attribute unless we consider "layer" as any. If that's not a problem, I can factorize the code like this, and put this function in the same module that defines the ShaderCodeWidget.

At the moment, what I did is to extract the creation of the top row in a separate function, but duplicated (with small variations) in each module. This way, it mimicks more the equivalent of the  `makeShaderCodeWidget` function in each of those modules.

jbms: You could make the `layer` parameter some new interface that specifies the required properties --- and similarly pass in the ShaderOverlay and help strings as parameters.

aranega: Thanks a lot for the feedback. Sorry for the delay. I introduced a new function then that takes all it needs as parameter, defining new types when necessary.

jbms:
Thanks!

PR #690, opened 2024-12-21
Add missing event handler for position plot when rendered as tool in a palette

aranega:
## Summary

This PR fixes the dimension fine-tuning when the dimension tool is integrated in a palette. The way the dimension tool is working, when not integrated in a palette, is by grabbing the events from the x/y/z input widgets. When the dimension tool is moved in a palette, only those input widget are reacting to the events from the mouse wheel. This patch adds the mouse event handling also on the `PositionPlot` instance.

### DIFF
--- a/src/widget/position_widget.ts
+++ b/src/widget/position_widget.ts
@@ -1566,20 +1566,7 @@ class DimensionTool<Viewer extends object> extends Tool<Viewer> {
       ).element,
     );
 
-    const shouldIgnoreEvent = (event: Event) => {
-      const target = event.target;
-      if (
-        target instanceof Element &&
-        target.matches(".neuroglancer-position-dimension-playback *")
-      ) {
-        return true;
-      }
-      return false;
-    };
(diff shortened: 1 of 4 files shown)


src/widget/position_widget.ts
```diff
@@ -1557,6 +1557,36 @@ class DimensionTool<Viewer extends object> extends Tool<Viewer> {
         }),
       ).element,
     );
+
+    this.registerDisposer(new MouseEventBinder(plot.element, inputEventMap));
+
+    registerActionListener<WheelEvent>(
```
jbms, line 1563: Can you factor out these handlers rather than duplicating them?

aranega: @jbms I factorized the code in two dedicated private methods. The event action is passed as parameter of those methods. I can remove this logic from the methods and place it back inside the closure that calls the private methods.

jbms:
Thanks!

PR #692, opened 2025-01-08
Add annotation color as a tool in the layer's render tab

aranega:
## Summary
This PR moves the widget for color management for local annotation from the annotation tab to the rendering tab, improving UI organization and aligning widget placement with its functionality. The local annotation default color widget is also now part of a `Tool` so that it can be used in the new tool palettes.

Additionally, this PR fixes an issue in the `removeJsonString` function to handle tool keys, resolving inconsistencies caused by differences in the `toJSON()` method across `Tool` and `LegacyTool` classes.

## User Interaction

Widget Relocation: The widget previously located in the state tab is now found in the rendering tab as a tool. This change simplifies navigation for users working with segment-related settings.
shadercolor

To notify the user that the default color widget moved to the "rendering tab", a new icon with a question mark is displayed in the local-annotation layer, next to the buttons that allows to create new annotations, with a tooltip stating what are the buttons used for, and where is located the color control.

## Implementation

Widget Relocation:
Code Changes: Updates in `layer/annotation/index.ts`  moved the initialization and rendering logic of the widget to the rendering tab.

`removeJsonString` Fix:
Problem: The `toJSON()` method for `Tool` and `LegacyTool` behaves differently. For `LayerTool` (inheriting from `Tool`), `toJSON()` returns a `jsonId` value, causing keys in `localToolBinder` to include quotes (e.g., `"mykey"` instead of `mykey`). This created mismatches when using `removeJsonString()`, as it required explicit `JSON.stringify()` calls.
Solution: The fix first attempts to locate the key directly. If it is not found, it falls back to checking the `JSON.stringify()` version. This approach ensures compatibility with both `Tool` and `LegacyTool` implementations.


src/layer/annotation/index.ts
```diff
@@ -455,6 +465,9 @@ export class AnnotationUserLayer extends Base {
     this.annotationDisplayState.shaderControls.restoreState(
       specification[SHADER_CONTROLS_JSON_KEY],
     );
+    this.annotationDisplayState.color.restoreState(
```
jbms, line 468: This is already restored by UserLayerWithAnnotationsMixin which is a base class.

src/layer/annotation/index.ts
```diff
@@ -709,6 +722,7 @@ export class AnnotationUserLayer extends Base {
     x[SHADER_JSON_KEY] = this.annotationDisplayState.shader.toJSON();
     x[SHADER_CONTROLS_JSON_KEY] =
       this.annotationDisplayState.shaderControls.toJSON();
+    x[ANNOTATION_COLOR_JSON_KEY] = this.annotationDisplayState.color.toJSON();
```
jbms, line 725: as above, this is redundant with the base class

src/layer/annotation/index.ts
```diff
@@ -812,9 +828,53 @@ class RenderingOptionsTab extends Tab {
...
+const LAYER_CONTROLS: Record<
+  string,
+  LayerControlDefinition<AnnotationUserLayer>
+> = {
+  [ANNOTATION_COLOR_JSON_KEY]: {
+    label: "Annotation color",
+    title:
+      "Annotation shader default color (enabled with 'defaultColor()' in the shader code)",
+    toolJson: ANNOTATION_COLOR_JSON_KEY,
+    ...colorLayerControl(
```
jbms, line 868: The isValid property can be specified here instead of the `hasDefaultColor` logic above.

jbms:
Thanks for your work on this.

For AnnotationUserLayer this seems like a good change.  However, note that other layers also have an Annotations tab due to UserLayerWithAnnotationsMixin, which normally just shows the overall bounds of the data for that layer.  For those layers, if we remove the color selector from the Annotations tab then we would need to put it somewhere else --- currently you have added it to the Rendering tab for AnnotationUserLayer only, but not to the Rendering tab of other layers that inherit from UserLayerWithAnnotationsMixin.

We could either leave the color selector on the Annotations tab for non-AnnotationUserLayer layers, or could move it to the Rendering tab on those layers.  Not sure what the best option is.  It may be slightly undesirable to clutter the rendering tab on e.g. ImageUserLayer and SegmentationUserLayer, which already have many controls, with something that most users won't care about.

src/ui/tool.ts
```diff
@@ -531,7 +531,9 @@ export class LocalToolBinder<
   }
 
   removeJsonString(toolJsonString: string) {
-    const key = this.jsonToKey.get(toolJsonString);
+    const key =
```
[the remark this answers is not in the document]
aranega: Ah, ok, got it. Yes, it was to deal with the `removeJsonString(ANNOTATION_COLOR_JSON_KEY)`, it was just really weird as an API user to have to deal with the JSON.stringify myself calling the method. I'll change that then.

aranega:
What we could do is to add a special static information on the UserLayer and overwrite it for the AnnotationUserLayer, this would be the main modification in the constructor of the `AnnotationLayerView` class:

```typescript
const annotationColorPickerEnabled = (layer.constructor as typeof UserLayer).supportColorPickerInAnnotationTab;
if (annotationColorPickerEnabled) {
  const colorPicker = this.registerDisposer(
    new ColorWidget(this.displayState.color),
  );
  colorPicker.element.title = "Change annotation display color";
  this.registerDisposer(
    new ElementVisibilityFromTrackableBoolean(
      makeCachedLazyDerivedWatchableValue(
        (shader) => shader.match(/\bdefaultColor\b/) !== null,
        displayState.shaderControls.processedFragmentMain,
      ),
      colorPicker.element,
    ),
  );
  toolbox.appendChild(colorPicker.element);
}
```

And the only two other modifications would be in the UserLayer to have `static annotationColorPickerEnabled` set to `true` by default and `false` in `AnnotationUserLayer`. This would mimick the way `supportPickOption` is handled accross the differnt layer types.

jbms:
Sounds good --- you can add the property to UserLayerWithAnnotations / UserLayerWithAnnotationsMixin

aranega:
Thanks for the review, and the modficiations suggestions! Here is the new version

jbms:
Thanks!

PR #694, opened 2025-01-11
feat(annotation) Add default annotation properties

chrisj:
add line length property

This came up during a meeting that the scale bar has been used as a rough measure of the size of features in datasets. I was thinking we should have a better way to measure distance and the line annotation is already a tool that has most of the requirements.

I think it would also make sense to add additional properties for other types such as ellipsoid volume and semi axis lengths, bounding box volume and dimensions.

I am currently making assumptions to cast float32array to a vec3 and there might be a better method of using the globalCoordinateSpace.

### DIFF
--- a/src/annotation/index.ts
+++ b/src/annotation/index.ts
@@ -18,7 +18,6 @@
  * @file Basic annotation data structures.
  */
 
-import { vec3 } from "gl-matrix";
 import type {
   BoundingBox,
   CoordinateSpaceTransform,
@@ -703,7 +702,7 @@ export interface AnnotationTypeHandler<T extends Annotation = Annotation> {
   ) => void;
   defaultProperties: (
     annotation: T,
-    scales: vec3,
+    scales: Float64Array,
   ) => {
     properties: AnnotationNumericPropertySpec[];
     values: number[];
@@ -827,24 +826,32 @@ export const annotationTypeHandlers: Record<
       callback(annotation.pointA, false);
       callback(annotation.pointB, false);
     },
-    defaultProperties(annotation: Line, scales: vec3) {
-      return {
-        properties: [
-          {
-            type: "float32",
-            identifier: "Length",
-            default: 0,
-            description: "Length of the line annotation in nanometers",
-            format: formatLength,
-          },
-        ],
-        values: [
-          vec3.dist(
-            vec3.mul(vec3.create(), scales, annotation.pointA as vec3),
-            vec3.mul(vec3.create(), scales, annotation.pointB as vec3),
-          ),
-        ],
-      };
+    defaultProperties(annotation: Line, scales: Float64Array) {
+      const properties: AnnotationNumericPropertySpec[] = [];
+      const values: number[] = [];
+      const rank = scales.length;
+      if (
+        rank === annotation.pointA.length &&
+        rank === annotation.pointB.length
+      ) {
+        properties.push({
+          type: "float32",
+          identifier: "Length",
+          default: 0,
+          description: "Length of the line annotation in nanometers",
+          format: formatLength,
+        });
+        let length = 0;
+        for (let dim = 0; dim < rank; dim++) {
+          length +=
+            (((annotation.pointA[dim] - annotation.pointB[dim]) / 1e-9) *
+              scales[dim]) **
+            2;
+        }
+        length = Math.sqrt(length);
+        values.push(length);
+      }
+      return { properties, values };
     },
   },
   [AnnotationType.POINT]: {
@@ -890,7 +897,7 @@ export const annotationTypeHandlers: Record<
     visitGeometry(annotation: Point, callback) {
       callback(annotation.point, false);
     },
-    defaultProperties(annotation: Point, scales: vec3) {
+    defaultProperties(annotation: Point, scales: Float64Array) {
       annotation;
       scales;
       return { properties: [], values: [] };
@@ -963,7 +970,10 @@ export const annotationTypeHandlers: Record<
       callback(annotation.pointA, false);
       callback(annotation.pointB, false);
     },
-    defaultProperties(annotation: AxisAlignedBoundingBox, scales: vec3) {
+    defaultProperties(
+      annotation: AxisAlignedBoundingBox,
+      scales: Float64Array,
+    ) {
       annotation;
       scales;
       return { properties: [], values: [] };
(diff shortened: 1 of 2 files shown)


jbms:
Overall I like this idea, thanks!

Originally some default properties like length, volume, etc. were shown for annotations but that support was left out when I refactored some years ago to support more than 3 dimensions.

The logic related to scales will need to change ---- the coordinate space of the annotation is not necessarily 3d and doesn't necessarily directly correspond to the global coordinate space, so some coordinate transforms will need to happen, and units taken into account also.

src/ui/annotations.ts
```diff
@@ -1854,6 +1854,19 @@ export function UserLayerWithAnnotationsMixin<
 
                 const { relationships, properties } = annotationLayer.source;
                 const sourceReadonly = annotationLayer.source.readonly;
+                const globalCoordinateSpace =
+                  this.manager.root.coordinateSpace.value;
+                const defaultProperties = annotationTypeHandlers[
+                  annotation.type
+                ].defaultProperties(annotation, globalCoordinateSpace.scales);
```
jbms, line 1861: This is still not correct --- you need to transform the scales, taking into account the various affine transforms that apply, similar to what is done here:

https://github.com/google/neuroglancer/blob/5898236dbc127b46ea280fc75c9e37bcee03a49b/src/ui/annotations.ts#L1073

src/annotation/index.ts
```diff
@@ -814,6 +826,33 @@ export const annotationTypeHandlers: Record<
...
+          type: "float32",
+          identifier: "Length",
+          default: 0,
+          description: "Length of the line annotation in nanometers",
+          format: formatLength,
+        });
+        let length = 0;
+        for (let dim = 0; dim < rank; dim++) {
+          length +=
+            (((annotation.pointA[dim] - annotation.pointB[dim]) / 1e-9) *
```
jbms, line 847: This shouldn't always be nanometers --- it should take into account the units set in the coordinate space.

If all dimensions have the same base unit, e.g. meter, then you could just choose an appropriate scale automatically as in the scale bar.

If there are mixed units, e.g. meters and seconds, then probably just don't display the length at all.

chrisj: I updated it to use globalCoordinateSpace.units and using `ALLOWED_UNITS` from scale_bar.ts to disable the property. If any unit is a non-length unit, it won't add the property. Also cleaned up the code

jbms:
The points stored in the annotation are in the "model" coordinate space.  There is an affine transform that maps between that coordinate space and the concatenation of the layer-local and global coordinate spaces. 

The lengths could be computed either in the "model" coordinate space or in the global coordinate space,.or maybe both, but right now you are using the untransformed points in the model coordinate space with the global coordinate space scales, which is not correct.

chrisj:
Hi Jeremy, I am now passing in the positions from `visitTransformedAnnotationGeometry` to the defaultAnnotationProperties function. I think I could clean it up a bit if `visitTransformedAnnotationGeometry` was changed from being a function that uses callbacks to instead returning an array of [layerPosition, isVector].

Does this fix the coordinate space issue? It  I'm still using global scales/units. `this.localCoordinateSpace.value.scales`, `this.localCoordinateSpace.value.units` show empty. When would it have a value?

Here is an example where i can see the length increase when I give the x dimension a 10x scale.
Screenshot 2025-01-21 at 4 34 02 PM

src/annotation/index.ts
```diff
@@ -751,6 +766,32 @@ function deserializeTwoFloatVectors(
...
+  }
+  const [pointA, pointB] = annotationLayerPositions;
+  const scalesRank = scales.length;
+  const lineRank = pointA.length;
+  if (scalesRank < lineRank) {
+    return;
+  }
+  let lengthSquared = 0;
+  for (let dim = 0; dim < lineRank; dim++) {
+    const unitInfo = ALLOWED_UNITS.find((x) => x.unit === units[dim]);
```
jbms, line 785: `units` will always be normalized to "m" if it is some multiple of meters, so there is no need to check ALLOWED_UNITS.

Note that ALLOWED_UNITS is actually from some now-dead code that was originally used to show annotation properties just like you are adding with this PR.

However, you may as well support any unit for length --- just as long as all units for which there is a non-zero delta are the same.

Or perhaps it could be represented as  X meters + Y seconds, if for example there are both meters and seconds units.  You would compute the length separately for each unit.

The actual SI prefix can be selected using pickSiPrefix, as for the scale bar.

src/annotation/index.ts
```diff
@@ -814,6 +855,28 @@ export const annotationTypeHandlers: Record<
...
+      annotation;
+      const properties: AnnotationNumericPropertySpec[] = [];
+      const values: number[] = [];
+      const length = lineLength(annotationLayerPositions, scales, units);
+      if (length) {
+        properties.push({
+          type: "float32",
+          identifier: "Length",
+          default: 0,
+          description: "Length of the line annotation in nanometers",
```
jbms, line 873: Should probably just say "length of the line annotation", with a type of string, and then the value can include the appropriate unit suffix.

jbms:
localCoordinateSpace will only be non-empty if a dimension has been renamed to e.g. `z'` to make it a local dimension.

Using visitTransformedAnnotationGeometry is good, but note that you need to take some care in using the result --- you should follow the pattern here (https://github.com/google/neuroglancer/blob/6f9335f1f2560d67f7c40378206414a7493c0c73/src/ui/annotations.ts#L904) and use `globalToRenderLayerDimensions` and `localToRenderLayerDimensions`.

src/annotation/index.ts
chrisj: Hi Jeremy, getting back to this now, are you suggesting adding "string" as a type of `AnnotationNumericPropertySpec` or Adding a `AnnotationStringPropertySpec`. I thought there was no string currently for performance reasons.

jbms: Oh, I forgot that string properties aren't supported for annotations --- only for segment properties.  They would be challenging to support within a shader, and some changes to the precomputed annotation encoding would be needed to support strings in that format, but there is no fundamental difficulty with supporting them solely for display in the "Selection details" panel.  However, I realize that with the `format` member you added to AnnotationNumericPropertySpec we can anyway accomplish the same thing without adding string annotation support.

chrisj: In that case, would you suggest skipping the "X meters + Y seconds" idea and instead only calculate length if all the non-zeros are the same unit?

stuarteberg:
This looks like a very useful idea!  Since this PR is now closed, I thought I'd ask if, by any chance, this feature happens to be under development in any other PR?

PR #703, opened 2025-01-27
Remove EventAction import to resolve dev-server compilation error

aaronkanzer:
`EventAction` is an unused import -- in which type-checking fails when running `npm run dev-server` locally -- simple PR to remove import until it is either used or intentionally removed

This issue persisted when trying on [node >= 22](https://github.com/google/neuroglancer/blob/master/package.json#L10-L12) (verified locally via `nvm use 22` and `nvm use 23`)

Screenshot 2025-01-27 at 2 23 16 PM

Cc @kabilar


aaronkanzer:
@aranega -- I think this error may have been introduced via this commit from 4 days ago: https://github.com/google/neuroglancer/commit/6f9335f1f2560d67f7c40378206414a7493c0c73

Issue #665, opened 2024-11-15
Suggestion: Alternative 4-Panel Layout


wanqingy:
Hi! I would like to propose an alternative 4-panel layout in Neuroglancer that may better facilitate cross-referencing three orthogonal planes alongside a 3D view.

**Suggested configuration:**
Top left: xy
Bottom left: xz view
Top right: yz view
Bottom right: 3d

**Examples:** 
[...]

PR #725, opened 2025-02-11
feat(data_panel_layout): add 4panel-alt layout

jbms:
Fixes https://github.com/google/neuroglancer/issues/665.

### DIFF
--- a/docs/json_schema/viewer_state.yml
+++ b/docs/json_schema/viewer_state.yml
@@ -55,7 +55,7 @@ properties:
     default: true
   showSlices:
     type: boolean
-    title: "Indicates whether to show cross sections in the 3-d view of `~DataPanelLayoutType.4panel` layouts."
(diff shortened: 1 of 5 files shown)


stuarteberg:
Nice feature; probably more intuitive to most newcomers.  But for those of us whose muscle memory was imprinted on neuroglancer...

Is there any way to enable the old `4panel` layout by default even when toggling between single-panel and 4-panel views repeatedly?

I think that's what the [following suggestion](https://github.com/google/neuroglancer/issues/665#issuecomment-2494982263) was referring to, but I don't know if it was implemented:

>If you wanted to ensure old states continue to use the old layout even when swapping between layouts, you could have some [indicator] in newly generated (blank) states indicating this is a new epoch.

For example, if I navigate to [this link](https://neuroglancer-demo.appspot.com/#!gs://flyem-user-links/short/example-manc.json) (which uses the old 4-panel layout) and then hit spacebar twice, it does not bring me back to the same view I started with.  It takes switches to the new 4-panel layout.

(This is not a big deal, but I thought I'd ask.)

Issue #570, opened 2024-04-05
Changing layer type is easily destructive


ceesem:
It is possible to change a layer type in the UI after it is already defined and in use. Since the layer type is immediately next to the various tool tabs, it is particularly easy to hit in tools like a graphene segmentation layer, where one often needs to move between different tabs to do work. If you accidentally change the layer type on such a misclick, all of the current settings are reset, which can destroy work. For example, accidentally changing the layer type of a segmentation layer removes all selected IDs and resets the various render settings like opacity.

In general, operation that destroy user work should probably be clearly deliberate and hard to trigger accidentally. I can imagine two solutions for this case, which are non-exclusive:
1) Have a confirmation dialogue for rare-but-destructive changes like changing the type for a non-fresh layer.
2) Implement a state-level "Undo," as was done in the older Seung-lab neuroglancer branch.

jbms:
I think state-level undo is the best solution.

PR #726, opened 2025-02-12
feat(layer-list-panel): add layer type tags to the list panel and modify layer list panel styling

seankmartin:
This PR proposes adding a tag indicating the type of the layer to the layer list panel. The primary changes in this PR are to add a widget that provides a read only tag for the layer type and some styling changes on the layer list panel.

The motivation for this is to give a quick visual indicator of the layer types. It is proposed to add to the layer list panel instead of the top bar, because it would likely make the top bar quite crowded. But the layer list panel has the space to support such a visual indicator.

Proposed in this PR:
Screenshot from 2025-02-12 12-44-01

Current:
Screenshot from 2025-02-12 12-44-05

## More details
The layer type indicator has a fixed size to ensure that the alignment in the layer list panel is uniform (this is why msh is used over mesh as an abbreviation, unlike in the `LayerTypeWidget`).

Another possibly would of course be to use the `LayerTypeWidget` directly, perhaps with a little bit of styling specifically when it is in the layer list panel instead of the new `LayerTypeIndicatorWidget` created here. The main potential drawback of that is that intention of this is to give a quick visual indication of all the layer types. And as layer types are not changed that often once created, perhaps it is better to obfuscate that away from the layer list panel (e.g. the concern in https://github.com/google/neuroglancer/issues/570 -- perhaps if state level undo becomes available the `LayerTypeWidget` could be used instead).

The styling changes in this PR are intended to help improve the alignment of some items and allow space for the new tag. The background color behind the layer number is also proposed to be removed here, but happy to add that back if the background color has a strong meaning.


src/widget/layer_type_indicator.ts
```diff
@@ -0,0 +1,48 @@
...
+export class LayerTypeIndicatorWidget extends RefCounted {
+  element = document.createElement("div");
+  typeToAbbreviation = new Map<string, string>();
+  constructor(public layer: ManagedUserLayer) {
+    super();
+    this.element.classList.add("neuroglancer-layer-type-indicator");
+    for (const [layerType, layerConstructor] of layerTypes) {
+      this.typeToAbbreviation.set(
+        layerType,
+        layerConstructor.typeAbbreviation === "mesh"
```
jbms, line 34: we can just change the `typeAbbreviation` to `msg` to simplify this.

seankmartin: Sure! Thanks @jbms, I changed the mesh `typeAbbreviation` to `msh`

PR #728, opened 2025-02-12
fix(graphene): fix breakage due to kvstore url encoding fixes

jbms:
Fixes https://github.com/google/neuroglancer/pull/727.


chrisj:
Thanks, seems to be working well. We will add tests, Sven has been discussing that.

I had a feeling my change was naive but figured might as well submit something in case it was that simple.

PR #736, opened 2025-02-20
Feat(display_dimensions_widget): FOV input on 2D panel dim widget and make inputs clearer

seankmartin:
For 2D panels only, an FOV input is now available. This FOV input is available always for isomorphic display units. But if the units are not isomorphic in display, the FOV input is only available if the slice is axis aligned and the normal of the slice is parallel to the default slice normal.

In the display dimensions widget, the scale factor and selected dimensions were inputs, but they kind of looked static. These have been changed to be input boxes
image

The scale itself was reported, but was not editable. This has been changed to be editable (both in 2d and 3d panels), and links into the zoom level:
image

We've also got some suggested restyling in this to improve alignment of items in the display dimensions widget, happy to iterate on that. There is also a new tooltip in the screenshot menu to point to this widget, and the scaleFactor in the display dimensions widget is suggested to change in display here from being in `VALUEpEXPONENT` format below `0.01` instead of below `1.0`


jbms:
This works very nicely!

A few minor comments:

- "/px" should maybe be excluded from the input field
- input element attributes need to be set to prevent the spellcheck squiggles

src/widget/display_dimensions_widget.ts
```diff
@@ -240,6 +348,71 @@ export class DisplayDimensionsWidget extends RefCounted {
...
+    const currentQuaternion = this.orientation.orientation;
+    if (quat.equals(defaultQuaternion, currentQuaternion)) return true;
+
+    const currentNormal = vec3.create();
+    const defaultNormal = AXES_NORMALS.get(this.axes);
+    if (defaultNormal === undefined) return false;
+    vec3.transformQuat(currentNormal, [0, 0, 1], currentQuaternion);
+    const epsilon = 1e-7;
+    return (
+      Math.abs(1 - Math.abs(vec3.dot(currentNormal, defaultNormal))) <= epsilon
```
jbms, line 398: Rather than checking if it is equal to the default orientation, it would be better to check if it is aligned to some axes --- note that using the "z" binding the user can easily snap the view to a different axis-aligned orientation.  This could potentially be checked by converting the quaternion to a rotation matrix and examining the coefficients, or maybe from the projection matrix directly.  In fact we don't need to special case isUniformDisplayScale at all --- instead we can just look at the coefficients that are not (approximately) zero, and based on that determine which display dimensions contribute to the viewport width and height directions.  If those display dimensions all have uniform units and scale, then we can show the viewport fov.

seankmartin: Thanks Jeremy, this makes sense. Working on it at https://github.com/MetaCell/neuroglancer/tree/feature/fov-improved

Two slight differences:
1. At the moment I'm using the normal vector of the plane instead of the rotation matrix directly to check which axes are aligned. I can try replace that later perhaps, but as I was previously doing a normal vector check it was a bit easier to just extend that.
2. If it is axisAligned then we don't need to have uniform units and scale so I think we might still need the special case in a way.

Either way this is a big improvement because the slices don't have to align with their original axis, they can align with any axis. And it also means we should be able to remove the awkward thing originally introduced in this PR where panels had to tell the widget they were an "xy" panel, which I never really liked.

Will keep at it and merge that temp branch into here when ready, thanks again

jbms: The reason I suggested rotation or projection matrix is that you can just directly read out the locations of the non-zero coefficients to determine which input dimensions contribute to the output dimensions.  So for the "width" dimension you can read out which of the display dimensions it depends on, i.e. for which there is a non-zero coefficient in the rotation/projection matrix, and then check if those display dimension on which the "width" dimension depends all have the same scale and unit.  If they do, then you can compute a "width" field of view, otherwise not.  The same can be done independently for the "height" dimension.  A rotation vector makes this harder because the rotation angle does impact the result if only 2 out of the 3 display dimensions have the same scale and units.

seankmartin: Thanks you for clarifying this Jeremy, I understood the intention better this time around.

I've used the rotation matrix idea, which hopefully makes this cleaner, appreciate the suggestion!

src/util/geom.ts
```diff
@@ -478,3 +482,45 @@ export function getViewFrustrumWorldBounds(
...
+  tolerance: number = 1e-6,
+): OrientedSliceScales | null {
+  function extractContributingScales(
+    matrixRow: 0 | 1 | 2,
+  ): { scale: number; unit: string } | null {
+    const indices = [0, 1, 2].map((i) => i + 3 * matrixRow);
+    const contributingScales: number[] = [];
+    const contributingUnits: string[] = [];
+
+    for (const index of indices) {
```
jbms, line 499:
```
for (let i = 0; i < 3; ++i) {
  let index = i + 3 * matrixRow;
  ...
}
```

Rather than use an array / set, you can just use:

```
let contributingScale: number | undefined;
let contributingUnit: number | undefined;
```

and then check if the current unit / scale matche the previous one if not undefined.

For units the exact comparison you have makes sense.  For the scales it would probably be better to use an approximate comparison, i.e. `Math.abs(a - b) / Math.max(a, b) < 1e-6` in case the scales are not exactly equal due to some numerical issue (not sure how likely that is, but may as well be defensive).

PR #740, opened 2025-02-21
refactor(uint64): replace custom Uint64 class with bigint

jbms:
Now that there is support in the major browsers for bigint and optimization of the uint64 as bigint case, the custom Uint64 class can be removed.


jbms:
@chrisj If you are able to test that this doesn't break graphene, that would be much appreciated!

chrisj:
@jbms I'll test it now

chrisj:
@jbms looks good to me. There are some regressions introduced with some previous pull requests that I am looking into now and I will put up an issue for soon but I don't see any problems related to this PR. Happy to see this change!

jbms:
Thanks for testing!

PR #748, opened 2025-03-05
fix(UI - layer_list_panel): Change layer number width spacing to consider archived layers

seankmartin:
The layer list panel number width is determined by the number of characters in the highest number non-archived layer. Since archived layers didn't contribute to the layer count for determining the width for the layer list panel numbers, it could result in cases like this where the number would overlap the visibility icon for archived layers:

image

We're thinking to change the count here to take all layers into account so that archived layer numbers don't overlap the visibility icon and be like this:

image

Maybe there is some reason though that archived layers were previously excluded for the number width count that we should be considering here? For example, maybe the layer numbers were intended to be hidden for all archived layers. If so, happy to amend this to account for that, thanks

### DIFF
--- a/src/ui/layer_list_panel.css
+++ b/src/ui/layer_list_panel.css
@@ -67,8 +67,6 @@
   display: inline-block;
   text-align: left;
   line-height: 18px;
-  width: auto !important;
-  min-width: 2ch;
 }
 
 .neuroglancer-layer-list-panel-item:not(:hover)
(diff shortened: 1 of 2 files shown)


seankmartin:
Changed the check now for hiding numbers of archived layers in the layer_list_panel.

Could be missing again some reason for the old width calculation, but updated that too to directly use the number of characters in the string representation of the number of non-archived layers, not the number of non-archived layers + 1

jbms:
Thanks

Yeah, the `+1` was a bug I guess --- I added that because it is numbered from 1, but that is already taken into account...

PR #750, opened 2025-03-11
feat: show layer on unarchive layer

seankmartin:
We've been using the archive feature for layers from the side bar, and thought it might be a bit more intuitive if unarchiving the layer also showed the layer.

I think it could be pretty common that a layer is unarchived because you want to see that layer. But very open to thoughts on this behaviour. Thanks!


jbms:
I imagine it is true that you would typically want to make it visible, but that can already be accomplished in one click using the visibility button next to the check box.

seankmartin:
I'll close this one since the visibility button allows the intended behaviour, thanks Jeremy

PR #757, opened 2025-03-17
adds aws_sigv4 signer for s3 requests, adds python s3 credentials

MatthewAitken:
- Adds aws sigv4 signer for fetch requests against private s3 buckets
- Implements neuroglancer python s3 credential provider
- Updates gcs kvstore to use python gcs credential provider
- Updates s3 kvstore to use python s3 credential provider

Todo:
Implement ngauth server


fcollman:
@jbms Matt is working on figuring out what the CLA signing situation is in general at the Allen, but we would appreciate any feedback you have on this sooner rather than later.

jbms:
In general this looks fine --- I don't think there are many issues.

MatthewAitken:
> Hi @MatthewAitken, thanks for implementing this. May I ask you for an MWE to use this?
> 
> Here is what I tried, but I still get some permission error in the client:
> 
> `AWS_PROFILE="my-profile" ipython`, which handles passing specific credentials to a boto3 session that only reads the default ones.
> 
> Then in ipython
> 
> ```
> import neuroglancer
> v = neuroglancer.Viewer()
> with v.txn() as state:
>     state.layers["test"] = neuroglancer.ImageLayer(source="s3://bucket/example.zarr/")
> print(v)
> ```

Apologies for the delay getting back to finish this.

You need to allow credentials (and replace mytoken with whatever token string you want)
`viewer = neuroglancer.Viewer(allow_credentials=True, token="mytoken")`

src/credentials_provider/aws_sigv4.ts
```diff
@@ -0,0 +1,126 @@
...
+      sessionToken: credentials.token,
+    },
+    sha256: Sha256,
+  });
+  const apiUrl = new URL(input.toString());
+  const request = {
+    hostname: apiUrl.hostname.toString(),
+    protocol: apiUrl.protocol,
+    path: apiUrl.pathname,
+    method: "GET",
```
ebezzi:
```suggestion
    method: init.method || "GET",
```

@MatthewAitken I was helping @bentaculum with [his issue](https://github.com/google/neuroglancer/pull/757#issuecomment-2932782515) and I was able to get it working with this fix. The reason is that we were getting 403 for all the HEAD requests.

MatthewAitken: thanks for that! @ebezzi will you sign the google cla? https://cla.developers.google.com/

PR #765, opened 2025-03-28
modifying annotation schema to match real world examples

fcollman:
I was attempting to build a set of helper functions to facilitate reading precomputed annotations using tensorstore, and I thought i'd utilize the jsonschemas in this repo to facilitate validation. 

In starting to use them on a real world example such as "gs://neuroglancer-20191211_fafbv14_buhmann2019_li20190805" 

i discovered this failed validation for a few reasons, some of which were clear errors, and others that were perhaps oversights or inconsistencies between neuroglancer and the written spec.  I've modified the spec so that this info file passes validation in a way that I think makes sense but feedback is welcome.

### DIFF
--- a/src/datasource/precomputed/json_schema/annotation.yml
+++ b/src/datasource/precomputed/json_schema/annotation.yml
@@ -39,13 +39,9 @@ properties:
       be contained with the bounding box defined by `.lower_bound` and
       `.upper_bound`.
   annotation_type:
-    title: |-
-      Annotation geometry type.
-    enum:
-      - "point"
-      - "line"
-      - "axis_aligned_bounding_box"
-      - "ellipsoid"
+    title: Annotation geometry type.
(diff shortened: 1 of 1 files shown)


jbms:
Thanks for bringing up these issues.

One issue is that the schemas are designed to serve both for validation but also for documentation.

The precomputed format generally allows "enums" to be case insensitive mostly because I was lazy when initially adding the support but I kind of wanted to discourage that going forward by only documenting them as lowercase.

Currently the json schema documentation generation in the sphinx-immaterial theme does not handle regexp patterns at all:

https://neuroglancer-docs--pr765-whh4irzt.web.app/datasource/precomputed/annotation.html#json-PrecomputedAnnotation.annotation_type

But even if it did, it would surely be less readable than with the enum.  Not sure what the best solution there is.

PR #767, opened 2025-04-09
fix: move attribute initialization from class level to constructor

aranega:
This PR is a following of https://github.com/google/neuroglancer/pull/764, we identified in our build the `renderViewport` leading to a problematic compilation with remix/esbuild. I moved also the initialization of `boundsUpdated` for sake of harmony.

### DIFF
--- a/src/display_context.ts
+++ b/src/display_context.ts
@@ -117,9 +117,8 @@ export abstract class RenderedPanel extends RefCounted {
   canvasRelativeLogicalLeft = 0;
   canvasRelativeLogicalTop = 0;
 
-  renderViewport = new RenderViewport();
-
-  boundsUpdated = new NullarySignal();
+  renderViewport;
+  boundsUpdated;
 
   private monitorState: PanelMonitorState = { isIntersecting: true };
 
@@ -130,6 +129,8 @@ export abstract class RenderedPanel extends RefCounted {
   ) {
     super();
     this.gl = context.gl;
+    this.renderViewport = new RenderViewport();
+    this.boundsUpdated = new NullarySignal();
     context.addPanel(this);
   }
 


jbms:
Can you share more details regarding the miscompilation?  This one is kind of surprising because we don't have cyclic/self references happening in the class member initializers.

Is there an eslint or oxlint rule that would catch this?  Or a way to catch it with `tests/example_project_test` --- i.e. does the problem occur with one of the existing bundlers tested, and it is just not exercised in that test?

aranega:
I was surprised also, as you said, there is no references to this, and even more about what happened now that I try to reproduce the error. Long story short, I'm not able to reproduce the bug again, or more, I was able to reproduce it once, but not anymore. I didn't capture the stacktrace, just a screenshot of the failing compiled portion of code, I was planning on seeing well what would be the difference between the versions without/with the patch. Here is the screenshot of the error I captured

error

We can see in red that `e.gl` is problematic and the stack trace indicated that `gl` was accessed on `undefined` (`e` is the `context` in the `kk` class). This bug was triggering systematically, and I was able to trigger it once. 

After I switched to the branch with the modified code (this PR), no more issue, the problem was gone. When I first experimented that, I thought it was due to a weird side effect somewhere that required the `renderViewport` being initialized after the reference to `context.gl` (`e.gl` in the compiled code), to let it the time to initialize something in another place as it solved the issue. 

However, when I switched back now to the `master` branch, I'm unable to retrigger again this issue for some reasons.

The project that uses neuroglancer as a library is using `pnpm` with remix (esbuild). Neuroglancer as a library is compiled with `vite` in a separate "workspace project" with `pnpm` in the same repository, then linked to the main project that uses `remix`. I have no idea why we had the bug in the first place, then why it solved, and why I cannot trigger it again. I tried to flush all the `node_modules` and central local repo induced by `pnpm`, change branches and commit, but I wasn't able to trigger the problem again. I guess it's something that we will have to try to reproduce well on our side to be able to go forward.

aranega:
Hi @jbms 

So, I was able to reproduce the problem, it turns out that I didn't flush well the pnpm local repository I had, there was some remaining artifacts from the previous compilation. After investigation, it comes more from some attribute initialization in the `invlerp.ts` file:

```typescript
dataValuesBuffer = this.registerDisposer(
    getMemoizedBuffer(this.gl, WebGL2RenderingContext.ARRAY_BUFFER, () => {
    const array = new Uint8Array(NUM_CDF_LINES * VERTICES_PER_LINE);
    for (let i = 0; i < NUM_CDF_LINES; ++i) {
        for (let j = 0; j < VERTICES_PER_LINE; ++j) {
        array[i * VERTICES_PER_LINE + j] = i;
        }
    }
    return array;
    }),
).value;
// [...] there is few others
private cornersBuffer = getSquareCornersBuffer(this.gl, -1, -1, 1, 1);
```
I'm not sure why moving the attribute initialization in the other file (the one of this PR) is actually impacting the initialization of the `context` and fixes the issue, but it has to be a side-effect somewhere. Moving those attributes initializations from the `CdfPanel` class level to the constructor fixes the issue.

Following your suggestion, I created a small eslint rule that tries to detect those cases: calss-level initialization with reference to `this`.
If we consider that any reference to `this` in the attribute initialization is problematic: the rule finds 215 locations

To refine a little bit this rule, I tried to consider that references to `this` that are in an anonymous function is not problematic, as the code execution is delayed, unless the anonymous function is directly called after its creation.

It's to avoid detection of initialization like this:

```typescript
private resizeObserver = new ResizeObserver(() => this.debouncedUpdateView());
```

However, it would detect as problematic:

```typescript
private resizeObserver = new ResizeObserver(this.resizeHandler);
```
or 

```typescript
private myattribute = new MyObject(() => this.gl)())
```

If we consider this rule, it finds 166 potential problematic locations, which is better I guess as we avoid to detect initializations that shouldn't be problematic.
However, we cannot be 100% sure about the moment the anonymous function will be executed, so perhaps this rule is actually yielding false negatives. At the same time, I didn't identified really places where a closure is passed as parameter of another function to be directly executed in this function, but it might happend.

What do you think? Should we just go with the first solution of considering all reference to `this` in the initialization of a class-level attribute being problematic?

jbms:
Thanks for looking into this.  I did a bit more investigation --- I think it comes down to the fact that swc, which is used by the default rspack build configuration for neuroglancer (which is used for the normal deployment and for testing), is transforming the property initializers into assignments within the constructor body that execute after the call to `super`.  That tends to be more useful in terms of making the code work and matches the behavior of typescript when `useDefineForClassFields = false` but it does not match what tsc and esbuild do when `useDefineForClassFields = true`, which is what we have set.

To detect problems like this we really need swc to not do this property initializer transform so that it produces a result that is consistent with tsc and esbuild --- then the normal testing would catch issues like this.  Unfortunately I'm not sure if that is possible --- I wasn't able to find the correct option.

Setting `jsc.transform.useDefineForClassFields` in the rspack configuration does not seem to have any effect --- the class fields are retained as native class fields (which means define does get used) but the property initializers are still moved.

jbms:
I have a patch to swc and rspack to try to make the behavior consistent with tsc and esbuild.  However, I still don't understand what the actual bug is --- the specific change in this PR doesn't make sense to me.

Can you provide a self-contained example that reproduces the issue?

PR #769, opened 2025-04-23
feat: Add new polyline annotation type

seankmartin:
A polyline is a single annotation that contains one or more lines. They can be added from the UI or from the precomputed format

## UI placement

Polylines can be added in the UI via:

1. Creating a local annotation layer
2. Selecting the polyline tool
3. Clicking to place the points as normal with other tools
4. To indicate completion of the polyline, place two points at the same location, or hit `enter`
5. Note, during placing a polyline you can delete a vertex via `backspace`

After placing a polyline, deleting vertices is currently not supported. You can move a single vertex via alt clicking on that vertex, or the whole polyline by clicking on any line in the polyline.

## Precomputed format

Polylines are stored slightly differently to other precomputed format annotations. They require an additional piece of information, which is the number of vertices in the polyline. As such, the storage format for a single polyline is:

```
NUM_POINTS_UINT32, POINT1_FLOAT32, POINT2_FLOAT32, ..., POINTK_FLOAT32, PROPERTIES
```

## Technical note - handling a variable length annotation

Because all previous annotations were both:
1. Of fixed size given the rank
2. In the same storage format in the precomputed file and buffer for rendering (outside of possibly shuffling properties to the end of the buffer with lots of properties)

There were a few changes needed for variable length annotations. The strategy chosen in the PR right now is generally that each instance for rendering is a fixed size, but not the annotation itself. So a polyline would have a variable number of instances depending on the length of the polyline, but each of those instances is a fixed size for the buffer.
This leads to a bit of data duplication in the buffer, as the properties for a polyline are per polyline, not per vertex or per line -- but those properties are copied per line instance of the polyline.

As such, changes across the annotation handling are setup for polylines specifically. Perhaps if another variable length annotation type is added in the future, this polyline handling could be made a little more generic and a variable length annotation handling could be added. In some places, this has already been setup in a generic way, but didn't want to risk an overdesign for a possible future addition

### DIFF
--- a/python/neuroglancer/viewer_state.py
+++ b/python/neuroglancer/viewer_state.py
@@ -188,6 +188,12 @@ class PlaceEllipsoidTool(Tool):
 
 
 @export_tool
+class PlacePolylineTool(Tool):
+    __slots__ = ()
+    TOOL_TYPE = "annotatePolyline"
+
+
+@export_tool
 class BlendTool(LayerTool):
     __slots__ = ()
     TOOL_TYPE = "blend"
--- a/src/datasource/precomputed/annotations.md
+++ b/src/datasource/precomputed/annotations.md
@@ -109,6 +109,7 @@ Within the annotation id index, each annotation is encoded in the following bina
   - For `"LINE"` type, the first endpoint position followed by the second endpoint position.
   - For `"AXIS_ALIGNED_BOUNDING_BOX"` type, the first position followed by the second position.
   - For `"ELLIPSOID"` type, the center position followed by the radii vector.
+  - For `"POLYLINE"` type, the number of points as a uint32le value, followed by the position vectors of each point as float32le.
 - For each property of type `uint32`, `int32`, or `float32`: the value encoded as a little endian value.
 - For each property of type `uint16` or `int16`: the value encoded as a little endian value.
 - For each property of type `uint8`, `int8`, `rgb`, or `rgba`: the encoded value.
--- a/python/examples/write_annotations.py
+++ b/python/examples/write_annotations.py
@@ -38,13 +38,13 @@ def write_some_annotations(
         ],
     )
     writer.add_polyline(
-        [[20, 30, 40], [10, 15, 20]],
+        [[25, 10, 20], [10, 15, 20]],
         size=10,
         cell_type=16,
         point_color=(0, 255, 0, 255),
     )
     writer.add_polyline(
-        [[50, 51, 52], [10, 15, 30], [40, 20, 25]],
+        [[5, 17, 29], [10, 15, 30], [40, 20, 25]],
         size=30,
         cell_type=16,
         point_color=(255, 0, 0, 255),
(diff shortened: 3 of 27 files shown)


fcollman:
I'm having trouble ending polylines with double clicking on a spot in max projection volume rendering in 3d, can other people reproduce this problem?

seankmartin:
I ran into this trouble in large datasets in 3D in general, `enter` should end the polyline for this reason and `backspace` can be used to delete a point if an incorrect point is created because of this

fcollman:
This has incidentally highlighted to me how the chunk loading order is not optimized to be to surrounding the cursor for volume rendering, when laying down points in 3d using max projection this is pretty critical for speed, particularly when zooming and and out and triggering new MIP reloading.

fcollman:
Also, made me wonder whether I want a mode where it keeps recentering the cursor over the annotations as i lay them down, or at least makes it easy to press a button to move me to the last annotation i laid down, rather than having to shift drag my view over.

seankmartin:
They could be related to this PR (the failing tests), I added a test for placing polylines via the polyline tool from Python. The unfortunate part is that all of those annotation tool tests for Python fail for me locally (including the existing point, line etc annotation tool tests). So it's a little tricky to debug what might be wrong in the test for me

seankmartin:
> In using this PR, i've noticed that moving the points is often hard, and clicking precisely on the point doesn't work and you have to aim a bit off the point. I wish the capture distance for clicking on the point versus the line was greater if that makes sense.

Thanks, is this just for polylines or also for regular lines? Wondering if the change would ideally be across both or if this is something polyline specific

fcollman:
I think i only notice it on polylines because you can approach a line ending clearly from one side and that seems to work well, but for polylines it is harder.

src/datasource/precomputed/annotations.md
```diff
@@ -109,6 +109,7 @@ Within the annotation id index, each annotation is encoded in the following bina
   - For `"LINE"` type, the first endpoint position followed by the second endpoint position.
   - For `"AXIS_ALIGNED_BOUNDING_BOX"` type, the first position followed by the second position.
   - For `"ELLIPSOID"` type, the center position followed by the radii vector.
+  - For `"POLYLINE"` type, the number of points as a uint32le value, followed by the position vectors of each point as float32le.
```
jbms: can you also update docs/datasource/precomputed/annotation.rst

seankmartin: absolutely, thanks Jeremy

seankmartin: Hi @jbms, I updated the `src/datasource/precomputed/annotation.rst` to account for a fifth annotation type which seems to be symlinked from the `docs/datasource/precomputed/annotation.rst`. Let me know if I'm missing the intent there and I'll take another look, and sorry for the delay on this

seankmartin:
I fixed the conflicts but I see some failures in the checks afterwards, they seem not related to fixing the conflicts from what I can see but please let me know if you want to take a look at something there, thanks

python/neuroglancer/viewer_state.py
```diff
@@ -1093,6 +1099,16 @@ def __init__(self, *args, **kwargs):
...
 
 
+@export
+class PolyLineAnnotation(Annotation):
+    __slots__ = ()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, type="polyline", **kwargs)
+
+    points = wrapped_property("points", typed_list(typed_list(number_or_string)))
```
jbms, line 1109: The documentation build failure seems to be related to type annotations not propagating properly here.

See `typed_list` in `json_wrappers.py`   Not sure if there is something about a nested typed_list that is causing a problem.

chrisj:
@jbms @seankmartin it was failing every time for me on a mac. I have a fix that works for my mac at least.

https://github.com/seung-lab/neuroglancer/commit/db0d2bada7e68b90f57461315221d2048e244641

I'm not sure why return works but not enter. Also, why is it that if I make a code change to neuroglancer, it isn't picked up when I run `uvx nox -s test -- --browser firefox annotation_tool_test.py`?

I tried `npm run build-python`.

python/neuroglancer/viewer_state.py
seankmartin: Thanks Jeremy, I took a look and it seemed to be that the annotation property type which was the issue

Issue #780, opened 2025-05-15
Error, parsing the spatial unit in an OME-Zarr.


tischi:
Hi,

Loading this OME-Zarr 

[https://uk1s3.embassy.ebi.ac.uk/bia-zarr-test/mobie/clem-example/em-overview.ome.zarr](https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/bia-zarr-test/mobie/clem-example/em-overview.ome.zarr)

into Neuroglancer results in this error:

![Image](https://github.com/user-attachments/assets/feda3bb0-a0e2-4ef4-b18c-73bdfb66cd75)

Do you think this is something that could be fixed?

BTW: the OME-Zarr is valid: https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/bia-zarr-test/mobie/clem-example/em-overview.ome.zarr

jbms:
In the spec, the supported units are:

https://ngff.openmicroscopy.org/latest/index.html#axes-md

I don't see those short units listed.

jbms:
I suppose an unsupported unit could just be ignored --- ideally we could also add a way to indicate a warning.

PR #781, opened 2025-05-15
Add support for µm in OME-Zarrs

joshmoore:
Fix: #780

cc: @tischi @will-moore @dstansby 

@jbms _et al_., is there interest in listing all the enums at this point or a stronger preference for providing an iterator or mapping from other unit systems?


joshmoore:
To be honest, I was working on the fact that the validator passed. Now I'm wondering if validation did not get applied.

I leave it to you if you want to support things that are outside the spec :smile: but for my part, I'd drive that by extending the spec if/when.

joshmoore:
To limit the noise, @jbms, I'm going to close. I was primarily hoping to use this as a starting point for more folks to contribute to neuroglancer. (That offer, I imagine, still stands :wink:)

PR #785, opened 2025-05-22
Fix: annotation enum values and labels set in JSON state stay in the state

seankmartin:
If you passed an annotation with enum labels and values in the state previously, it would get parsed out of the state and show in the UI -- but the values wouldn't remain in the state. For example:

```
{
      "type": "annotation",
      "source": {
        "url": "local://annotations",
        "transform": {
          "outputDimensions": {
            "x": [
              8e-9,
              "m"
            ],
            "y": [
              8e-9,
              "m"
            ],
            "z": [
              8e-9,
              "m"
            ]
          }
        }
      },
      "tool": "annotatePoint",
      "tab": "annotations",
      "annotations": [
        {
          "point": [
            25086.916015625,
            21193.849609375,
            18610
          ],
          "type": "point",
          "id": "9209a9dc9c16bbd6181deafa3d384fed3154ab41",
          "props": [
            "#000000",
            0
          ]
        },
        {
          "point": [
            22971.654296875,
            17071.798828125,
            18610
          ],
          "type": "point",
          "id": "f911129824bff974a4edd74f986098385d36f237",
          "props": [
            "#000000",
            0
          ]
        }
      ],
      "annotationProperties": [
        {
          "id": "color",
          "type": "rgb"
        },
        {
          "id": "enumTest",
          "type": "uint8",
          "enum_values": [0, 1],
          "enum_labels": ["test0", "test1"]
        }
      ],
      "name": "annotation"
    }
 ```

Would lose the `enum_values` and `enum_labels`

This is to change so that they are also in the main viewer state. I also noticed that the `AnnotationNumericPropertySpec` in `annotation/index.ts` supports a `min`, `max`, and `step`. Not sure if we also need to handle these? But right now I can't see where `min`, `max` and `step` would get used

### DIFF
--- a/src/annotation/index.ts
+++ b/src/annotation/index.ts
@@ -608,21 +608,24 @@ function parseAnnotationPropertySpec(obj: unknown): AnnotationPropertySpec {
 function annotationPropertySpecToJson(spec: AnnotationPropertySpec) {
   const defaultValue = spec.default;
   const handler = annotationPropertyTypeHandlers[spec.type];
-  const jsonSpec: any = {
+  let enumValues: number[] | undefined;
+  let enumLabels: string[] | undefined;
+  if ("enumValues" in spec) {
+    enumValues = spec.enumValues;
+    enumLabels = spec.enumLabels;
+    if (enumValues !== undefined) {
+      enumValues = enumValues.map(handler.serializeJson);
+    }
+  }
+  return {
     id: spec.identifier,
     description: spec.description,
     type: spec.type,
     default:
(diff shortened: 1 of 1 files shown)


seankmartin:
@chrisj I made a kind of hybrid here in https://github.com/google/neuroglancer/pull/785/commits/f2272d703fffc80dd72d6c9fcfeacac7fee05d92 with what you had in https://github.com/google/neuroglancer/pull/650/files once I noticed it

Slight change is that I'm still calling the serialize to JSON here on the numbers just in case there is ever a numeric property type which doesn't just pass the number through

PR #789, opened 2025-06-04
feat(url_hash_binding) added settings to rate limit url updates and optionally use session storage

chrisj:
also updated state share ui to match the look/feel of the tool pallete dropdown (on right click) along with the new option to copy url.

Sven passed along a suggestion from you, iirc, to use a browser event to update the url when the user clicks into the url bar. The blur event seems to work pretty well. Sometimes I end up getting a partial selection of the url but that might be from a slight movement of the cursor when clicking. Most of the time, it works perfectly, updating the selection with the new state.

With that feature, maybe things like the "copy as url" action in the state share dropdown, are now unnecessary.

Screenshot 2025-06-04 at 1 18 43 PM

Also, I can split the state_share changes into a separate PR. There is definitely an opportunity to reuse the dropdown logic/styling between the two.

### DIFF
--- a/src/datasource/state_share.ts
+++ b/src/datasource/state_share.ts
@@ -1,8 +1,14 @@
+import "#src/datasource/state_share.css";
 import { ReadableHttpKvStore } from "#src/kvstore/http/common.js";
 import { joinBaseUrlAndPath } from "#src/kvstore/url.js";
 import { StatusMessage } from "#src/status.js";
+import { WatchableValue } from "#src/trackable_value.js";
+import { encodeFragment } from "#src/ui/url_hash_binding.js";
 import { RefCounted } from "#src/util/disposable.js";
+import { removeFromParent } from "#src/util/dom.js";
+import { positionRelativeDropdown } from "#src/util/dropdown.js";
 import { bigintToStringJsonReplacer } from "#src/util/json.js";
+import { getCachedJson } from "#src/util/trackable.js";
 import type { Viewer } from "#src/viewer.js";
 import { makeIcon } from "#src/widget/icon.js";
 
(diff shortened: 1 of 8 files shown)


src/viewer.ts
```diff
@@ -282,7 +288,7 @@ class TrackableViewerState extends CompoundTrackable {
     this.add("statistics", viewer.statisticsDisplayState);
     this.add("helpPanel", viewer.helpPanelState);
     this.add("settingsPanel", viewer.settingsPanelState);
-    this.add("selection", viewer.selectionDetailsState);
+    this.add("selection", viewer.selectionDetailsState); // TODO, if I disable this, do we stop updating the state when we move the mouse?
```
jbms: this should only trigger an update when the "pinned" state changes, but maybe there is a bug --- you could check

src/datasource/state_share.ts
```diff
@@ -15,119 +21,240 @@ type StateServers = {
   [name: string]: StateServer;
 };
 
-declare const STATE_SERVERS: StateServers | undefined;
-
+const STATE_SERVERS: StateServers | undefined = {
```
jbms: I assume this was not intended to be included in this PR

src/viewer.ts
chrisj: ~I see now, if selection is visible but not pinned, it will trigger state updates whenever it changes, I will look for a way to disable state changes when not pinned.~

Never mind, this is handled by the `stateString !== this.prevStateString` check in ` setUrlHash`

ceesem:
Just wanted to nudge this PR! Both of these features offer huge quality of life improvements both for annotation-rich workflows and to allow the same neuroglancer deployment to be used across multiple server deployments.

PR #790, opened 2025-06-05
fix(datasource/graphene) improve multicut off color behavior

chrisj:
Make all segments aside from the focus segment (including segment id 0) hidden (same as off color) as well as making them appear to have no highlight effect by setting the highlight color to be the same as the off color (other than the focus segment)


fcollman:
@chrisj @seankmartin @jbms I think this is really an isolated changed to graphene and can be merged.

jbms:
Sorry I let this sit for a long time but my general concern is that it modifies layer settings that are part of the json stat.  I introduced the temporary settings specifically to avoid having tools like this corrupt the persisted layer settings but we don't have temporary settings for everything.  The solution is probably to add more.

That said it is true that this only affects graphene and won't affect users that don't use graphene.

chrisj:
@jbms we already started down the path of the temporary values and having the useTemporary toggle. I'm thinking of some alternatives since that could start to feel bloated if we might eventually want to do it for every possible setting.

What about way to override with something like a `Partial<DisplayState>`?

jbms:
Yeah in general a way to override everything would be good.  And furthermore a way to associate the override with a temporary scope that either merge or stack so that multiple tools being used so once (not really supported now but will surely be useful in the future) don't have to worry about interfering with each other.

PR #791, opened 2025-06-07
Fix state share copy to clipboard in insecure context (http)

RussTorres:
Using the "Share" button to post to a state server on an http deployment on most browsers will send the correct request, but quietly fail copying the link to the clipboard because the navigator requires a secure context.

In this commit I introduce a function that replicates the original behavior in the secure context, while providing a workaround when the original method would fail.  I also modify the returned link to include the pathname, which I believe was leading to incompatible copied links when being serving on a subpath.

It's not uncommon to want to stand up http versions of neuroglancer to talk to custom state servers or view local http data sources -- I think this makes deployment easier for those cases and hopefully eliminates some confusion -- let me know what you think.

Thanks!
Russel

### DIFF
--- a/src/datasource/state_share.ts
+++ b/src/datasource/state_share.ts
@@ -6,6 +6,30 @@ import { bigintToStringJsonReplacer } from "#src/util/json.js";
 import type { Viewer } from "#src/viewer.js";
 import { makeIcon } from "#src/widget/icon.js";
 
+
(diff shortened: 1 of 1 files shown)


jbms:
Thanks for looking into this.  There is an existing method in src/util/clipboard.ts --- does that solve this problem?

jbms:
Please fix the lint error.  Thanks

Issue #782, opened 2025-05-19
Position plot ticks and bounds with archived layers. Also possible to just display min/max bounds on position plot?


seankmartin:
When there are archived layers in neuroglancer, the position plot still shows the bounds of the data for those archived layers

![Image](https://github.com/user-attachments/assets/a3f92b84-b135-491b-b967-a3327239b973)

Perhaps it might be good to exclude these, and maybe also non-visible layers? Though less sure on the non-visible layers.

I was also wondering if neuroglancer might support a build-time option that allows the position plot to only display the minimum lower bound and maximum upper bound of all the provided data, instead of plotting a tick for each separate lower and upper bound in the co-ordinate space. That could look something like this:

![Image](https://github.com/user-attachments/assets/c91b8348-63ae-42cf-b64d-d2311bf347d8)

PR #792, opened 2025-06-12
Feat/cc 376

seankmartin:
Relates to https://github.com/google/neuroglancer/issues/782. This PR would allow the position plots to only show the min/max bounds of the data optionally, and not all the unique bounds as is the current behaviour.

The proposed solution here is to allow a new watchable value in the `viewer.uiConfiguration` which is called `showAllDimensionPlotBounds`. So you could see only the min/max dimension bounds by calling `viewer.uiConfiguration.showAllDimensionPlotBounds.value = false`. That would look like:
image

In contrast to:
image

The setting should apply to the global viewer position plots, and also to position plots that are made from new layer groups which don't have their position linked to the global position.

I also considered this being in the settings that the user can change, so it would be in the state. But I think it is more in line with other UI change patterns that this is part of the `uiConfiguration`.

### DIFF
--- a/src/widget/position_plot.ts
+++ b/src/widget/position_plot.ts
@@ -164,7 +164,7 @@ export class PositionPlot extends RefCounted {
         canvas.height = this.canvasWidth;
       }
 
-      let normalizedDimensionBounds = getNormalizedDimensionBounds(
+      const normalizedDimensionBounds = getNormalizedDimensionBounds(
         coordinateSpace,
         dimensionIndex,
         canvasHeight,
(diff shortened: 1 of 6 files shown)


src/widget/position_plot.ts
```diff
@@ -177,6 +179,25 @@ export class PositionPlot extends RefCounted {
         this.visible = false;
         return;
       }
+      if (this.showAllBounds?.value === false) {
+        let minLowerBound = Infinity;
```
chrisj: Any issue with simplifying it to this?
```ts
const { normalizedBounds } = normalizedDimensionBounds;
normalizedDimensionBounds.normalizedBounds = [
  {
    lower: Math.min(...normalizedBounds.map((b) => b.lower)),
    upper: Math.max(...normalizedBounds.map((b) => b.upper)),
  },
];
```
It might be possible for normalizedBounds to be empty though it would be an existing bug. If it can happen, it should be dealt with in the prior if block so the plot isn't displayed

seankmartin:
> Not relevant to the change but playing around with this made me realize the current plot behavior doesn't react to layer visiblity/archive.

Yes I noticed the same, I'm going to update https://github.com/google/neuroglancer/issues/782 to focus on just that part, thanks Chris

src/widget/position_plot.ts
seankmartin: Happy to, thank you very much for the suggestion!

PR #803, opened 2025-07-07
Store CoordinateSpace elements in JSON state as array not object since order matters

seankmartin:
## Summary

Changes all `CoordinateSpace`s from an object with names as keys to arrays where each element has a `name` property and the value is stored under the `scale` property. This is to indicate that order matters and be more consistent with the JSON specification as objects are not guaranteed to maintain order of entries.

## New format example

For example, on the state dimensions, this is the current:
```
"dimensions": {
    "x": [
      4e-9,
      "m"
    ],
    "y": [
      4e-9,
      "m"
    ],
    "z": [
      4e-8,
      "m"
    ]
  }
```

This is the proposed new dimensions:
```
"dimensions": [
    {
      "name": "x",
      "scale": [
        4e-9,
        "m"
      ]
    },
    {
      "name": "y",
      "scale": [
        4e-9,
        "m"
      ]
    },
    {
      "name": "z",
      "scale": [
        4e-8,
        "m"
      ]
    }
  ],
```

## Update as of Feb 20th

We've expanded this PR quite a bit since the original. All of these changes are on the Python code. The original missed the fact that all `CoordinateSpace`s needed to change, not just the `dimensions` (also so `inputDimensions`, `outputDimensions` and `localDimensions`). Additionally, the original did not respect the `_readonly` flag which could be present on the viewer state. Lastly, the original did not account for the fact that neuroglancer allows `layers` to be both an array and an object, or that `source` on a layer is allowed to be a single object instead of an array with one object element in the array. These extra cases are now handled. 

Given the extra complexity involved, multiple Python tests have been included to ensure that the different types of `CoordinateSpace` are handled and the different representation of `layers`, `source` and if the state is `_readonly`.
cc @fcollman @chrisj

## Motivation and more details

Since JSON objects are unordered in the spec, this could cause issues in processing of neuroglancer states by external tools for objects stored in the state where the name order matters. For example, in something that saves states or processes states.

The primary case where this has come up is in the global viewer `dimensions` and all other `CoordinateSpace`s, so `inputDimensions`, `outputDimensions` and `localDimensions` - where the order of the names determines the default order of the viewer, so the order matters. I know the precomputed annotation info metadata has a similar pattern too though.

In neuroglass it directly stores neuroglancer JSON states using django ORM https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.JSONField. This uses PostgreSQL https://www.postgresql.org/docs/current/datatype-json.html jsonb format, which doesn't guarantee to keep object keys in order.

Overall we are suggesting a change here because having the JSON state being reliant on object key ordering means it's harder to provide a reliable way to store and share the state in JSON format. 

To help with this, this PR is proposing to change the dimensions to be stored in an array, not an object. Here, the dimensions are handled in a split path. One "legacy" path, for pure dictionary style objects, and a new path, which handles arrays. New states are saved as arrays in the JSON state. This happens pretty automatically in the viewer, but from Python there is an extra kind of conversion call which happens on init of a new state.

### DIFF
--- a/python/neuroglancer/coordinate_space.py
+++ b/python/neuroglancer/coordinate_space.py
@@ -336,12 +336,10 @@ class CoordinateSpace:
     def supports_validation(value):
         """Validator function for CoordinateSpace in wrapped_property contexts.
 
-        This allows CoordinateSpace to accept:
+        This allows CoordinateSpace to validate:
         - CoordinateSpace instances (returns as-is)
         - dict format: {"x": [scale, "unit"], "y": [scale, "unit"], ...}
         - list format: [{"name": "x", "scale": [scale, "unit"]}, ...]
-
-        Without mutation of the input JSON.
         """
         if isinstance(value, CoordinateSpace):
             return value


src/coordinate_transform.ts
```diff
@@ -333,21 +316,66 @@ export function coordinateSpaceFromJson(
...
+    };
+  }
+}
+
 export function coordinateSpaceToJson(coordinateSpace: CoordinateSpace): any {
   const { rank } = coordinateSpace;
   if (rank === 0) return undefined;
   const { names, units, scales, coordinateArrays } = coordinateSpace;
-  const json: any = {};
+  const json: any = [];
```
chrisj: minor, but the `any` isn't required here anymore

Also, it looks like there is a mix in the return type for the various toJSON functions in neuroglancer, some specify any, some leave it out. Since it is an array, it feels more natural to me to not cast it to any.

Rest of the PR looks good!

chrisj:
I'm in favor of using arrays to make it clear that the order matters/more compatible

src/coordinate_transform.ts
seankmartin: Agreed, thanks for spotting!

python/neuroglancer/viewer_state.py
```diff
@@ -1867,6 +1867,61 @@ class DataSelectionState(SidePanelLocation):
     layers = wrapped_property("layers", LayerSelectedValues)
 
 
+def convert_legacy_coordinate_spaces_in_viewer_state(json_data):
```
jbms: I think it would be better to do this conversion on the individual types -- e.g. `CoordinateSpace` should just accept either a dict or a list.  That way existing code that creates the wrapper types from JSON directly will continue to work.

Also we don't want to mutate the existing json object in place.

seankmartin: Thank you Jeremy for the feedback, I will look at making this change

seankmartin: I changed this so hopefully it is clearer now. This comes in two parts:

1. I've removed the viewer state level json conversion to the array format. This was intended to help applications using neuroglancer python in case they needed to have the state in JSON format always be in array format for anything which uses CoordinateSpaces in case they run into unintended consequences. But thinking on this further, I believe it likely better to remove that. Instead an application can rely on https://github.com/google/neuroglancer/pull/803#issuecomment-3927172731 and the application can maintain the needed conversion function.
2. I was trying to understand the comment about the `CoordinateSpace` accepting either `dict` or `list`. This code had already made that change. But I dug through how the `JsonObjectWrapper` works in a bit more detail and I think I see the issue. The `CoordinateSpace` didn't have a validator setup. I've added `supports_validation` to the `CoordinateSpace` class with a custom function there and a test for this to ensure you can call `isinstance` etc on wrappers that use the `CoordinateSpace`. I've also added a test about the json object mutation in place, though I believe that was likely aimed at the viewer state conversion from point 1. That was mutating the json object in place if `_readonly` was not true for application reasons. But again, as in point 1 better to remove that likely and just let the application handle that.

PR #805, opened 2025-07-08
feat(ui): use the new standard dialog with header and footer across shader code dialogs and state editor

seankmartin:
We introduced a new framed dialog class, and this PR intends to use it across the shader code editors and state editor.
This would allow removing custom code setting up different kinds of shader code editor overlays/dialogs.

Open to exact stylings there, don't feel strongly about those. The main aim was to reduce code duplication and have something to use as a basis for future dialogs. Overall, the dialogs would look something like this:

(edit - this originally introduced the framed dialog and applied it to the screenshot menu, but those changes have since been merged elsewhere so now this PR is just about applying it in a new place)

image
image
image


src/widget/shader_code_widget.ts
```diff
@@ -82,6 +85,7 @@ export class ShaderCodeWidget extends RefCounted {
       value: this.state.fragmentMain.value,
       mode: "glsl",
       gutters: ["CodeMirror-lint-markers"],
+      lineNumbers: true,
```
seankmartin, line 88: I think this is a left over change from a choice in neuroglass. Can safely be removed if we prefer

src/overlay.ts
```diff
@@ -41,10 +41,10 @@ export class Overlay extends RefCounted {
     this.keyMap.addParent(defaultEventMap, Number.NEGATIVE_INFINITY);
     ++overlaysOpen;
     const container = (this.container = document.createElement("div"));
-    container.className = "overlay";
+    container.className = "neuroglancer-overlay";
```
seankmartin, line 44: This change is so that the class name is less likely to overlap with anything which might be using neuroglancer

seankmartin (commented):
@chrisj I fixed the conflicts on this one

PR #809, opened 2025-07-16
feat: add annotation schema tab to annotation panel and editable properties for annotations

seankmartin:
Readonly datasources get a readonly view of the schema. Editable datasources (e.g. local annotation layers) get an editable version of the schema to add properties to. Editable datasources can also modify properties of annotations from the selection panel.

image

image


seankmartin:
Thanks for the suggestions! I opened the first as an issue with some ideas for now

fcollman:
@seankmartin I think you need to rebase to get the mac tests to pass.. @chrisj has got the CLA agreement fixed.

chrisj:
The schema tab feels great and feels like a naturall fit!

Screenshot 2026-04-07 at 1 19 09 PM

It seems the change type dropdown is incorrectly showing the numeric property icon instead of the enum icon when changing an enum.

The dropdown is also missing two things that seem to be working for the add property dropdown. It doesn't flip direction if it would otherwise overflow off the bottom of the screen and it doesn't have the same hover highlight.

chrisj:
The paste from clipboard feature is very convenient though I was a little unsure how to use it at first. Now I know, I wouldn't suggest any changes. I would add an upload schema action to match the corresponding download.

seankmartin:
thanks @chrisj! added/fixed those

src/annotation/index.ts
```diff
@@ -1479,14 +1572,112 @@ export class LocalAnnotationSource extends AnnotationSource {
...
+    if (!isConvertible) {
+      console.error(
+        `Cannot convert property ${oldProperty.identifier} from ${oldProperty.type} to ${newProperty.type}`,
+      );
+      return;
+    }
+
+    const convertValue = (value: any) => {
+      if (oldType === "rgb" && newType === "rgba") {
+        const rgba = new Uint8Array(4);
```
chrisj: The value here seems to be a packedRGB so you need to unpack and repack it. Converting a property from RGB to RGBA destroys the values.

chrisj:
AI analysis is concerned about some situations with multiple annotation sources per layer, particularly if some are editable and some are read only. I'm still figuring out if any of them are real/reproducible.

src/annotation/index.ts
seankmartin: Thank you for spotting that Chris, I wrote an extender instead of unpacking and repacking but same thing really so hopefully that fixes it. Cheers!

seankmartin:
> AI analysis is concerned about some situations with multiple annotation sources per layer, particularly if some are editable and some are read only. I'm still figuring out if any of them are real/reproducible.

Thanks Chris, I think that's likely an issue, as admittedly it wasn't really a situation thought through. I can, for example, add a local annotation source to an annotation layer and then because one of the sources is editable even though the source doesn't match in properties - the annotation schema now thinks the whole schema is editable. Of course the edits don't make it through in most cases because the most common type (precomputed) doesn't support writing. But it's definitely an issue if more sources can be read/write, a source declares itself to be in read only mode, and then you add a local annotation source to the layer and now you can write schema changes to the initial source.

To have two annotation sources in a layer their properties need to be an exact match. However, for local annotations the annotation properties in the state are stored at the layer level not the source level. I think the safest bet for now would be to change the logic on the editable check to check if all annotation sources in the layer are editable. Right now it checks if any source is editable.

Very happy to discuss alternatives, especially if that blocks some kind of workflow I'm missing

src/annotation/index.ts
chrisj: even better!

chrisj:
@seankmartin thanks for these updates! A few remaining issues.

* The upload schema button you added at my request is showing up for readonly sources.
* Annotation selection details panel does not update when the schema changes.

src/ui/annotations.ts
```diff
@@ -1986,15 +2007,38 @@ export function UserLayerWithAnnotationsMixin<
       const reference = context.registerDisposer(
         annotationLayer.source.getReference(state.annotationId),
       );
+      let newAnnotation = reference.value;
+      const throttledUpdate = context.registerCancellable(
+        throttle(() => {
+          if (!newAnnotation) return;
+          this.allowDependentAnnotationViewUpdate.value = false;
```
chrisj, line 2021: I would add a little comment here because it was hard for me to follow but I understand it as:

When editing an annotation from the ui controls within the selection details, we don't need to update the selection details since the user input already performed the necessary update. This is important for at least the color picker control because otherwise it causes the color picker to immediately close after changing the value.

seankmartin:
> * Annotation selection details panel does not update when the schema changes.

Thanks for spotting these issues! Didn't realise I'd forgotten to add a call to the signal that the layers changed on property update (was present for the property add/remove)

chrisj (approved):

PR #812, opened 2025-07-21
fix: allow tool palette dim tool to set playback velocity

seankmartin:
Previously the tool palette dimension playback velocity input was disabled
image

I guess this was because the tool which shows on keybind intended to have that input disabled, but I think it's not really possible to change it there anyway from the input


jbms:
I think I added this because with the keybind since you have to hold down shift+letter if the input field got focus, it was getting filled with the letter.  However I can't replicate that problem now with your change.

PR #813, opened 2025-07-22
feat: Adds support for DVID nodes like 'f3969:master'

neomorphic:
To support special nodes like "<UUID>:master" (which  means "the most distant unlocked descendant of <UUID>"), use only "<UUID>" for the validity lookup. If it is valid, then DVID itself will resolve the part after the ':'.

DVID Data source urls can now be formatted like this:

dvid://https://dvid.janelia.org/f3969:master/segmentation


stuarteberg:
Thanks, @jbms!

PR #820, opened 2025-07-30
feat(skeleton): skeleton line width control and conversion from nanometers

chrisj:
added setLineWidth, setNodeDiameter, glsl_nanometersToPixels. Migrated skeleton shader editor to control vertex shader instead of fragment shader (match behavior of annotation layer).

Example shader:

```glsl
void main() {
  setLineWidth(max(1.0, nanometersToPixels(min(1500.0, radius))));
  setNodeDiameter(nanometersToPixels(radius));
  
  if (compartment == 1.0) {
    emitRGB(vec3(1.0, 0.0, 0.0));
  }
  
  if (compartment == 2.0) {
    emitRGBA(vec4(1.0, 0.0, 1.0, 1.0));
  }
  
  if (compartment == 3.0) {
    emitRGB(vec3(0.0, 1.0, 1.0));
  }
}
```

Screenshot 2025-07-30 at 2 08 58 PM


src/skeleton/frontend.ts
```diff
@@ -165,41 +181,62 @@ class RenderHelper extends RefCounted {
...
+  vColor = color;
+}
+`);
           let vertexMain = `
 highp vec3 vertexA = readAttribute0(aVertexIndex.x);
 highp vec3 vertexB = readAttribute0(aVertexIndex.y);
-emitLine(uProjection, vertexA, vertexB, uLineWidth);
 highp uint lineEndpointIndex = getLineEndpointIndex();
 highp uint vertexIndex = aVertexIndex.x * lineEndpointIndex + aVertexIndex.y * (1u - lineEndpointIndex);
+highp uint vertexIndex2 = aVertexIndex.y * lineEndpointIndex + aVertexIndex.x * (1u - lineEndpointIndex);
```
chrisj: I am a bit confused why I needed to add this other vertexIndex. The original seems to access the data for the opposite node for each line segment.

chrisj:
I added support for non-float vertex attributes built on top of this branch since I already made the step here of moving the editable shader from being a fragment shader to a vertex. This simplifies things because we don't have to worry about the flat qualifier and provoking vertex but that is an alternative. There are some backwards compatibility issues here, maybe the non-float change should be merged with https://github.com/google/neuroglancer/pull/819 to deal with backwards compatiblity.

I could keep the original name and vCustom as floats and `prop_name()` as the new type.

3a11257f8a526270036fdf76f7ff2fc795979d34

jbms:
Taking radius into account when rendering skeletons has been requested a lot and would certainly be a valuable feature.  However, I think it may be better to implement it directly as a separate rendering mode for a few reasons:
1. We could then render as cylinders (with appropriate shading) rather than just wider lines.
2. It is simpler to deal with the coordinate transforms and physical units outside of the user-defined shader.

The main downside, I suppose, would be less flexibility in deriving the radius from arbitrary properties in the shader.  But is that really needed?

fcollman:
"The main downside, I suppose, would be less flexibility in deriving the radius from arbitrary properties in the shader. But is that really needed?"

I think the short answer to this is yes..

here is an example using chris's branch.
image

```
  #uicontrol float myelin_thrsh slider(min=0, max=100, default=90)
  #uicontrol float myelin_thickness slider(min=0, max=50, default=10)
  
  void main() { 
    if (myelin() > myelin_thrsh) {    
      setLineWidth(myelin_thickness);
  	  setNodeDiameter(myelin_thickness);
    } else {
      setLineWidth(uLineWidth);
  	  setNodeDiameter(uNodeDiameter);
    }
  }
```
which was able to be done with this shader, where the user had defined a mylenation score along the skeleton. 

In general you could use this to transform skeleton vertex data into customized line widths to emphasize certain portions of a skeleton.

fcollman:
"For 3-d perspective projection views, even if there is uniform scaling for non-point annotations there is more than one location involved meaning there isn't a single scale factor." 

I think perhaps you didn't understand that this is the point of this function, it will dynamically calculate the line width based on the camera location, so that when the line is closer it is scaled differently than when that same line is farther away.

jbms:
> "For 3-d perspective projection views, even if there is uniform scaling for non-point annotations there is more than one location involved meaning there isn't a single scale factor."
> 
> I think perhaps you didn't understand that this is the point of this function, it will dynamically calculate the line width based on the camera location, so that when the line is closer it is scaled differently than when that same line is farther away.

For a line annotation, the "nanometersToPixels" mapping may not be the same for the two points (if they aren't at the same depth relative to the perspective camera view).  Currently, each annotation is rendered via a number of vertices ---- 4 for each line, plus 4 more for each endpoint marker --- and the user-defined shader runs as part of the vertex shader for each vertex, and is expected to produce identical output for each vertex, since the user currently has no (documented) way to vary the behavior depending on which vertex is being processed.  In order for `nanometersToPixels` to work correctly it would need to give different results depending on which vertex is being processed.  This might work though it would potentially be confusing to users given how annotation shaders currently work.  With skeletons, in contrast, the shader explicitly just operates on a single vertex so nanometersToPixels is well defined.

jbms:
I suppose it is fine to add this just to skeletons for now if that is useful, provided that we can figure out a reasonable way to handle the coordinate transforms and units.

chrisj:
@jbms I added an implementation for the line annotation, looks like this

```glsl
float sourceDimToPixels(float value) {
  vec2 lineOffset = getLineOffset();
  highp vec3 vertexA = projectModelVectorToSubspace(getVertexPosition0());
  highp vec3 vertexB = projectModelVectorToSubspace(getVertexPosition1());
  highp vec3 vertex = mix(vertexA, vertexB, lineOffset.x);
  return nanometersToPixels(value, vertex, uProjection, uModelView, 1.0 / uLineParams.x);
}
```

For now I renamed it to sourceDimToPixels to pixels as that seems to be more accurate as our skeleton uses a 1x1x1 source dimension whereas local annotations by default use voxel scale by default. This way, we shouldn't have to worry about the units if I understand it correctly. Below is a screenshot showing a line annotation in blue that I have roughly positioned in line with a segment of the skeleton.

It could be useful to expose the source dimensions in the shader.

Screenshot 2025-08-14 at 9 38 58 AM


I see the issue with coordinate transforms, I'll look into that correction.

Screenshot 2025-08-14 at 9 47 40 AM

Issue #814, opened 2025-07-25
Should not toggle selection widget when selecting new segment


unidesigner:
When I select a segment in the 3d view with Ctrl - Right Click, I'd like to see the segment in the pinned selection widget (i.e. the segment is available in the json state). However, if I now select another segment, the selection widget is unpinned (and the segment removed from the state). Can this behavior be changed so that the widget remains pinned? With a reference to the relevant code, I can try to submit a PR. Thank you.

jbms:
This is intended behavior --- ctrl+right click toggles whether the selection is pinned.  To go from one pinned selection to another, you can control+right click twice.  That way a separate binding isn't needed for unpinning.

I'm open to proposals for improvement --- but I do think being able to unpin easily is important.

ceesem:
I just ran across this issue after discussing with @fcollman about how the selection pin kept winding up in different states without me understanding why.

I'm strongly in agreement with @unidesigner and @seankmartin. If the user ctrl-right-clicks to makes a selection, it's quite clear that they want an ID to become selected. Having the state switch every other click is difficult to square with user intentions and makes the UX quite unpredictable in a way that destroys information. In addition, this makes the unpinned state a confusing double action — it is both the mechanism to reset the selection by unpinning and hovering the cursor somewhere unselected and a mode where you have to carefully control where the cursor goes to avoid changing selection state. Which of these two aspects do you find important? I would also be curious about the important use case for the unpinned state, which is less clear to me than the pinned state.

Would it be possible a more specific about what more you're looking for in terms of a proposal for improvement, @jbms? It seems like several of us have typical use cases that differ significantly from the use case your current implementation aims to make convenient, and so without a better idea of that I worry that we can't propose an improvement that goes beyond this PR.

chrisj:
@jbms we were brainstorming a few ideas

First is to separate selection and hover into separate modes. You can toggle between displaying the selection and what is actively hovered. We could change the title of the panel to "Hover" when it is in hover mode to make it more obvious. This would allow you to preserve your selection while hovering.

As far as un-pinning, we could add add a +shift keybind, or un-pin if every layer value is the same as the previous pinned selection. Un-pinning would swap back to hover mode.

fcollman:
I think "hover mode" was just a rebranding of the "unpinned" mode we had before, we were hoping a larger visual change in the widget title/color might make it more clear to users what mode they were in.  The pin is kind of subtle. 

[...]

PR #826, opened 2025-08-15
feat: ctrl+right click always pins the new selection instead of toggling the pinned state

seankmartin:
See https://github.com/google/neuroglancer/issues/814 I think this enables the intended behaviour there if I'm understanding correctly. I've also been a little confused in the past when trying to pin a new segment but actually just unpinning my current segment. It does mean you have to manually unpin now from what I can tell though. Anyway open to thoughts from @unidesigner @jbms

### DIFF
--- a/src/layer/index.ts
+++ b/src/layer/index.ts
@@ -1432,10 +1432,8 @@ export class TrackableDataSelectionState
   select() {
     const { pin } = this;
     this.location.visible = true;
-    pin.value = !pin.value;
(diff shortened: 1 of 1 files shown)


seankmartin:
I'll close this for now, could be reopened if a different mechanism to unpin is found

fcollman:
I think this is a relatively simple PR code wise and functionality I think is an improvement.  I'd give it a thumbs up.

PR #827, opened 2025-08-21
feat(sliceview): add uEffectiveVoxelSize to sliceview shader so it can be used in the user editable shader

chrisj:
One such use case is max projection shader that wants to search over a certain physical distance rather than voxels which also allows it to maintain a similar search space between scales.

```glsl
int voxel_range = int(ceil(float(max_range) / (length(uPlaneNormal * uEffectiveVoxelSize) * 1000000.0)));
```


fcollman:
@seankmartin @jbms do you guys have any thoughts about this? Seems like a simple addition that wouldn't break much and would enable some interesting shaders.

fcollman:
The immediate use case was to enable maximum projection views within the slice shader, where the maximum intensity would need to be a fixed physical distance to make sense, and so the number of voxels you would render above/below any particular voxel would change.  We were planning on making the step size coordinated with the physical voxel size to ensure that any max projection would not cross chunk boundaries.  A general purpose max slider won't work properly in all cases, so it felt like it had to be a shader that was only meant for certain volumes, but without access to the physical scale one cannot write it to behave correctly.

jbms:
I don't think this is the sort of thing we can reasonably support in the shader, because it breaks through the intended abstraction layer.

But we could have an option to do a volume rendering mode for the cross section views also.  There already is a depth parameter just like in the 3d projection view that could be used, unless you think a separate per-layer depth parameter is needed

PR #831, opened 2025-09-01
making showLayerPanel toggleable in the layerPanel

fcollman:
image
There were some options for layer panels that were not part of the state, and no way for UI users to set those options.  I have added a showLayerPanel visibility toggle to the layer list side panel, so that people can turn off the visibility of the layer list. 

This is particularly useful for those people who are trying to make simplified neuroglancer visualizations, and/or users who are working with large numbers of layers.

### DIFF
--- a/src/ui/layer_list_panel.ts
+++ b/src/ui/layer_list_panel.ts
@@ -359,7 +359,7 @@ export class LayerListPanel extends SidePanel {
       toggleButton.element.style.order = "50"; // Position before close button (order: 100)
       titleBar.appendChild(toggleButton.element);
       this.registerDisposer(toggleButton);
-      
+
       // Watch for showLayerPanel configuration changes and hide/show button accordingly
       this.registerDisposer(
         this.showLayerPanel.changed.add(() => {
--- a/src/ui/layer_list_panel.ts
+++ b/src/ui/layer_list_panel.ts
@@ -24,6 +24,7 @@ import type {
   TopLevelLayerListSpecification,
 } from "#src/layer/index.js";
 import { deleteLayer } from "#src/layer/index.js";
+import type { TrackableBoolean } from "#src/trackable_boolean.js";
 import { TrackableBooleanCheckbox } from "#src/trackable_boolean.js";
 import type { DropLayers } from "#src/ui/layer_drag_and_drop.js";
 import {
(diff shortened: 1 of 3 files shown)


fcollman:
makes sense.  I think having the layer side panel makes it possible to get to the layer controls, I did consider putting some kind of eyeball button next to all the layers to just hide all the elements of the list, but since there was already this mechanism I thought this would be a better direction. 

I will try to make the layerPanel visible option a seperate option that is tracked in the state.. and if the existing UI configuration option is disabled I'll also turn this option off (so that the user cannot make the layerPanel visible).  

I do think that having it available in the UI is useful, particularly given that you have the layerPanel side panel now which can modally accessed and hidden.

src/ui/layer_list_panel.ts
```diff
@@ -338,11 +339,44 @@ export class LayerListPanel extends SidePanel {
...
+            if (!toggleButton.element.parentElement) {
+              titleBar.appendChild(toggleButton.element);
+            }
+          } else {
+            // Configuration disabled - hide button
+            if (toggleButton.element.parentElement) {
+              toggleButton.element.remove();
+            }
+          }
+        }),
```
seankmartin: Could maybe simplify this check if removing the icon fully is not needed and instead hiding is sufficient along the lines of this?
```suggestion
          toggleButton.element.style.display = this.showLayerPanel!.value
            ? ""
            : "none";
```

seankmartin (approved):
I think this is a really great change. I've seen many cases where the layer bar takes up so much space. I know I can hide it from console by accessing the viewer, but so few users will know that. 

I appreciate though that this raises a bit of a question of which viewer UI config elements should be in the state and which shouldn't. My only thinking would be that to anticipate that, we maybe nest this under some kind of section. And then if we need to bring any other pieces of UI config into the state later, they could essentially use the commented out code in `viewer.ts` L293 but be placed into this new section. Suggestions for names could be `uiVisibility` or `uiControlVisibility`.

Would you agree @fcollman @jbms @chrisj?

(edit, I'm approving because I think even without the change suggested above it is worth. In case we don't agree on where in the state it should go)

fcollman:
I got sent a link like this today which inspired me to come back to this PR

image

Sure there are other ways they could have made this link now, but there are people making things like this for various reasons.

seankmartin:
> I got sent a link like this today which inspired me to come back to this PR
> 
> image
> Sure there are other ways they could have made this link now, but there are people making things like this for various reasons.

I still think this is a great change, any thoughts on it @chrisj @jbms?

fcollman:
I have moved the visibility into the uiControlVisibility subsection and made it tri-state... going to look into what this test failure is just in python 3.13

fcollman:
improved version of the same link with the PR
image

fcollman:
@chrisj @jbms can someone rerun these tests.. i think they just timed out.

chrisj (approved):

PR #834, opened 2025-09-17
perf(jxl): faster decoding via fewer frame decodes

william-silversmith:
Faster decoding via avoiding two extraneous full frame decodes.

### DIFF
--- a/src/sliceview/jxl/src/lib.rs
+++ b/src/sliceview/jxl/src/lib.rs
@@ -106,3 +106,4 @@ pub fn decode(ptr: *mut u8, input_size: usize, output_size: usize) -> *const u8 
     ptr
 }
 
+


src/sliceview/jxl/src/lib.rs
```diff
@@ -62,22 +50,10 @@ pub fn height(ptr: *mut u8, input_size: usize, output_size: usize) -> i32 {
...
-
-    for keyframe_idx in 0..image.num_loaded_keyframes() {
-        let frame = match image.render_frame(keyframe_idx) {
-            Ok(frame) => frame,
-            Err(_frame) => return -3,
-        };
-
-        let stream = frame.stream();
-        return stream.height() as i32;
+    match JxlImage::builder().read(data) {
```
jbms, line 37: This is definitely a big improvement.  We could also avoid decoding the header multiple times by writing width and height to a *mut u8 buffer provided by the caller since we can only return a single value.

william-silversmith: ~thanks! couldn't we write it as a u64 composed of two i32s if we need to parse it anyway? fewer messy allocations.~ nvm I forgot Javascript only supports integers in arrays.

jbms: I think that does actually work --- I think u64 and i64 correspond to bigint in javascript

william-silversmith: It worked!

william-silversmith:
I believe the failed macos check is related to some network issue, not this PR.

fcollman:
@william-silversmith can you rebase this to see if it clears up the tests?

william-silversmith:
Nice, the tests are passing now!

PR #836, opened 2025-09-18
updating framerate adjustment parameters

fcollman:
I have noticed that the framerate downsampling has felt overly aggressive, and so i looked into whether there was a way to make it less aggressive and maintain reasonable performance.  

I added some logging statements to see what it was doing, and noticed that the actual target the code was going for was a mean of 8ms which is 125 Hz, which I think is way too aggressive.  The combination of MEAN, where an outlier number can drive things up significantly, and the overly high refresh rate was causing regular downsampling of 4 and 8 on my mac.  However, changing this to a 33 ms target (roughly a 30Hz refresh rate) means with a reasonable amount of data loaded I don't have any downsampling happening and it still feels quite snappy to me.  

I've left the logging in for now (but turned off) in order to help facilitate investigation across some use cases, but i'd advocate for changing this to give users a better experience while rotating volume rendering.

### DIFF
--- a/src/util/framerate.ts
+++ b/src/util/framerate.ts
@@ -183,8 +183,11 @@ export class DownsamplingBasedOnFrameRateCalculator {
   constructor(
     public numberOfStoredFrameDeltas: number = 10,
     private maxDownsamplingFactor: number = 8,
-    private desiredFrameTimingMs = 1000 / 60,
+    private desiredFrameTimingMs = 1000 / 30,
     private downsamplingPersistenceDurationInFrames = 15,
+  /** If true, logs factor changes and timing stats to console */
+  private enableLogging: boolean = false,
+  /** If true, always use MIN timing regardless of supplied method argument */
   ) {
     this.validateConstructorArguments();
     for (let i = 1; i <= this.maxDownsamplingFactor; i *= 2) {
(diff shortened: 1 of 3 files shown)


src/util/framerate.ts
```diff
@@ -183,8 +183,11 @@ export class DownsamplingBasedOnFrameRateCalculator {
   constructor(
     public numberOfStoredFrameDeltas: number = 10,
     private maxDownsamplingFactor: number = 8,
-    private desiredFrameTimingMs = 1000 / 60,
+    private desiredFrameTimingMs = 1000 / 30,
     private downsamplingPersistenceDurationInFrames = 15,
+    /** If true, logs factor changes and timing stats to console */
+    private enableLogging: boolean = false,
+    /** If true, always use MIN timing regardless of supplied method argument */
```
seankmartin: Was this comment related to a param that was added and then removed?

src/util/framerate.ts
```diff
@@ -183,8 +183,11 @@ export class DownsamplingBasedOnFrameRateCalculator {
   constructor(
     public numberOfStoredFrameDeltas: number = 10,
     private maxDownsamplingFactor: number = 8,
-    private desiredFrameTimingMs = 1000 / 60,
+    private desiredFrameTimingMs = 1000 / 30,
     private downsamplingPersistenceDurationInFrames = 15,
+    /** If true, logs factor changes and timing stats to console */
```
seankmartin: Could maybe make this an `@param` tag for documenting this for consistency with the other params in this constructor

src/util/framerate.ts
```diff
@@ -296,9 +299,23 @@ export class DownsamplingBasedOnFrameRateCalculator {
       Math.pow(2, Math.round(Math.log2(downsampleFactorBasedOnFramerate))),
       this.maxDownsamplingFactor,
     );
-    return this.updateMaxTrackedDownsamplingRate(
+    const applied = this.updateMaxTrackedDownsamplingRate(
       downsampleFactorBasedOnFramerate,
     );
+    if (this.enableLogging) {
+      // Only log when the applied factor actually changes compared to previous frame.
+      if ((this as any)._lastLoggedFactor !== applied) {
```
seankmartin: Perhaps `lastLoggedFactor` could be a regular class attribute to avoid the type cast to any

src/util/framerate.ts
seankmartin: Alternatively neuroglancer already uses a number of DEBUG_XYZ consts for these kind of debugging logs which we could considering using instead. Not sure if @jbms has a guideline or preference there for this

src/util/framerate.ts
fcollman: indeed! I was thinking of exploring MIN as the parameter because i was wanting to be more optimistic, but I decided that was extra complicated and probably using MEDIAN was sufficient if we wanted to be more optimistic than MEAN, but really after i realized that the target was being set to 8ms and that was the far more important adjustment to make.

src/util/framerate.ts
fcollman: sounds good.. i think the DEBUG_XYZ approach is probably better. Will implement.

src/util/framerate.ts
fcollman: agreed, fixed.

jbms:
Thanks -- can you fix the format issue and then I'm happy to merge this.

Also, thanks Sean for reviewing this!

I agree that making it user configurable could make sense --- but I wonder if it would be better to make this a per-user setting stored in browser local storage rather than in the viewer state?

fcollman:
thanks.. fixed the formatting.

@jbms what is an example of a setting that is stored in this manner? or are you suggesting a new kind of setting/parameter paridigm?

PR #849, opened 2025-10-03
add nifti float64 support

fcollman:
I've added downcasting of Float64 to Float32 to enable NIFTI float 64 support.

### DIFF
--- a/src/datasource/nifti/backend.ts
+++ b/src/datasource/nifti/backend.ts
@@ -306,8 +306,8 @@ export class NiftiVolumeChunkSource extends WithParameters(
     // If original header datatype was FLOAT64 (downcast) convert the buffer to Float32.
     if (data.header.datatypeCode === NiftiDataType.FLOAT64) {
       const src = new Float64Array(imageBuffer);
-      const dst = new Float32Array(src.length);
-      for (let i = 0; i < src.length; ++i) dst[i] = src[i];
+      let dst = new Float32Array(src.length);
+      dst = Float32Array.from(src);
       imageBuffer = dst.buffer;
     }
     await decodeRawChunk(
--- a/src/datasource/nifti/backend.ts
+++ b/src/datasource/nifti/backend.ts
@@ -303,8 +303,7 @@ export class NiftiVolumeChunkSource extends WithParameters(
     // If original header datatype was FLOAT64 (downcast) convert the buffer to Float32.
     if (data.header.datatypeCode === NiftiDataType.FLOAT64) {
       const src = new Float64Array(imageBuffer);
-      let dst = new Float32Array(src.length);
-      dst = Float32Array.from(src);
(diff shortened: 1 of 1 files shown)


src/datasource/nifti/backend.ts
```diff
@@ -295,7 +302,14 @@ export class NiftiVolumeChunkSource extends WithParameters(
       this.parameters.url,
       { signal },
     );
-    const imageBuffer = readImage(data.header, data.uncompressedData);
+    let imageBuffer = readImage(data.header, data.uncompressedData);
+    // If original header datatype was FLOAT64 (downcast) convert the buffer to Float32.
+    if (data.header.datatypeCode === NiftiDataType.FLOAT64) {
+      const src = new Float64Array(imageBuffer);
+      const dst = new Float32Array(src.length);
+      for (let i = 0; i < src.length; ++i) dst[i] = src[i];
```
jbms: Instead of looping you can just do `Float32Array.from(src)`.  Not sure if it is any faster

fcollman: done

src/datasource/nifti/backend.ts
```diff
@@ -140,6 +140,13 @@ const DATA_TYPE_CONVERSIONS = new Map([
   [NiftiDataType.INT64, { dataType: DataType.UINT64 }],
   [NiftiDataType.UINT64, { dataType: DataType.UINT64 }],
   [NiftiDataType.FLOAT32, { dataType: DataType.FLOAT32 }],
+  // Downcast unsupported float64 to float32 for visualization.
+  [
+    NiftiDataType.FLOAT64,
+    { dataType: DataType.FLOAT32, forceConvertFrom: NiftiDataType.FLOAT64 },
```
chrisj: Is `forceConvertFrom` used?

fcollman: oh agreed.. removed

src/datasource/nifti/backend.ts
```diff
@@ -295,7 +299,14 @@ export class NiftiVolumeChunkSource extends WithParameters(
       this.parameters.url,
       { signal },
     );
-    const imageBuffer = readImage(data.header, data.uncompressedData);
+    let imageBuffer = readImage(data.header, data.uncompressedData);
+    // If original header datatype was FLOAT64 (downcast) convert the buffer to Float32.
+    if (data.header.datatypeCode === NiftiDataType.FLOAT64) {
+      const src = new Float64Array(imageBuffer);
+      let dst = new Float32Array(src.length);
```
jbms: This first creates an array of zeros, then reassigns it.

Instead, `const dst = Float32Array.from(src);`

fcollman: thanks! still improving my typescript :)

PR #854, opened 2025-10-21
feat: always store volume rendering depth samples in JSON state if volume rendering is enabled

seankmartin:
Always store the number of depth samples from volume rendering in the JSON state if volume rendering is enabled. Now the JSON state should contain the number of volume rendering samples if it was either changed from the default, or never changed from the default, but volume rendering is active.

This should allow us to make future changes to the volume rendering number of samples more readily without affecting existing states that have volume rendering in use.


seankmartin:
I'm going to close this one, I think it would want a broader system about forcing parts of the state like @chrisj suggested or to have versioning of the state overall

PR #856, opened 2025-10-28
feat: define `DATA_VALUE_TYPE` and `DATA_VALUE_TYPE_IS_*` for shaders

LDeakin:
This is handy for writing generic user shaders that use `getDataValue()` etc

### DIFF
--- a/src/sliceview/image_layer_rendering.md
+++ b/src/sliceview/image_layer_rendering.md
@@ -222,15 +222,6 @@ DATA_VALUE_TYPE value = getDataValue();
   // Do something for other data types
 #endif
 ```
-The full list of data types are:
-- `float`, `vec2`, `vec3`, `vec4`
-- `uint64_t`
-- `uint32_t`
-- `uint16_t`, `uint16x2_t`
-- `uint8_t`, `uint8x2_t`, `uint8x3_t`, `uint8x4_t`
-- `int32_t`
-- `int16_t`, `int16x2_t`
-- `int8_t`, `int8x2_t`, `int8x3_t`, `int8x4_t`
 
 Note that only `float` is a builtin GLSL type. The remaining types are defined as simple structs in order to avoid ambiguity regarding the nature of the value:
 
@@ -248,6 +239,8 @@ struct uint64_t {
   highp uvec2 value;
 };
 ```
+
+Similar structures exist for signed integer types.
 
 To obtain the raw value as a float, call the `toRaw` function:
 
--- a/src/sliceview/image_layer_rendering.md
+++ b/src/sliceview/image_layer_rendering.md
@@ -210,14 +210,27 @@ may still be specified.
 The `getDataValue` function returns the nearest value without interpolation, while the
 `getInterpolatedDataValue` function uses trilinear interpolation.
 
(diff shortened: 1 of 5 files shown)


jbms:
Maybe `DATA_VALUE_TYPE` --- since `DATA_TYPE` might be expected to more correspond to the typescript `DataType`.

jbms:
Can you also add this to the documentation (https://github.com/google/neuroglancer/blob/master/src/sliceview/image_layer_rendering.md)?

src/sliceview/image_layer_rendering.md
```diff
@@ -210,6 +210,15 @@ may still be specified.
 The `getDataValue` function returns the nearest value without interpolation, while the
 `getInterpolatedDataValue` function uses trilinear interpolation.
 
+The `DATA_VALUE_TYPE` macro constant holds the return type of `getDataValue` and `getInterpolatedDataValue`. This can be used for type-generic programming. For example:
+
+```glsl
+DATA_VALUE_TYPE value = getDataValue();
+#if DATA_VALUE_TYPE == float
```
jbms: wait does that actually work?

I thought the GLSL preprocessor was similar to the C preprocessor except more limited?  Without token pasting `##` most of the "advanced C preprocessor" tricks are unavailable.

In particular, using the C preprocessor it is possible through advanced techniques to perform an equality check like you have above but I don't know that it would be possible in GLSL.

chrisj: Having `float` in the conditional errors for me, it seems to restricted to integer values.

```glsl
  #define FOO 6

  #if FOO == 6
    setColor(vec4(1.0, 0.0, 0.0, 1.0));
  #endif
  ```

Comment (LDeakin) [reply src/sliceview/image_layer_rendering.md]: Thanks both for catching that! I should have tested that basic use case. How about the proposed solution in https://github.com/google/neuroglancer/pull/856/commits/c93afc37c43412762078a23f7c503c208c3a637f?

E.g. shader

```glsl
#uicontrol invlerp normalized
void main() {
  DATA_VALUE_TYPE value = getDataValue();
  #if DATA_VALUE_TYPE_IS_UINT8_T
  if (value.value > 80u) {
    emitRGBA(vec4(1.0f, 0.0, 0.0f, 1.0f));
    return;         
  }
  #endif
  emitGrayscale(normalized());
}
```

jbms: I'm curious what your actual use case is for this.

LDeakin: The main use case for my lab is to check for a data-type dependent sentinel value for masking. Doing this at runtime in `neuroglancer` is much simpler than what we do currently, and I'd guess other people may find some utility in this.

src/sliceview/image_layer_rendering.md
```diff
@@ -210,6 +210,28 @@ may still be specified.
...
+
+```glsl
+DATA_VALUE_TYPE value = getDataValue();
+#if DATA_VALUE_TYPE_IS_FLOAT
+  // Do something specific for float data
+#else
+  // Do something for other data types
+#endif
+```
+The full list of data types are:
```
jbms: The vector types aren't actually possible here

LDeakin:
Closing as my lab is just handling this outside of neuroglancer now rather than using an ubershader

Issue #825, opened 2025-08-14
Dropdown control for shaders


stuarteberg:
In addition to other `#uicontrol` options like checkboxes and sliders, it would be great to have an option for dropdown lists.

**Bonus points:** For annotation shaders, it would be super cool if the dropdown items could be auto-populated by specifying the name of an `enum` property from the annotation source.

stuarteberg:
@chrisj Would you say this can be closed now that #909 has been merged?

seankmartin:
Closing as completed by https://github.com/google/neuroglancer/pull/909

PR #862, opened 2025-11-05
segment color shader

chrisj:
Rather than transpiling the shader to javascript, I setup a basic shader to render onto an offscreen canvas, and using gl.readPixels to get the value back into javascript. glsl-transpiler assumes a node environment, rather than modify or use polyfills, this seemed like a simpler solution. I also have a prototype for calculating multiple segment colors with one render pass, outputting a pixel per segment, I noted in the code that `get automaticLayerBarColors` would be a great place to use that.

It doesn't have much purpose yet, I need to re-add my segmentaton property shader additions, but I wanted to get feedback on if this readPixels approach is preferable.


src/layer/index.ts
```diff
@@ -364,6 +364,9 @@ export class UserLayer extends RefCounted {
   get isReady() {
     return this.loadingCounter === 0;
   }
+  isReadyWatchable: WatchableValueInterface<boolean> = new WatchableValue(
```
chrisj, line 367: segmentColorShaderControlState needs to way to differentiate between the lack of a segment property source and it not being loaded yet. I fixed it by waiting for layer.IsReady. If using isReady is the right way to handle it, I think the best fix is to convert loadingCounter and isReady to watchable/derivedwatchable values.

stuarteberg:
For whatever my opinion is worth, that all sounds great to me!

I like the idea of a dropdown.  Tangentially, I think it would also be a [great addition for other shaders, especially annotation shaders.][1] Hopefully whatever you implement here will be conceptually similar (from a user's perspective) to whatever gets implemented for other shaders (if such a feature is implemented).

[1]: https://github.com/google/neuroglancer/issues/825

fcollman:
is there a way to support stuart's ask for having a standalone string property that can be initialized with values from the shader code? or is that a seperate feature?

fcollman:
now that #909 is merged and we have a way of handling enums in shaders I think we should revisit this PR.

chrisj:
@fcollman it's in progress, considering an mvp with just support via uicontrol

PR #864, opened 2025-11-10
fix(tool_palette): display LayerTool description in tool palettes

chrisj:
Before:
Screenshot 2025-11-10 at 10 11 51 AM

After:
Screenshot 2025-11-10 at 10 12 06 AM

### DIFF
--- a/src/ui/tool.ts
+++ b/src/ui/tool.ts
@@ -105,7 +105,9 @@ export abstract class Tool<Context extends object = object> extends RefCounted {
   abstract activate(activation: ToolActivation<this>): void;
   renderInPalette(context: RefCounted): HTMLElement | undefined {
     context;
-    return undefined;
+    const content = document.createElement("div");
+    content.textContent = this.description;
+    return content;
   }
 
   abstract toJSON(): any;


src/ui/tool.ts
```diff
@@ -131,6 +131,12 @@ export abstract class LayerTool<
   get mouseState() {
     return this.layer.manager.root.layerSelectedValues.mouseState;
   }
+  renderInPalette(context: RefCounted): HTMLElement | undefined {
+    context;
+    const content = document.createElement("div");
+    content.innerHTML = this.description;
```
jbms: This should be `textContent` rather than `innerHTML` just for sanity/safety.

src/ui/tool.ts
```diff
@@ -131,6 +131,12 @@ export abstract class LayerTool<
   get mouseState() {
     return this.layer.manager.root.layerSelectedValues.mouseState;
   }
+  renderInPalette(context: RefCounted): HTMLElement | undefined {
```
jbms: I guess this could actually be defined in `Tool` rather than `LayerTool`?

chrisj: done both, thanks for checking. So now the only one that should show Loading... are LayerControlTools which return undefined when this.isLoading()

PR #868, opened 2025-12-08
feat: Adds copy view URL button to bypass browser URL truncation

neomorphic:
## Summary

  Adds a "Copy view URL to clipboard" button to the top bar that copies the complete URL including the full state fragment, bypassing browser truncation issues.

  ## Problem

  When neuroglancer state URLs become very long, browsers truncate them in the address bar. Copying the URL directly from the address bar results in an incomplete URL that cannot restore the full state when pasted into another browser

  ## Solution

  - Added a new button to the viewer top bar with a clipboard icon
  - The button programmatically constructs the complete URL from the current state and copies it to the clipboard

  ## Changes

  - `src/widget/copy_button.ts`: Added `makeCopyUrlButton()` function that uses the clipboard icon
  - `src/ui/url_hash_binding.ts`: Added `encodeStateAsFragment()` helper function and refactored existing code to use it
  - `src/viewer.ts`: Added the copy URL button to the top bar

### DIFF
--- a/src/viewer.ts
+++ b/src/viewer.ts
@@ -27,6 +27,7 @@ import {
   makeCoordinateSpace,
   TrackableCoordinateSpace,
 } from "#src/coordinate_transform.js";
+
 import { getDefaultCredentialsManager } from "#src/credentials_provider/default_manager.js";
 import type { CredentialsManager } from "#src/credentials_provider/index.js";
 import { SharedCredentialsManager } from "#src/credentials_provider/shared.js";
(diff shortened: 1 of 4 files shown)


seankmartin:
Hi @neomorphic thank you for this and sorry for getting around to it late. Could you please   add a status message based on the result of set clipboard in the `onClick` of the new copy url button? Here is a snippet for reference from `position_widget.ts` on L1071 which does a very similar thing. Seems good to go after that in my eyes

```ts
  const copyButton = makeCopyButton({
    title: "Copy position to clipboard",
    onClick: () => {
      const result = setClipboard(this.getPositionText());
      StatusMessage.showTemporaryMessage(
        result
          ? "Position copied to clipboard"
          : "Failed to copy position to clipboard",
      );
    },
  });
```

neomorphic:
Hi @seankmartin, Thanks for looking at this and for the example snippet. The status message has been added.

seankmartin (approved):
Looks good to me

src/viewer.ts
```diff
@@ -879,7 +884,29 @@ export class Viewer extends RefCounted implements ViewerState {
       );
       topRow.appendChild(button);
     }
-
+    {
+      const button = makeCopyUrlButton({
+        title: "Copy view URL to clipboard",
```
chrisj: I would remove "view" or replace with "state"

chrisj:
@neomorphic one small change requested above. This also needs a `npm run lint:fix`

other than that, good to go!

neomorphic:
Hi @chrisj, 
Thanks for the review, I have made all the suggested changes.

PR #878, opened 2026-02-02
 feat(skeleton): skeleton line width control and conversion from model units

chrisj:
continuing from: https://github.com/google/neuroglancer/pull/820

Based on this section from the skeleton documentation:
> The special vertex attribute id of "radius" may be used to indicate the radius in "stored model" units; it should have a "data_type" of "float32" and "num_components" of 1.

I've renamed the shader function to "modelToPixels"

The documentation also indicates:
> If using a "radius" attribute, the scaling applied by "transform" should be uniform

So I think it is fine to keep the calculation of the scaling factor to be based on just a single dimension. What is lacking is the ability to use the display dimension scale. I think a 3d approach to rendering skeletons is the right choice if that matters.

I will look into finding a way to add good tests for this feature

I left in the line annotation code so we can see what adding annotation support would look like, but I will strip that out of this PR

### DIFF
--- a/src/annotation/line.ts
+++ b/src/annotation/line.ts
@@ -41,7 +41,10 @@ import {
   initializeLineShader,
 } from "#src/webgl/lines.js";
 import type { ShaderBuilder, ShaderProgram } from "#src/webgl/shader.js";
-import { defineVectorArrayVertexShaderInput } from "#src/webgl/shader_lib.js";
+import {
+  defineVectorArrayVertexShaderInput,
+  glsl_nanometersToPixels,
+} from "#src/webgl/shader_lib.js";
 import { defineVertexId, VertexIdHelper } from "#src/webgl/vertex_id.js";
 
 const FULL_OBJECT_PICK_OFFSET = 0;
@@ -89,6 +92,16 @@ class RenderHelper extends AnnotationRenderHelper {
       this.defineShader(builder);
       defineLineShader(builder);
       builder.addVarying(`highp float[${rank}]`, "vModelPosition");
+      builder.addVertexCode(glsl_nanometersToPixels);
+      builder.addVertexCode(`
+float nanometersToPixels(float value) {
+  vec2 lineOffset = getLineOffset();
+  highp vec3 vertexA = projectModelVectorToSubspace(getVertexPosition0());
+  highp vec3 vertexB = projectModelVectorToSubspace(getVertexPosition1());
+  highp vec3 vertex = mix(vertexA, vertexB, lineOffset.x);
+  return nanometersToPixels(value, vertex, uProjection, uModelView, 1.0 / uLineParams.x);
+}
+`);
       builder.addVertexCode(`
 float ng_LineWidth;
 `);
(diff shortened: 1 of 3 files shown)


chrisj:
@jbms
We are planning to implement skeleton rendering using cylinders and spheres as mentioned in the previous comment.

There is still interest in scaling annotations based on physical units.

Options:
1) Create a 3d equivalent rendering mode for each annotation type. circles become spheres, lines become cylinders, polylines become cylinders and spheres, bounding boxes are ignored

2) Use the approach as in this PR, having a `modelToPixels` or similar shader function

Both options require a choice of how to deal with non-uniform transforms, the skeleton documentation specifies the transform should be uniform is a radius is given. We can scale using the transformation matrix or also add similar documentation indicating annotations should also have a uniform transform when scaling using physical sizes.

jbms:
A 3-d rendering mode would allow the cross section views to properly show cross sections of the 3-d shape, and for the 3d view to properly shade them and handle occlusion --- that would not be possible with just a `modelToPixels` function.

If the geometry just gets encoded as a new annotation type (point is already handled by ellipsoid, but there could be a new cylinder type) then transforms and proper clipping will get handled correctly automatically.  However that won't allow using shader code to determine the geometry based on user-defined properties.

Note one caveat with setting the geometry from the shader:  when using chunked annotations (e.g. with precomputed spatial index) that means Neuroglancer doesn't know the full geometry when selecting the chunks that are needed, which means e.g. a large cylinder may only get drawn if its center line intersects the viewport, even if part of it intersects the viewport.

seankmartin:
> @seankmartin I believe the plan was for Metacell to take this over and implement a cylinder+sphere approach to rendering the skeleton. I had an idea AI agents could handle it pretty well and gpt-5.4 came up with a solution that looks good so far https://cj-cap-skel-share-dot-neuroglancer-dot-seung-lab.ue.r.appspot.com/#!middleauth+https://global.daf-apis.com/nglstate/api/v1/6228802970583040
> 
> [seung-lab@6104b84#diff-31bd13c4f5980b931d888df0a45bf4303b010565bcf488e3021098006a7e5f17](https://github.com/seung-lab/neuroglancer/commit/6104b84c2347f6d6646bb6df86155d013da68dc7#diff-31bd13c4f5980b931d888df0a45bf4303b010565bcf488e3021098006a7e5f17)
> 
> https://github.com/seung-lab/neuroglancer/tree/cj-mesh-skeleton
> 
> If you haven't started work on this, I could wrap it up myself

We recently had a go at this as well, it aims to put a decent bit of the logic into the webgl side and draw very few quads where we essentially draw a bounding box of the shape in the vertex shader and then do an analytic raycast in the fragment shader to discard fragments and "form" the shape between discards and lighting. It aimed to have kind of baseline webgl raycast primitives, and then set up cylinders and spheres as two types of these. I'll need to tinker with a bit more and then can look at making a PR. Also it would ideally want to use cones instead of cylinders to handle the future case where node radius is accounted for, but started with cylinders like done here also as they are a bit simpler.

PR #909, opened 2026-03-16
feat(shader_ui_controls): add select control

chrisj:
Related to https://github.com/google/neuroglancer/issues/825

This PR adds a selectable option to shader UI controls.

Users can now define a select uicontrol in shader code with a #uicontrol directive, for example:

`#uicontrol uint myMode select(options=["one", "two", "three"], default="two")`

The value of the select is stored in the state and accessed in the shader as the string value

Example
```glsl
#uicontrol uint mode select(options=["large", "small", "default"], default="default")
void main() {
  if (mode == "large") {
    setPointMarkerSize(50.0);
  } else if (mode == "small") {
    setPointMarkerSize(2.0);
  }
  setColor(defaultColor());
}
```

[Example NG Link](https://neuroglancer--pr909-i7czmqvx.web.app/#!%7B%22dimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%2C%22position%22:%5B49166.84375%2C49150.390625%2C1694.5%5D%2C%22crossSectionScale%22:0.5856692901447939%2C%22projectionScale%22:131072%2C%22layers%22:%5B%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22toolBindings%22:%7B%22A%22:%7B%22type%22:%22shaderControl%22%2C%22control%22:%22mode%22%7D%7D%2C%22tab%22:%22rendering%22%2C%22annotations%22:%5B%7B%22point%22:%5B49149.74609375%2C49037.05078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2279a0f82872520aa6471d37c0589fcff1931dc7da%22%7D%2C%7B%22point%22:%5B49220.8046875%2C49090.0859375%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22f5fda6876e7ddb4d49d5926e869c0483b615e15b%22%7D%2C%7B%22point%22:%5B49268.84765625%2C49172.140625%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22930b95f8c109ee320ef875551cf024d957f9e90d%22%7D%2C%7B%22point%22:%5B49172.765625%2C49222.171875%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%225b152a16ddf37e05a1ddd777c88cea0973cd4a06%22%7D%2C%7B%22point%22:%5B49079.69140625%2C49161.1328125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2280623dbccc502080b45f446b5ffaeb22eebca051%22%7D%2C%7B%22point%22:%5B49066.6796875%2C49083.078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%220071719479ad78f5de3299afe8bf44b1573df4e3%22%7D%5D%2C%22shader%22:%22#uicontrol%20uint%20mode%20select%28options=%5B%5C%22large%5C%22%2C%20%5C%22small%5C%22%2C%20%5C%22default%5C%22%5D%2C%20default=%5C%22default%5C%22%29%5Cnvoid%20main%28%29%20%7B%5Cn%20%20if%20%28mode%20==%20%5C%22large%5C%22%29%20%7B%5Cn%20%20%20%20setPointMarkerSize%2850.0%29%3B%5Cn%20%20%7D%20else%20if%20%28mode%20==%20%5C%22small%5C%22%29%20%7B%5Cn%20%20%20%20setPointMarkerSize%282.0%29%3B%5Cn%20%20%7D%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%7D%5Cn%22%2C%22shaderControls%22:%7B%22mode%22:%22large%22%7D%2C%22name%22:%22annotation%22%7D%5D%2C%22selectedLayer%22:%7B%22size%22:568%2C%22visible%22:true%2C%22layer%22:%22annotation%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22size%22:568%2C%22visible%22:false%7D%7D)

<details>
<summary>Old version</summary>
This PR adds dropdown support to shader UI controls.

Users can now define a dropdown in shader code with a #uicontrol directive, for example:

`#uicontrol uint myMode dropdown(options=["one", "two", "three"], default=1)`

The value of the dropdown is stored in the state and accessed in the shader as the index

Example
```glsl
#uicontrol uint mode dropdown(options=["large", "small", "default"], default=2)
void main() {
  if (mode == 0u) {
    setPointMarkerSize(50.0);
  } else if (mode == 1u) {
    setPointMarkerSize(2.0);
  }
  setColor(defaultColor());
}
```

https://cj-shader-dropdown-share-dot-neuroglancer-dot-seung-lab.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%2C%22position%22:%5B49167.62890625%2C49151.15234375%2C1694.5%5D%2C%22crossSectionScale%22:0.5856692901447939%2C%22projectionScale%22:131072%2C%22layers%22:%5B%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22tab%22:%22rendering%22%2C%22annotations%22:%5B%7B%22point%22:%5B49149.74609375%2C49037.05078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2279a0f82872520aa6471d37c0589fcff1931dc7da%22%7D%2C%7B%22point%22:%5B49220.8046875%2C49090.0859375%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22f5fda6876e7ddb4d49d5926e869c0483b615e15b%22%7D%2C%7B%22point%22:%5B49268.84765625%2C49172.140625%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22930b95f8c109ee320ef875551cf024d957f9e90d%22%7D%2C%7B%22point%22:%5B49172.765625%2C49222.171875%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%225b152a16ddf37e05a1ddd777c88cea0973cd4a06%22%7D%2C%7B%22point%22:%5B49079.69140625%2C49161.1328125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2280623dbccc502080b45f446b5ffaeb22eebca051%22%7D%2C%7B%22point%22:%5B49066.6796875%2C49083.078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%220071719479ad78f5de3299afe8bf44b1573df4e3%22%7D%5D%2C%22shader%22:%22#uicontrol%20uint%20mode%20dropdown%28options=%5B%5C%22large%5C%22%2C%20%5C%22small%5C%22%2C%20%5C%22default%5C%22%5D%2C%20default=2%29%5Cnvoid%20main%28%29%20%7B%5Cn%20%20if%20%28mode%20==%200u%29%20%7B%5Cn%20%20%20%20setPointMarkerSize%2850.0%29%3B%5Cn%20%20%7D%20else%20if%20%28mode%20==%201u%29%20%7B%5Cn%20%20%20%20setPointMarkerSize%282.0%29%3B%5Cn%20%20%7D%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%7D%5Cn%22%2C%22name%22:%22annotation%22%7D%5D%2C%22selectedLayer%22:%7B%22size%22:568%2C%22visible%22:true%2C%22layer%22:%22annotation%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22size%22:568%2C%22visible%22:false%7D%7D

</details>


src/webgl/shader_source_preprocessing.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface ShaderStringPreprocessingResult {
+  code: string;
+  stringLiteralIds: ShaderStringLiteralIdMap;
+}
+
+export function preprocessStrings(
+  userShader: string,
+): ShaderStringPreprocessingResult {
+  const stringLiteralIds = new Map<string, number>();
+  const code = userShader.replace(/"(?:\\.|[^\\"])*"/g, (token) => {
```
lanesawyer: Could you pull this out into a variable with a good name to give a little bit more context on what the regex is doing?

src/webgl/shader_ui_controls.browser_test.ts
lanesawyer, line 1: New tests look solid!

src/webgl/shader_ui_controls.ts
```diff
@@ -375,6 +389,81 @@ function parseCheckboxDirective(
...
+    resolvedDefaultValue = options[0];
+  }
+  if (errors.length > 0) {
+    return { errors };
+  }
+  return {
+    control: {
+      type: "select",
+      valueType: "uint",
+      options: options!,
```
lanesawyer: I'm not a fan of the non-null assertion operator since it can lead to bugs down the line if things are refactored poorly, but is that commonly used in this codebase?

Not asking for a change, just trying to understand the practices here 😄

lanesawyer (commented):
Looking good! One request and one question, but I tested this by playing with the shader and it's working well.

I'm not going to be approving or requesting changes on PRs at this point, just here to start wrapping my head around the code by looking at some PRs!

src/webgl/shader_ui_controls.ts
chrisj: thanks for the feedback!

Sorry I skipped your question. Non-null assertion is accepted and used sporadically, I would say primarily in cases where typescript can't follow what is otherwise clearly correct.

There is definitely some usage in neuroglancer that could be cleaned up, here for instance

https://github.com/google/neuroglancer/blob/f88c50fb8fcbde7c01a3c4f1588890af88ba425e/src/chunk_manager/backend.ts#L362-L365

My feeling is we would prefer to avoid it unless the alternative is harder to follow.

src/widget/layer_control_select.ts
```diff
@@ -0,0 +1,34 @@
+import type { UserLayer } from "#src/layer/index.js";
+import type { WatchableValueInterface } from "#src/trackable_value.js";
+import type { LayerControlFactory } from "#src/widget/layer_control.js";
+
+export function selectLayerControl<LayerType extends UserLayer>(
```
seankmartin: generally looks good! Only comment I have is that the tool which shows on keypress related to the select is not very easy to use. I can't seem to manually change the dropdown by clicking, and there is no bind like shift+wheel or similar to move through the options from the tool

seankmartin (approved):
This looks good to me and the string pre-processing is interesting in general. Could imagine it later being expanded to help enums be compared by enum label instead of enum value in the shader, and don't see anything in the setup that would prevent that in the future

lanesawyer (approved):
Updates look good, things are working for me outside of my USB mouse not being able to scroll to change it while the hotkey is held down (but my laptop's touchpad _does_ work). Very strange.

But also an issue with other hotkeys so not this PR's problem 😄

Thanks for answering questions and helping me orient myself as I dive deeper into this codebase!

jbms:
This is fairly elegant in how it effectively just allows GLSL to use arbitrary string literals via interning.

A few comments:

- The numeric value assigned to string literals should be wrapped in a struct, e.g. named stringliteral_t (or perhaps just `string_t`), so that they don't accidentally get treated as numbers.   The actual number can be in the `value` member.  Equality comparison will still work.  This will discourage users directly accessing the numeric value, since the numeric value is not stable.  The downside is that you won't be able to use string literals for `switch` cases unless we specifically make it work.
- Since the type name is included in the `#uicontrol` directive, the type name does need to be reasonably short and easy to remember.  So that could favor `string_t` over just `stringliteral_t`.
- Arguably the `select` control should allow either just a plain list of values, as in the current PR, as well as a mapping of label/description -> value, and the value should be allowed to be stringliteral_t, as in the current PR, but also uint, float, etc.  The select control could show `label (value)` if there is a separate label.
- This approach is syntactically the lightest weight, which is nice, and works for cases where we don't need a stable mapping to a numeric value.  However, we should consider how to handle cases where we do need a stable numeric value, e.g. existing precomputed enum properties.  For those the most common situation would probably be comparing the enum numeric value to a label given as a string, which we could handle by making `prop_XXX("string")` return the numeric value associated with the string.  That function could be defined as a "real" GLSL+strings function just using the string interning support from this PR and would presumably be optimized to a constant by the shader compiler anyway; the only downside is that an invalid string wouldn't get caught at compile time and would instead just have to get mapped to an arbitrary value.  Alternatively we could add special preprocessing support for that syntax so that we could catch such errors at compile time but then the generic string interning from this PR isn't being used.

I think the key question is: for what uses cases is the generic string interning from this PR a better solution than context-specific string conversion, e.g. `prop_XXX("string")`?  We could have both but if the generic string interning is only used for `select` I'm not sure it would be worth it.

chrisj:
[original link updated for string_t (just replaced uint with string_t)](https://neuroglancer--pr909-i7czmqvx.web.app/#!%7B%22dimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%2C%22position%22:%5B49166.84375%2C49150.390625%2C1694.5%5D%2C%22crossSectionScale%22:0.5856692901447939%2C%22projectionScale%22:131072%2C%22layers%22:%5B%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22toolBindings%22:%7B%22A%22:%7B%22type%22:%22shaderControl%22%2C%22control%22:%22mode%22%7D%7D%2C%22tab%22:%22rendering%22%2C%22annotations%22:%5B%7B%22point%22:%5B49149.74609375%2C49037.05078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2279a0f82872520aa6471d37c0589fcff1931dc7da%22%7D%2C%7B%22point%22:%5B49220.8046875%2C49090.0859375%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22f5fda6876e7ddb4d49d5926e869c0483b615e15b%22%7D%2C%7B%22point%22:%5B49268.84765625%2C49172.140625%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22930b95f8c109ee320ef875551cf024d957f9e90d%22%7D%2C%7B%22point%22:%5B49172.765625%2C49222.171875%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%225b152a16ddf37e05a1ddd777c88cea0973cd4a06%22%7D%2C%7B%22point%22:%5B49079.69140625%2C49161.1328125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2280623dbccc502080b45f446b5ffaeb22eebca051%22%7D%2C%7B%22point%22:%5B49066.6796875%2C49083.078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%220071719479ad78f5de3299afe8bf44b1573df4e3%22%7D%5D%2C%22shader%22:%22#uicontrol%20string_t%20mode%20select%28options=%5B%5C%22large%5C%22%2C%20%5C%22small%5C%22%2C%20%5C%22default%5C%22%5D%2C%20default=%5C%22default%5C%22%29%5Cnvoid%20main%28%29%20%7B%5Cn%20%20if%20%28mode%20==%20%5C%22large%5C%22%29%20%7B%5Cn%20%20%20%20setPointMarkerSize%2850.0%29%3B%5Cn%20%20%7D%20else%20if%20%28mode%20==%20%5C%22small%5C%22%29%20%7B%5Cn%20%20%20%20setPointMarkerSize%282.0%29%3B%5Cn%20%20%7D%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%7D%5Cn%22%2C%22shaderControls%22:%7B%22mode%22:%22large%22%7D%2C%22name%22:%22annotation%22%7D%5D%2C%22selectedLayer%22:%7B%22size%22:568%2C%22visible%22:true%2C%22layer%22:%22annotation%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22size%22:568%2C%22visible%22:false%7D%7D)
[float map](https://neuroglancer--pr909-i7czmqvx.web.app/#!%7B%22dimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%2C%22position%22:%5B49166.4609375%2C49149.97265625%2C1694.5%5D%2C%22crossSectionScale%22:0.5856692901447939%2C%22projectionScale%22:131072%2C%22layers%22:%5B%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22tab%22:%22rendering%22%2C%22annotations%22:%5B%7B%22point%22:%5B49149.74609375%2C49037.05078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2279a0f82872520aa6471d37c0589fcff1931dc7da%22%7D%2C%7B%22point%22:%5B49220.8046875%2C49090.0859375%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22f5fda6876e7ddb4d49d5926e869c0483b615e15b%22%7D%2C%7B%22point%22:%5B49268.84765625%2C49172.140625%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22930b95f8c109ee320ef875551cf024d957f9e90d%22%7D%2C%7B%22point%22:%5B49172.765625%2C49222.171875%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%225b152a16ddf37e05a1ddd777c88cea0973cd4a06%22%7D%2C%7B%22point%22:%5B49079.69140625%2C49161.1328125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2280623dbccc502080b45f446b5ffaeb22eebca051%22%7D%2C%7B%22point%22:%5B49066.6796875%2C49083.078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%220071719479ad78f5de3299afe8bf44b1573df4e3%22%7D%5D%2C%22shader%22:%22#uicontrol%20float%20markerSize%20select%28options=%7B%5C%22Large%5C%22:%2050.0%2C%20%5C%22Small%5C%22:%202.5%2C%20%5C%22Default%5C%22:%201.0%7D%2C%20default=2.5%29%5Cnvoid%20main%28%29%20%7B%5Cn%20%20setPointMarkerSize%28markerSize%29%3B%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%7D%5Cn%22%2C%22name%22:%22annotation%22%7D%5D%2C%22selectedLayer%22:%7B%22size%22:568%2C%22visible%22:true%2C%22layer%22:%22annotation%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22size%22:568%2C%22visible%22:false%7D%7D)

>  That function could be defined as a "real" GLSL+strings function just using the string interning support from this PR and would presumably be optimized to a constant by the shader compiler anyway; the only downside is that an invalid string wouldn't get caught at compile time and would instead just have to get mapped to an arbitrary value

I was imagining having the string map could be used here to improve the prop function. I will give this a try.

Regardless I think this raw string support has value for string segment properties where I would do something like

`if (prop("celltype") == "Mi1")`

rather than the two prior ideas we had:
`if (stringPropertyEquals("celltype", "Mi1"))`
`if (prop("celltype") == propval("celltype", "Mi1"))`

jbms:
For these enum values there is a small fixed set of strings but for string segment properties or annotation properties, it would be nice to be able to upload the data (i.e. some encoding of the string properties) into GPU memory, and furthermore for annotations to do it chunkwise without knowing up front what all the possible values are, and not having to re-upload it frequently.

That essentially means we have to encode the strings using some sort of hash function.  To be confident that there will be no collisions we would need 128 bits, but maybe we'd be okay with a 64-bit hash and still ignore the possibility of collisions.  That would basically mean making string_t 64 or 128 bits.  My initial idea was that we could use a smaller size, like 64 bit or 32 bit even, but change the hash seed value if we find a collision and then reupload all of the data.  However that would mean storing a hash table in memory in order to check for collisions and that would be an annoying amount of overhead.  Therefore we probably should just go with either 64-bit or 128-bit string_t and ignore collisions.  With a 64-bit hash, there is a 50% chance of collisions among a set of 2^32 items.

jbms:
Note: Switching to using a hash code rather than sequential numbers can be done later without breaking anything, so that decision doesn't have to block this PR.

chrisj:
@jbms are you thinking of using a hash to allow something like colorHash on a segment string property? I do think that has value.

jbms:
The hash would just be to allow equality comparison on string_t values without needing to maintain an explicit hash map of all possible values.  Basically to compare if two strings are equal, you can just compare if their hashes are equal if the chance of collision is sufficiently low.  I can see that it would also be valuable to do a colorHash on a string property, which could certainly be done by applying an additional hash function to whatever numeric representation is stored in string_t, and if we use a hash function the string_t numeric representation would be stable which means the colors would be stable.  The downside is that it means if we later change the hash function used to compute the string_t values, which would otherwise be purely an implementation detail (and could involve a random seed), the colors would also change.

chrisj:
@jbms in that case I think we hold off on adding colorHash support for string_t until we settle on the value implementation

chrisj:
@jbms earlier you wrote
> Yes, default value should support both string and index

I haven't added support yet for index for a array options.

For map options, currently I am only supporting the `value` for specifying the default, I could support the string key or both.

For example:
```
// map options
#uicontrol float markerSize select(options={"Large": 50.0, "Small": 2.5, "Tiny": 1.0}, default=2.5)
#uicontrol float markerSize select(options={"Large": 50.0, "Small": 2.5, "Tiny": 1.0}, default="small")

// array options
#uicontrol float markerSize select(options=["Large", "Small", "Tiny"], default="Small")
#uicontrol float markerSize select(options=["Large", "Small", "Tiny"], default=0)
```

Do we want support for all of these methods of specifying the default value?

@seankmartin what are your thoughts?

jbms:
> @jbms earlier you wrote
> 
> > Yes, default value should support both string and index
> 
> I haven't added support yet for index for a array options.
> 
> For map options, currently I am only supporting the `value` for specifying the default, I could support the string key or both.
> 
> For example:
> 
> ```
> // map options
> #uicontrol float markerSize select(options={"Large": 50.0, "Small": 2.5, "Tiny": 1.0}, default=2.5)
> #uicontrol float markerSize select(options={"Large": 50.0, "Small": 2.5, "Tiny": 1.0}, default="small")

If we allow both it is ambiguous if the key and value have the same type.  But maybe it is nice to allow it when it is unambiguous.

> 
> // array options
> #uicontrol float markerSize select(options=["Large", "Small", "Tiny"], default="Small")
> #uicontrol float markerSize select(options=["Large", "Small", "Tiny"], default=0)

Actually I'm now thinking that for these the type should be specified as string_t or whatever we use as the string type, not float, and the default would also have to be a string.

> ```
> 
> Do we want support for all of these methods of specifying the default value?
> 
> @seankmartin what are your thoughts?

chrisj:
@jbms thanks for the response, yeah I forgot to fix the type for that example

The ambiguous situation would be
```
#uicontrol string_t mode select(options={"production": "prodv1", "staging": "testv3"}, default="testv3")
```

For all the others, we should pick either the key or the value to not deal with edge cases. We can easily support either for the numeric value types but it might be good to just be consistent so the string string case doesn't surprise. In which case we would use value so it matches the return type?

seankmartin:
> @jbms thanks for the response, yeah I forgot to fix the type for that example
> 
> The ambiguous situation would be
> 
> ```
> #uicontrol string_t mode select(options={"production": "prodv1", "staging": "testv3"}, default="testv3")
> ```
> 
> For all the others, we should pick either the key or the value to not deal with edge cases. We can easily support either for the numeric value types but it might be good to just be consistent so the string string case doesn't surprise. In which case we would use value so it matches the return type?

Here I'd also lean towards using value to match the return type

chrisj:
@seankmartin that matches the current behavior of the PR. I added documentation (though not yet in the user guide) and additionally added a restriction that select option values must be unique

I think it is ready to be merged

seankmartin (approved):
Thanks for the feature!

stuarteberg:
🎉

PR #946, opened 2026-04-02
fix: Update S3 prefix listing to work with ListObjectsV2 API 

dxenes1:
Got a fix here for the s3 autocomplete. If this has been fixed or addressed in another PR or issue, please disregard. Thanks for your review!

**Summary of issue** 
I noticed for the past couple months that S3 URIs have not been completing in the source input box in Chrome/Safari/Firefox. Should be reproducible in any [mainline Neuroglancer distribution]((https://neuroglancer.bossdb.io/)). (Try `s3://bossdb-open-data/`) 

Google cloud bucket URIs were autocompleting. I assumed it was an issue with the listing operation itself but the XML response looked correct to me (properly showing all prefixes delimited by a `/`), so I think I tracked the issue with the XML parser in the S3 listing script here. 

**What Changed**

- File changed: src/kvstore/s3/list.ts
- Updated root element validation to use documentElement.localName instead of tagName.
- Replaced XPath-based extraction with namespace-aware DOM traversal via getElementsByTagNameNS for:
    - CommonPrefixes > Prefix (directories)
    - Contents > Key (entries)

Technically we might not need `Keys` or individual objects. 
**Impact**
S3-backed datasource URL autocomplete now surfaces prefixes/files as expected for namespace-qualified ListObjectsV2 XML responses.

**Testing**
- Ran: `npm test -- tests/kvstore/s3.spec.ts`
    - Passed
- Manually tested S3 and GCP URIs for autocomplete on Chrome/Firefox/Safari

This change was implemented with assistance from Codex (AI coding assistant). CLA has been signed.

### DIFF
--- a/src/kvstore/s3/list.ts
+++ b/src/kvstore/s3/list.ts
@@ -28,7 +28,7 @@ const EXPECTED_XML_NAMESPACE_URIS = [
 function isValidListObjectsResponse(documentElement: Element): boolean {
   return (
     EXPECTED_XML_NAMESPACE_URIS.includes(documentElement.namespaceURI!) &&
-    documentElement.tagName === "ListBucketResult"
+    documentElement.localName === "ListBucketResult"
   );
 }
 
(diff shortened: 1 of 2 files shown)


src/kvstore/s3/list.ts
```diff
@@ -76,35 +76,39 @@ export async function getS3BucketListing(
...
+      namespaceURI,
+      "CommonPrefixes",
+    );
+    for (let i = 0; i < commonPrefixes.length; ++i) {
+      const commonPrefix = commonPrefixes.item(i)!;
+      const prefixNodes = commonPrefix.getElementsByTagNameNS(
+        namespaceURI,
+        "Prefix",
+      );
+      for (let j = 0; j < prefixNodes.length; ++j) {
```
chrisj, line 90: these for loops would be a bit cleaner as for..of

src/kvstore/s3/list.ts
```diff
@@ -76,35 +76,39 @@ export async function getS3BucketListing(
...
+        let name = prefixNodes.item(j)!.textContent;
+        if (name === null) continue;
```
chrisj, line 92: I don't think null is a possibility because these are `Element` not `Node` if we trust lib.dom.d.ts

chrisj (approved):
Thanks @dxenes1! I have a few suggestions but they are really about the prior code so I'm good with merging this PR

seankmartin (approved):
thanks for the fix!

PR #954, opened 2026-04-14
fix(status): limit mousedown focus suppression to credential prompts

chrisj:
This prior fix: https://github.com/google/neuroglancer/commit/9a20ef0e6cf791b9068c653348fb3a8ed0017b3a

prevents users from being able to click on select input elements that are in status messages. In the below link, if you press shift+a, the blending tool will become active and the select element won't respond to a click. I altered the prior fix to only preventDefault on mousedown for the interactive credentials provider status message.

https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B17213.5%2C19862.5%2C20697.5%5D%2C%22crossSectionScale%22:1%2C%22projectionScale%22:65536%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22gs://neuroglancer-janelia-flyem-hemibrain/emdata/clahe_yz/jpeg/%7Cneuroglancer-precomputed:%22%2C%22toolBindings%22:%7B%22A%22:%22blend%22%7D%2C%22tab%22:%22rendering%22%2C%22name%22:%22jpeg%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22jpeg%22%7D%2C%22layout%22:%22xy%22%7D

### DIFF
--- a/src/credentials_provider/interactive_credentials_provider.ts
+++ b/src/credentials_provider/interactive_credentials_provider.ts
@@ -28,6 +28,7 @@ export function getCredentialsWithStatus<Token>(
 ): Promise<Token> {
   const { requestDescription = "login" } = options;
   const status = new StatusMessage(/*delay=*/ true);
+  status.setPreventFocusChangeOnMouseDown(true);
   let abortController: AbortController | undefined;
   return new Promise<Token>((resolve, reject) => {
     const disposeAbortCallback = scopedAbortCallback(signal, (reason) => {
--- a/src/status.ts
+++ b/src/status.ts
@@ -28,17 +28,9 @@ export const DEFAULT_STATUS_DELAY = 200;
 
 export type Delay = boolean | number;
 
-function setupStatusContainer(container: HTMLElement) {
-  container.addEventListener("mousedown", (event) => {
-    // Prevent focus changes due to clicking on status message.
-    event.preventDefault();
-  });
-}
-
 function getStatusContainer() {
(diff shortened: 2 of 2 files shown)


seankmartin:
From trying this out it causes the tool to need to be closed manually by repressing the keybind, instead of the tool disappearing on keybind release. And also the items in the dropdown are often moved through at first when the dropdown is clicked on because of holding down the keybind. Is it the same behaviour for you? Still, that's likely an improvement but it does seem a bit confusing if so.
Maybe the tools with dropdowns should become toggle on/off on keypress instead of holding? Not sure

chrisj:
@seankmartin I tried to find a way to listen for the user to stop interacting with the select element but couldn't find a way. Having it be a toggle tool could work but I don't like how it would act differently to other similar tools.

I think the only way to get ideal behavior is a custom select element but it seems too much for a minor problem. If we prefer the current behavior, we can drop this PR, come back to it later.

seankmartin (approved):
Thanks for the PR Chris, I think it's an improvement and agree although not the perfect solution, for the minor issue it's not worth the further complexity

jbms:
Looks good to me, thanks.

Issue #952, opened 2026-04-09
OME-Zarr v0.6 affine transforms: source transformation matrix wrong in case of anisotropic spacing (e.g. rigid to shear)


m-albert:
## Bug description

I was testing the functionality introduced in https://github.com/google/neuroglancer/pull/876 and found the following: When loading OME-Zarr v0.6 multiscale data with **anisotropic voxel spacing** and a **rotation** in the multiscale-level `coordinateTransformations`, sources in the viewer obtain an affine transformation with a shearing component, despite the metadata containing a pure rigid transform.

## Environment

- Tested with neuroglancer master branch
- OME-Zarr v0.6 metadata
- Data example: anisotropic multi-view light sheet dataset with rotation transforms

PS: Sorry for the many edits, I didn't have my coding agent under control and it had opened the issue prematurely 😅. Now the issue is ready!

<details>
  <summary>Debugging comments by LLM</summary>

  Put the hidden content here.

  - bullet points
  - code blocks
  - images

## Root cause

In `src/datasource/zarr/ome.ts`, the `parseOmeMultiscale` function computes `baseTransformScaled` by dividing each **row** `i` by `baseScales[i]` (the norm of **column** `i`):

```typescript
for (let i = 0; i < rank; ++i) {
  for (let j = 0; j <= rank; ++j) {
    baseTransformScaled[j * (rank + 1) + i] /= baseScales[i];
  }
}
```

This is equivalent to left-multiplying by a diagonal matrix:

`baseTransformScaled = diag(1/baseScales) × baseTransform`

For a combined transform `baseTransform = R × diag(spacing)` (rotation × anisotropic scale), this gives:

`baseTransformScaled = diag(1/baseScales) × R × diag(spacing)`

When spacing is **anisotropic**, `baseScales` (the column norms) equal the spacing values, so:

`baseTransformScaled = diag(1/spacing) × R × diag(spacing)`

This **similarity conjugation** is NOT orthogonal — it introduces visual shearing.

## Mathematical demonstration

Consider a 3D volume with spacing `[z=2.0, y=0.7311, x=0.7311]` and a 135° rotation around z:

```
R = [[-0.7071, -0.7071,  0,  0],
     [ 0.7071, -0.7071,  0,  0],
     [ 0,       0,       1,  0],
     [ 0,       0,       0,  1]]

baseTransform = R × diag(2.0, 0.7311, 0.7311, 1) =
    [[-1.4142, -0.5170,  0,  0],
[...]

PR #957, opened 2026-04-15
Fix: ome zarr 0.6 anisotropic scale handling with non-diagonal transforms

seankmartin:
See https://github.com/google/neuroglancer/issues/952. This should fix the user facing side of the parsing, where the issue was being caused by how the scale is backed out of the base transform for the more general transforms introduced in version 0.6. See the [link from issue in client preview](https://neuroglancer--pr957-qakl2mu6.web.app/#!%7B%22dimensions%22:%7B%22z%22:%5B0.0000019999808199080306%2C%22m%22%5D%2C%22y%22:%5B4.999952049770077e-7%2C%22m%22%5D%2C%22x%22:%5B5e-7%2C%22m%22%5D%7D%2C%22position%22:%5B-14.5%2C67.5%2C83.00008392333984%5D%2C%22crossSectionScale%22:1%2C%22projectionScale%22:512%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22https://m-albert.github.io/ngff-rfc5-coordinate-transformation-examples/3d/simple/neuroglancer_example.zarr/%7Czarr3:%22%2C%22tab%22:%22source%22%2C%22name%22:%22neuroglancer_example.zarr%22%7D%5D%2C%22selectedLayer%22:%7B%22size%22:648%2C%22visible%22:true%2C%22layer%22:%22neuroglancer_example.zarr%22%7D%2C%22layout%22:%224panel-alt%22%7D). The scale is slightly different due to floating point calculations since the rotation involves a `sqrt(2)` in the numerator.
image

This changes the scale handling to divide each column of the transformation matrix by the scale for that column (excluding the translation column), and then separately divides each element in the translation column by the scale for that element. Previously this was happening per row instead of per column. For isotropic voxel spacing the row division was equivalent to the new column division and then translation scaling. But for anisotropic voxel spacing this was incorrect.

Added a Python test which sets the transform and scale in neuroglancer directly after loading in an ome-zarr with identity transform, and compares that to the result from having this scale and transform in the metadata. Also added a Typescript test to ensure the metadata shown to the user in the UI is as expected, and a test around sequence order handling - since the question on that came up in the issue. Also fixes some tests which were designed around row division being the correct way to do this.


chrisj:
The failing action isn't related to this PR

src/datasource/zarr/ome.ts
chrisj: can we move this block and the useNewBehavior check prior to the inverseBaseTransformUnscaled and baseTransformScaled as they appear to only be used for the new behavior?

The `for (const scale of scales) {` would also have to move up.

chrisj: my two comments here are about the existing state of the code. To me the names of these two matrices seem the opposite of how I see it. I would call it something like inverseBaseTransformWithScale and baseTransformWithoutScale

seankmartin: I applied it slightly differently because the half voxel offset is needed in both new and old paths, but in the new path it is needed after the inverse + baseTransform calculations as otherwise we need to copy the baseTransform again. Hopefully the boundary is now clearer but happy to refine further, thanks!

chrisj: Good catch, another way to handle it would be to use do `const baseTransform = scales[0].transform.slice();`

It looks like you are missing the applyHalfVoxelOffset for the new behavior.

seankmartin: That sounds best thanks, I was forgetting that the matrix representation here is simply `Float64Array` and not a more complex object that necessitates a `matrix.copy` call. I also replaced the `matrix.copy` call for the `baseTransformWithoutScale`

m-albert:
Thank you for this fix @seankmartin @chrisj 🎉

I confirm that things work for me as expected now!

PR #960, opened 2026-04-19
fix: load omero metadata for NGFF 0.5 data

### LINKED ISSUE #959: Bug: OMERO metadata ignored for NGFF 0.5 data
OME-NGFF's [transitional OMERO metadata](https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0079A/idr0079_images.zarr/0/) works only with Zarrv2. It is not respected for nor newer data using Zarrv3.

To reproduce, click on the Neuroglancer view icon on each validator:

**NGFF 0.4 (Zarr v2)**
https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0079A/idr0079_images.zarr/0/
This works, shows green/red channels, and the contrast limits are set.

**NGFF 0.5 (Zarr v3)**
https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0033A/BR00109990_C2.zarr/0/
This doesn't work, first channel should be blue but is red. The contrast limits get auto adjusted.

krokicki:
This PR fixes #959 which prevents [transitional OMERO metadata](https://ngff.openmicroscopy.org/specifications/0.5/index.html#omero-metadata-transitional) from being correctly loaded for NGFF 0.5. 

Previously `parseOmeMetadata` read `multiscales` from either `attrs.ome.multiscales` (v0.5+) or `attrs.multiscales` (≤v0.4), but always read `omero` from `attrs.omero`. As a result, channel rendering metadata (colors, windows, labels) was silently dropped for OME-Zarr 0.5 datasets.

This PR mirrors the existing `multiscales` pattern for `omero`.

@StephanPreibisch @mkitti

### DIFF
--- a/src/datasource/zarr/ome.ts
+++ b/src/datasource/zarr/ome.ts
@@ -729,7 +729,7 @@ export function parseOmeMetadata(
 ): OmeMetadata | undefined {
   const ome = attrs.ome;
   const multiscales = ome == undefined ? attrs.multiscales : ome.multiscales; // >0.4
-  const omero = attrs.omero;
+  const omero = ome == undefined ? attrs.omero : ome.omero; // >0.4
 
   if (!Array.isArray(multiscales)) return undefined;
   const errors: string[] = [];
--- a/src/datasource/zarr/ome.ts
+++ b/src/datasource/zarr/ome.ts
@@ -727,9 +727,9 @@ export function parseOmeMetadata(
   attrs: any,
   zarrVersion: number,
 ): OmeMetadata | undefined {
-  const ome = attrs.ome;
-  const multiscales = ome == undefined ? attrs.multiscales : ome.multiscales; // >0.4
-  const omero = ome == undefined ? attrs.omero : ome.omero; // >0.4
+  const { ome } = attrs;
+  const metadata = ome ?? attrs; // 0.5+ nests under `ome`; 0.4 keeps fields at root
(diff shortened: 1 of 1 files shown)


src/datasource/zarr/ome.ts
```diff
@@ -729,7 +729,7 @@ export function parseOmeMetadata(
 ): OmeMetadata | undefined {
   const ome = attrs.ome;
   const multiscales = ome == undefined ? attrs.multiscales : ome.multiscales; // >0.4
-  const omero = attrs.omero;
+  const omero = ome == undefined ? attrs.omero : ome.omero; // >0.4
```
chrisj: Thanks @krokicki 

I think we can clean this section now by using something like

```ts
  const {ome} = attrs;
  const metadata = attrs.ome ?? attrs; // >0.4
  const { multiscales, omero } = metadata;
```

Unfortunately version is not as simple

krokicki: @chrisj I made the change, could you please take a look?

seankmartin (approved):
Thanks for the fix!

chrisj (approved):

PR #961, opened 2026-04-19
fix: add channels correctly for auto layers

TheMooseman:
[Original Issue](https://github.com/google/neuroglancer/issues/931)

This PR aims to fix adding channels to all layer groups that reference a newly added layer. We do this by propagating all channel layers to all layers that contain the original layer.

JSON to test with:
```json
{
  "layers": [
    {
      "type": "auto",
      "source": "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0079A/idr0079_images.zarr/0",
      "name": "image"
    },
    {
      "type": "segmentation",
      "source": "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0079A/idr0079_images.zarr/0/labels/0",
      "name": "labels"
    }
  ],
  "layout": {
    "type": "row",
    "children": [
      {
        "type": "viewer",
        "layers": ["image"],
        "layout": "xy"
      },
      {
        "type": "viewer",
        "layers": ["labels"],
        "layout": "xy"
      }
    ]
  }
}
```

### DIFF
--- a/src/layer/multi_channel_setup.ts
+++ b/src/layer/multi_channel_setup.ts
@@ -379,6 +379,7 @@ export function createImageLayerAsMultiChannel(
       "Input omera metadata is not set up correctly. Colors are either missing or all close to black, and all ranges are the same or missing. Using default values for display purposes.",
     );
   }
+  const newChannelLayers: ManagedUserLayer[] = [];
   for (let i = 0; i < totalLocalChannels; i++) {
     const channelMetadata = getLayerChannelMetadata(managedLayer, i);
     const { localPosition, chanName } = getAdjustedLocalPositionAndName(i);
@@ -405,6 +406,7 @@ export function createImageLayerAsMultiChannel(
         parentIndex !== -1 ? parentIndex + i : undefined,
       );
       addedLayer = newLayer;
+      newChannelLayers.push(newLayer);
     }
     setupLayerPostCreation(
       addedLayer,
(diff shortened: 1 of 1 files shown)


TheMooseman:
@seankmartin I think this should fix the issue you posted in #931 but if I've misunderstood the problem, I'd be happy to get some more info and go back to the drawing board.

src/layer/multi_channel_setup.ts
```diff
@@ -415,8 +417,28 @@ export function createImageLayerAsMultiChannel(
       ignoreInputMetadata,
     );
   }
-  if (isTopLevelLayerListManager(managedLayer.manager)) {
-    managedLayer.manager.display.multiChannelSetupFinished.dispatch();
+  // Propagate new channel layers to all layer group subsets that contain the original managed layer
+  // The initial manager.add() above only adds to the layer's own manager and its owner
+  // Update other subsets referencing the original layer
+  if (newChannelLayers.length > 0) {
+    const root = managedLayer.manager.root;
```
chrisj: We prefer to use `const {root} = managedLayer.manager;`

src/layer/multi_channel_setup.ts
```diff
@@ -415,8 +417,28 @@ export function createImageLayerAsMultiChannel(
...
+      if (subsetIndex === -1) continue;
+      for (let j = 0; j < newChannelLayers.length; j++) {
+        subset.layerManager.addManagedLayer(
+          newChannelLayers[j].addRef(),
+          subsetIndex + 1 + j,
+        );
+      }
+    }
+  }
+  const root = managedLayer.manager.root;
```
chrisj: any reason not to move this above the previous if block and remove the other root declaration?

chrisj (commented):
Looks good other than some minor code suggestions. At some point we should add tests for layer creation.

src/layer/multi_channel_setup.ts
```diff
@@ -415,8 +417,28 @@ export function createImageLayerAsMultiChannel(
...
+  if (isTopLevelLayerListManager(root)) {
+    root.display.multiChannelSetupFinished.dispatch();
```
seankmartin: I think what is done here is more elegant than my original idea actually. So I believe what we can do is to delete the function `isTopLevelLayerListManager` altogether. And then just directly call `root.display.multiChannelSetupFinished.dispatch();` here since we know the root is always a `TopLevelLayerListSpecification` and don't need to check.

seankmartin (commented):
Thanks for the changes! I only have one small comment since the `multiChannelSetupFinished` call is changing anyway

TheMooseman:
Thanks for the feedback! I updated the code with all the suggestions, let me know if there's anything else you want me to do in this one.

src/layer/multi_channel_setup.ts
```diff
@@ -415,9 +405,26 @@ export function createImageLayerAsMultiChannel(
       ignoreInputMetadata,
     );
   }
-  if (isTopLevelLayerListManager(managedLayer.manager)) {
-    managedLayer.manager.display.multiChannelSetupFinished.dispatch();
+  // Propagate new channel layers to all layer group subsets that contain the original managed layer
+  // The initial manager.add() above only adds to the layer's own manager and its owner
+  // Update other subsets referencing the original layer
+  const { root } = managedLayer.manager.root;
```
chrisj: there is an extra root here, you can chain root indefinitely as it just returns `this`

chrisj:
Looks good other than that minor issue I mentioned above!

src/layer/multi_channel_setup.ts
TheMooseman: Oops, yeah just fixed it, thanks!

chrisj (approved):

seankmartin (approved):
thanks again for the fix!

PR #974, opened 2026-04-29
feat: add ome-zarr 0.6.dev3

seankmartin:
Changes ome-zarr version from 0.6.dev1 to recommended 0.6.dev3 instead and updates the test data

Edit, added version instead

### DIFF
--- a/src/datasource/zarr/ome.ts
+++ b/src/datasource/zarr/ome.ts
@@ -68,6 +68,7 @@ const SUPPORTED_OME_MULTISCALE_VERSIONS = new Set([
   "0.5-dev",
   "0.5",
   "0.6.dev1",
+  "0.6.dev3",
   "0.6",
 ]);
 
--- a/src/datasource/zarr/ome.ts
+++ b/src/datasource/zarr/ome.ts
@@ -67,7 +67,7 @@ const SUPPORTED_OME_MULTISCALE_VERSIONS = new Set([
   "0.4",
   "0.5-dev",
   "0.5",
-  "0.6.dev3",
+  "0.6.dev1",
   "0.6",
 ]);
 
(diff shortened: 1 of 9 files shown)


src/datasource/zarr/ome.ts
```diff
@@ -67,7 +67,7 @@ const SUPPORTED_OME_MULTISCALE_VERSIONS = new Set([
   "0.4",
   "0.5-dev",
   "0.5",
-  "0.6.dev1",
```
chrisj, line 70: I did a quick look at the emails about this but from what I see here

there is no implementation change between dev1 and dev3, is that because our implementation is correct for dev3 and not dev1?

Is there any issue with leaving dev1 as a supported version, otherwise some (if any) existing neuroglancer states would stop working?

jbms:
Yes we should probably leave 0.6.dev1 in the list.

src/datasource/zarr/ome.ts
seankmartin: One main difference between dev1 and dev3 for us is that dev3 makes it more explicit how it expects the transforms from array to physical coordinates, and physical coordinates to other coordinates to be split up. This is the [dev1](https://github.com/ome/ngff-spec/blob/e3d2f8ffbbcfb0e0906e901ec572f5c49b36328d/ngff_spec/specification.md?plain=1#L563-L564) version, but it is made more explicit in the constraints section in [dev3](https://github.com/ome/ngff-spec/blob/44ffdf7742bcef5147fb320175e9432dccc9dabf/index.md?plain=1#L472). I don't see any real issue keeping dev1 around, but if we do add some kind of validation later we technically might need to validate dev1 and dev3 differently. Since it was very early, I figured it might just be cleaner to replace dev1 by dev3 as a result.

seankmartin: What may end up happening is that we don't add any validation until 0.6, and then the devX versions are not validated

chrisj (approved):

PR #978, opened 2026-05-03
fix: improve viewer initialization error handling

noabartal:
Additional check for Google VRP report 495686553


chrisj:
@jbms should we disable auto-approved workflows for contributors?

jbms:
These tests aren't a problem in themselves, just that they waste time with their bogus vulnerability reports.

Instead we could be more wary of low-effort initial PRs just intended to get contributor status.

PR #985, opened 2026-05-08
feat(datasource/graphene): supervoxel splitting

chrisj:


src/chunk_manager/generic_file_source.ts
```diff
@@ -69,6 +69,22 @@ export class SimpleAsyncCache<Key, Value> extends ChunkSourceBase {
...
+    if (chunk !== undefined) {
+      chunk.freeSystemMemory();
+      this.chunkManager.queueManager.updateChunkState(chunk, ChunkState.QUEUED);
+    }
+  }
+
+  invalidateAll() {
+    for (const chunk of this.chunks.values()) {
+      chunk.freeSystemMemory();
+      this.chunkManager.queueManager.updateChunkState(chunk, ChunkState.QUEUED);
```
chrisj, line 84: move these two lines into an invalidateChunk method

src/kvstore/ocdbt/metadata_cache.ts
```diff
@@ -84,6 +89,67 @@ export function getManifest(
...
+// already registered by the time invalidation runs, so `memoize.get` returns
+// the existing cache instance without ever calling these stubs.
+//
+// Scope is the whole shared context: if multiple OCDBT databases are open
+// they are all flushed. Metadata is small and fast to re-fetch so this is
+// acceptable.
+export function invalidateOcdbtCaches(
+  sharedKvStoreContext: SharedKvStoreContextCounterpart,
+) {
+  const manifestCache = sharedKvStoreContext.chunkManager.memoize.get(
```
chrisj, line 104: we should add a `getIfExist` to memoize so we don't need to pass the information to construct these async cache. Do something like

```ts
for (const key of [
  MANIFEST_CACHE_KEY,
  BTREE_CACHE_KEY,
  VERSION_CACHE_KEY,
]) {
  const cache = sharedKvStoreContext.chunkManager.memoize.getIfExists(key);
  if (cache instanceof SimpleAsyncCache) {
    cache.invalidateAll();
  }
}
```

src/kvstore/ocdbt/metadata_cache.ts
```diff
@@ -84,6 +89,67 @@ export function getManifest(
...
+    () =>
+      makeIndirectDataReferenceCache(
+        sharedKvStoreContext,
+        "b+tree node",
+        decodeBtreeNode,
+      ),
+  );
+  btreeCache.invalidateAll();
+  const versionCache = sharedKvStoreContext.chunkManager.memoize.get(
+    "ocdbt:version",
```
chrisj, line 131: does `ocdbt:versionnode` need to be invalidated?

jbms: Only the manifest would need invalidation, and really should just invalidate the specific manifest, not all of them.  However, instead of using invalidation at all it may be better to add a `stalenessBound` parameter to DriverReadOptions similar to tensorstore and keep track of a timestamp for entries in the cache.

chrisj: ~@akhileshh is `invalidateOcdbtCaches` for debugging? It doesn't seem to be called~
Issue with the ~rebase~ local files

chrisj: @jbms were you thinking of something like this?
https://github.com/seung-lab/neuroglancer/commit/535f406e51f34800431abd0e47f4a61c873b0603

also by `invalidate the specific manifest, not all of them` you mean only the chunks that had had a supervoxel split inside? (I'm not sure if it might modify multiple chunks)

@akhileshh should re-create this PR as it is his work

PR #987, opened 2026-05-09
doc: Add npmjs badge to readme

kabilar:
cc @satra @ayendiki


chrisj (approved):
Thanks @kabilar, seems like a good idea!

PR #988, opened 2026-05-10
doc: Add gallery page to documentation

kabilar:
## Context

I think it would be helpful to have a dedicated page to showcase Neuroglancer scenes with example data.  So here I propose moving the `Examples` from the readme to such a page in the documentation. Hopefully others using Neuroglancer will add their data/scenes to this page in the future.

## Changes

- [x] Move examples from readme to `Gallery` page in documentation
- [x] Embed each Neuroglancer viewer as an iframe and lazy load each iframe
- [x] Test locally

cc @satra @ayendiki

### DIFF
--- a/docs/gallery/index.rst
+++ b/docs/gallery/index.rst
@@ -1,0 +1,119 @@
+.. _gallery:
+
+.. role:: raw-html(raw)
+   :format: html
+
+A selection of public datasets pre-loaded in Neuroglancer.
+Interact with each embedded viewer directly, or click **Open in new tab**.
+
+----
+
+FlyEM Hemibrain
+---------------
+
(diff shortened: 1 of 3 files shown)


chrisj:
https://neuroglancer-docs--pr988-0v2qxzaa.web.app/gallery/index.html

Looks pretty good!

chrisj:
The loading="lazy" works great for this, the only concern I have is performance but it works surprisingly well for me

docs/gallery/index.rst
```diff
@@ -0,0 +1,119 @@
+.. _gallery:
```
seankmartin, line 3: Thanks for the changes, I suggest we add `:hide-toc:` to the top of this file to hide the right hand side contents since they are unused in this and will give a bit more space to the gallery

kabilar: Good idea.  I just pushed a commit.

seankmartin (approved):
I think it looks nice, thanks for the pr!

PR #999, opened 2026-05-20
fix: Correct picking formulas

afonsobspinto:
## Summary

Fixes some picking bugs caused by inconsistent mouse-coordinate conversion and incorrect pick-buffer indexing.

### 1. Panel Border Was Applied Incorrectly

The previous code handled the panel borders inconsistently:

```ts
mouseX = ... - element.clientLeft;
mouseY = ... + element.clientTop;
```

`getBoundingClientRect()` gives coordinates relative to the panel border box. To get the mouse position relative to the panel content, both borders need to be subtracted. The X path already subtracted the left border, but the Y path added the top border instead.

With the current border size, that makes picking about 4 pixels off in Y. If the border size is expanded later, the picking error grows with it.

This is fixed by making the Y path match the X path:

```ts
mouseX = ... - element.clientLeft;
mouseY = ... - element.clientTop;
```

That gives the picking code the mouse position relative to the actual rendered panel content instead of relative to the outer border box.

### 2. Out-Of-Bounds Clearing Used Window Coordinates As Buffer Indices

The pick readback buffer is a small `pickDiameter x pickDiameter` row-major buffer. Each entry should be addressed with coordinates relative to that pick window:

```ts
relativeY * pickDiameter + relativeX
```

The previous clearing code computed absolute window coordinates `x` and `y`, then used those as though they were local buffer coordinates:

```ts
buffer[baseOffset + (y * pickDiameter + x) * stride] = 0;
```

That could clear the wrong buffer entry when the pick window overlapped the viewport edge. This probably did not show up often because it only affects picking close to panel boundaries, where picking is comparatively rare.

This is fixed by still using absolute window coordinates only to decide whether a sample is outside the viewport, but using the relative pick-window coordinates to address the readback buffer:

```ts
const x = startX + relativeX;
const y = startY + relativeY;
if (x < 0 || y < 0 || x >= viewportWidth || y >= viewportHeight) {
  buffer[baseOffset + (relativeY * pickDiameter + relativeX) * stride] = 0;
}
```

That keeps the two coordinate systems separate: `x`/`y` are viewport coordinates used for bounds checks, while `relativeX`/`relativeY` are buffer coordinates used for indexing.

### 3. 2D Picking Mixed Up Search Order And Buffer Offset

The previous 2D picking code mixed up two different indices:

- `offset`: the actual row-major pixel location inside the readback buffer.
- `i`: the search-order index inside `pickOffsetSequence`.

`pickOffsetSequence` stores buffer offsets sorted by distance from the pick-window center. The old 2D code used `offset` to compute the selected pixel position, but read the pick value using `i`:

```ts
const offset = pickOffsetSequence[i];
const pickId = data[4 * i];
```

That means the code could use the nearest sample's position while reading the pick ID from a different buffer pixel. The 3D path was already doing this correctly by reading from `data[4 * offset]`.

This is fixed by adding a shared helper, `resolveNearestPanelPickSample`, that always reads the pick data from the sampled row-major buffer offset:

```ts
const offset = pickOffsetSequence[i];
const pickValue = data[pickBaseOffset + stride * offset] ?? 0;
```

The helper returns both the sampled value and the corresponding `relativeX`/`relativeY` derived from that same `offset`. Both 2D and 3D picking now use this helper, so the selected position and selected object ID always come from the same pick-buffer sample.

(This explains why picking nodes in 2D polylines was unreliable while 3D picking worked.)

---

## Updated Test explanation in detail

The test_pick_radius was updated as I believe it was passing before incorrectly due to the 2D Picking Mixed Up Search Order And Buffer Offset problem.

`setPointMarkerSize` updates `ng_markerDiameter`, but not `ng_markerBorderWidth`:

The default point marker border width remains `1.0` unless explicitly overridden:

```ts
// src/annotation/point.ts
ng_markerDiameter = 5.0;
ng_markerBorderWidth = 1.0;
```

The 2D point draw path initializes the line shader with a `1px` feather too. Therefore we get:

```ts
// src/webgl/lines.ts
float totalLineWidth =
  lineWidthInPixels +
  2.0 * uLineParams.z +
  2.0 * borderWidth;
```

Here:

- `lineWidthInPixels` is the point marker interior size, `1`.
- `uLineParams.z` is the feather width, `1`.
- `borderWidth` is the marker border width, `1`.

```text
totalLineWidth = 1 + 2 * 1 + 2 * 1 = 5px
```

The pick buffer receives the annotation pick ID wherever the rendered annotation fragment is emitted:
That means the pick ID follows the rendered 2D marker footprint. It is not limited to the single center pixel implied by `setPointMarkerSize(1.0)`.

In the test we have a scenario similar to:

Assume the annotation center is `C`.

Because the effective pickable footprint is about `5px` wide, the pick buffer can contain the annotation pick ID in these cells:

```text
C-2   C-1    C    C+1   C+2
[ID]  [ID]  [ID]  [ID]  [ID]
```

`pick_radius = 1` means the click pixel plus one pixel on each horizontal side are checked. Along the x-axis, if the click is at `X`, Neuroglancer checks:

```text
X-1   X   X+1
```

With `check_pick(webdriver, 3)`, the click is at `C+3`.

```text
Annotation pick-ID cells:

C-2   C-1    C    C+1   C+2   C+3   C+4
[ID]  [ID]  [ID]  [ID]  [ID]  [ ]   [ ]

Click at offset 3:

                              X
                             C+3

pick_radius = 1 checks:

                       P      P      P
                      C+2    C+3    C+4
```

There is overlap at `C+2`:

```text
C+2 has [ID] and is checked by pick_radius = 1
```

Therefore:

```py
check_pick(webdriver, 3) is True (test was expecting it to be False)
```


seankmartin (approved):
I had reviewed this separately before the PR was made here. Gave it another glance over and looks good to me

src/rendered_data_panel_picking.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export function getPickDiameter(pickRadius: number): number {
+  return 1 + pickRadius * 2;
+}
+
+let _cachedPickRadius = -1;
+let _cachedPickOffsetSequence: Uint32Array | undefined;
+
+/**
+ * Sequence of offsets into C order (pickDiamater, pickDiamater) array in order of increasing
```
chrisj: there was an existing spelling error on diameter that got copied over here

jbms:
It would be great to add some end-to-end tests to verify that picking is precise and that detects all of the problems mentioned in the PR description.  That way we can be sure there are no regressions in the future.

chrisj (approved):
These are some great fixes! Thanks @seankmartin for confirming you reviewed this. I think that is a good practice we should do going forward.

I agree new tests to verify picking accuracy is preferable.

[this is a state that easily demonstrates the fix](https://neuroglancer--pr999-l7sz83n6.web.app/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%2C%22position%22:%5B366599.59375%2C90861.984375%2C2167.5%5D%2C%22crossSectionScale%22:0.08915791181328377%2C%22crossSectionDepth%22:-42949672960%2C%22projectionOrientation%22:%5B0.38193279504776%2C-0.13822562992572784%2C-0.08421601355075836%2C0.9099058508872986%5D%2C%22projectionScale%22:18243.341105960575%2C%22layers%22:%5B%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22tab%22:%22rendering%22%2C%22annotations%22:%5B%7B%22point%22:%5B366606.21875%2C90853.0625%2C2167.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22689132c4d31b9f0470d4b0b9afd8a7b50ce243e5%22%7D%5D%2C%22shader%22:%22%5Cnvoid%20main%28%29%20%7B%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%20%20setPointMarkerSize%28100.0f%29%3B%5Cn%20%20setPointMarkerBorderWidth%280.0f%29%3B%5Cn%5Cn%7D%5Cn%22%2C%22name%22:%22annotation%22%7D%2C%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22tab%22:%22annotations%22%2C%22annotationColor%22:%22#ff0000%22%2C%22annotations%22:%5B%7B%22point%22:%5B366606.21875%2C90853.0625%2C2167.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2255cdebad2c5bd57d027399c1134141e934330afb%22%7D%5D%2C%22shader%22:%22%5Cnvoid%20main%28%29%20%7B%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%20%20setPointMarkerSize%2850.0f%29%3B%5Cn%20%20setPointMarkerBorderWidth%280.0f%29%3B%5Cn%7D%5Cn%22%2C%22name%22:%22annotation1%22%7D%5D%2C%22showAxisLines%22:false%2C%22showSlices%22:false%2C%22selectedLayer%22:%7B%22col%22:1%2C%22size%22:444%2C%22visible%22:true%2C%22layer%22:%22annotation1%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22row%22:0%2C%22visible%22:false%7D%7D)

[compared to prior](https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%2C%22position%22:%5B366599.59375%2C90861.984375%2C2167.5%5D%2C%22crossSectionScale%22:0.08915791181328377%2C%22crossSectionDepth%22:-42949672960%2C%22projectionOrientation%22:%5B0.38193279504776%2C-0.13822562992572784%2C-0.08421601355075836%2C0.9099058508872986%5D%2C%22projectionScale%22:18243.341105960575%2C%22layers%22:%5B%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22tab%22:%22rendering%22%2C%22annotations%22:%5B%7B%22point%22:%5B366606.21875%2C90853.0625%2C2167.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22689132c4d31b9f0470d4b0b9afd8a7b50ce243e5%22%7D%5D%2C%22shader%22:%22%5Cnvoid%20main%28%29%20%7B%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%20%20setPointMarkerSize%28100.0f%29%3B%5Cn%20%20setPointMarkerBorderWidth%280.0f%29%3B%5Cn%5Cn%7D%5Cn%22%2C%22name%22:%22annotation%22%7D%2C%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22tab%22:%22annotations%22%2C%22annotationColor%22:%22#ff0000%22%2C%22annotations%22:%5B%7B%22point%22:%5B366606.21875%2C90853.0625%2C2167.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2255cdebad2c5bd57d027399c1134141e934330afb%22%7D%5D%2C%22shader%22:%22%5Cnvoid%20main%28%29%20%7B%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%20%20setPointMarkerSize%2850.0f%29%3B%5Cn%20%20setPointMarkerBorderWidth%280.0f%29%3B%5Cn%7D%5Cn%22%2C%22name%22:%22annotation1%22%7D%5D%2C%22showAxisLines%22:false%2C%22showSlices%22:false%2C%22selectedLayer%22:%7B%22col%22:1%2C%22size%22:444%2C%22visible%22:true%2C%22layer%22:%22annotation1%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22row%22:0%2C%22visible%22:false%7D%7D)

If you mouse over the boundary between the red and yellow annotations, you can see the accuracy issues

seankmartin:
Thanks @chrisj and @jbms, I took a crack at the simplest of these to test, which is the y-panel border fix. For the others we will have to think a bit more on them. It's a bit unfortunate that the tests have failed after this test I made, but as far as I can tell those are unrelated failures.

It might be a little bit of time before we can add the other tests for cases 2 and 3. If you want to wait until we try to have e2e tests for case 2 and 3, that's of course fine. If you'd rather merge in the meantime, I will make an issue regarding those tests.

seankmartin:
I have added some tests for these now, I'll probably need to refine them to work in GitHub actions as they do locally. Roughly the idea is:

1. Y-border fix: Very similar tests to the existing pick radius checks that only moved only x offsets. We now also move along y-offsets and combined x- and y- offsets.
2. OOB clearing: Conceptually like this, if we imagine the requested pick window is square instead of circular:

```
      OOB  OOB  OOB  OOB  OOB
      OOB  OOB  OOB  OOB  OOB
      [ ]  [ ]  [C]  [ ]  [ ]
      [X]  [ ]  [ ]  [ ]  [ ]
      [ ]  [ ]  [ ]  [ ]  [ ]
```
If the cursor was at C right below the panel top edge, and any of the OOB values contained a pickID, then instead of finding the true target X, we could find an OOB pickID. That is the purpose of clearOutOfBoundsPickData, to set all the OOB values to 0. But it wasn't doing that properly.
We test this by checking we can pick a point when it is close to the border, and then can't pick the point anymore when outside of the border.
3. Search order of pickIDs within the pick window. Previously the pickIDs in 2D were essentially checked in row major scan order along the pick window. They were meant to be searched in distance order from the cursor position.  Like this, where `X` is the cursor, `N` are near candidates, and `F` are far candidates. The search should start from `X`, but instead the search started at `F`. We test this by putting two points within the same picking buffer window and ensuring the closer point to the cursor is found.

```
            F

      N     N     N

F     N     X     N    F

      N     N     N

            F
```

PR #1010, opened 2026-05-22
chore: deploy to neuroglancer-demo.appspot.com on push to master

jbms:


chrisj:
@jbms was the appspot deployment previously covered by firebase?

and are you keeping the firebase deployment around in order to keep client links to .firebaseapp.com and .web.app deployment working?

jbms:
The firebase deployment was always independent --- I added it mainly for the PR previews and doc deployment.

And yes not sure how many people depend on the firebase instance but since it is needed for docs and PR previews anyway it doesn't add much effort to keep it.

PR #1014, opened 2026-05-25
Add colormaps and ui control

fcollman:
This PR expands the set of built in colormaps to include some of the more popular scientific colormaps and creates a new uicontrol which simply lets people select from the built in colormaps and pushes data through it.  Combining this with invlerp allows users to easily colormap data for image or annotation layers, and have users be able to try out different colormaps without looking into the neuroglancer source code to find string which defines them.

example usage
```
#uicontrol invlerp normalized
#uicontrol colormap cmap colormap(default="viridis")
void main() {
  setColor(cmap(normalized()));
}
```

image


jbms:
(1) Can we deploy the colormaps as a plain binary data file, N*256*3 bytes, and just load color maps on demand (byte range read) rather than encoding them as GLSL functions?  JavaScript would just contain the list of names.

(2) It would be nice to make the dropdown menu for selecting a colormap actually show a preview of the colormap, possibly using the existing virtual list support.

fcollman:
This does include a dropdown preview

image

I will try to refactor it into 256 value binary data that we load by name > offset in a colormap file that gets compile time built.

fcollman:
i have tried to refactor this with a binary file that contains the colormaps.  I have it downloading all of them at once, since the dropdown preview widget requires that all of them to be loaded at once it felt like it wasn't worth the complexity to load them individually lazily on demand with a range read. 

Currently the binary is checked into the repo, but also the python script that is used to generate that binary file.

I also added "oranges" colormap to the list.

fcollman:
i don't think these build failures have anythign to do with me and have to do with github going down when they ran.  Would appreciate someone retriggering all the failed builds.

chrisj:
tried re-running but it needed a fix, seems to be related to the node v24 upgrade. Not sure why it didn't start failing immediately

https://github.com/google/neuroglancer/pull/1020

src/webgl/colormaps.ts
```diff
@@ -13,30 +13,222 @@
...
+  const idx = COLORMAP_BIN_NAMES.indexOf(name);
+  if (idx < 0) return undefined;
+  return colormapDataCache.subarray(
+    idx * COLORMAP_STRIDE,
+    (idx + 1) * COLORMAP_STRIDE,
+  );
+}
+
+/**
+ * Returns a 768-byte grayscale LUT (R=G=B=i for i in 0..255). Used as a
```
jbms: It would be better to just not render it, I think, rather than render with the wrong color map.  The shader should somehow just be marked not ready.  When toggling between color maps using the ui control, ideally it would continue to display with old colormap while waiting for new one --- maybe that can be accomplished with fallback shader or something.

src/webgl/colormaps.ts
```diff
@@ -13,30 +13,222 @@
...
+  return colormapDataPromise;
+}
+
+/**
+ * Synchronous accessor: returns a 768-byte RGB view into the cached LUT for
+ * `name`, or `undefined` if the data hasn't loaded yet. Callers that need
+ * the data must either await `getColormapDataPromise()` or subscribe to
+ * `colormapDataLoaded` and retry on dispatch.
+ */
+export function getColormapBytes(
```
jbms: For preview dropdown just fetching all of them is reasonable (you could just fetch the visible ones but I don't think that is necessarily a useful optimization).  However, in most cases users will just be loading an existing view and then it is not necessary to download the whole list, which could easily become rather large as more and more colormaps are added.

src/webgl/colormaps.ts
```diff
@@ -13,30 +13,222 @@
...
+ * concatenation of N 256x3 RGB LUTs in the order of `COLORMAP_NAMES`. The
+ * file is fetched once at app start; shaders sample it through a 1D texture
+ * uniform instead of inlining polynomial approximations.
+ *
+ * Regenerate the binary with `build_tools/generate_colormaps_bin.py`.
+ */
+
+import { RefCounted } from "#src/util/disposable.js";
+import { NullarySignal } from "#src/util/signal.js";
+import COLORMAPS_BIN_URL from "#src/webgl/colormaps.bin";
```
jbms: Use "new URL" syntax as used elsewhere to avoid depending on custom packing rules.

src/webgl/colormaps.ts
```diff
@@ -13,30 +13,222 @@
...
+import type { GL } from "#src/webgl/context.js";
+import { setRawTextureParameters } from "#src/webgl/texture.js";
+
+// Full list of colormaps present in colormaps.bin, in offset order. Includes
+// back-compat-only colormaps (e.g. `jet`) that are not exposed in the user-
+// facing dropdown but are reachable via free GLSL functions like
+// `colormapJet`. MUST match generate_colormaps_bin.py.
+export const COLORMAP_BIN_NAMES = [
```
jbms: We can generate a JSON file from the generate script and import that to get the list without duplicating it here.

rspack.config.ts
```diff
@@ -53,6 +53,12 @@ export default defineConfig((env, args) => {
           resourceQuery: /raw/,
           type: "asset/source",
         },
+        // Needed for .bin imports (e.g. the bundled colormaps.bin asset);
```
jbms: We won't need this extra rule if we use `new URL` syntax.

src/webgl/colormaps.ts
fcollman: how do we get the typing?

fcollman:
I played around in the latest commit with a zarr based backing for the colormaps, to facilitate lazy loading and partial loading of the colormaps and set us up toward having an external store of colormap data that is self described versus opaque. Curious if people like this approach.

src/webgl/colormaps.ts
```diff
@@ -13,30 +13,477 @@
...
+import { NullarySignal } from "#src/util/signal.js";
+import type { GL } from "#src/webgl/context.js";
+import { setRawTextureParameters } from "#src/webgl/texture.js";
+
+// Full list of colormaps present in colormaps.zarr, in N-axis order. Includes
+// back-compat-only colormaps (e.g. `jet`) that are not exposed in the user-
+// facing dropdown but are reachable via free GLSL functions like
+// `colormapJet`. MUST match generate_colormaps_zarr.py and the
+// `attributes.colormap_names` array in colormaps.zarr/zarr.json.
+export const COLORMAP_BIN_NAMES = [
```
chrisj: rather than defining it, can we just load whatever colormaps are in the zarr file?

For display names, it looks like it's only capitalizing the first letter so something like a css `text-transform: capitalize;` should work.

chrisj:
I like the idea of zarr but I as per my comment above, I think it would better leverage zarr's capabilities if the metadata was the source of truth

jbms:
Using zarr rather than a custom binary format is kind of nice but it introduces a number of problems: 

- no longer a single file, which means we can't rely on broadly-supported `new URL` syntax and instead need bundler-specific configuration.  That means any transitive users of "neuroglancer as an npm package" need to add neuroglancer-specific configuration to their bundler configuration.  I spent a lot of time a while back to try to make "neuroglancer as an npm package" just work without any bundler-specific configuration, except for the html oauth redirect pages that do require bundler specific configuration but can be omitted if you don't need those specific authentication methods.
- Using sharded format adds an extra fetch needed to read the index, which is unnecessary if we just store it uncompressed (in which case we can just calculate byte range directly).
- Overall code size, even reusing existing zarr support in neuroglancer, ends up being larger

You could still write it as a zarr array, but as uncompressed, single-chunk, non-sharded --- meaning it will be identical to the "custom binary format" --- and then access the chunk file itself via `new URL` syntax.  Then the zarr.json file won't be used or even included in the bundle.

Arguably we should just build the colormap file as part of the build step rather than storing the generated data in the repo itself, since unlike the wasm build steps it is very fast and doesn't add extra dependencies.  We would still include the generated file in the published PyPI package (including the sdist) and published NPM package, since we already have a build step that runs to generate those packages.

fcollman:
I have refactored the zarr generation to be a single chunk file that is generated during build rather than checked into the repo, and simplified neuroglancer to access the bytes as a standalone file without use of the zarr infrastructure.  

I also removed the name redirection.  I still have this pattern i'm not totally comfortable with where the python script is generating a typescript file that contains the byte order of the colormap, and the build script first runs this python script to generate the typescript.  It's just one file with two arrays of names (the list of colormaps in the file, and the list of colormaps that should be presented in the dropdown.  

As an aside, this is to support "colormapJet" for backward compatibility, but only present "turbo" as an option to the user in the dropdown since it's really a better version of jet.  

I had to add this python script to many of the package.json hooks and make sure that uv was available for the various actions to make this approach function.

PR #1016, opened 2026-05-25
Add annotation reordering

fcollman:
Allows reordering of local annotations via alt+drag in the annotation list. Includes the reorder() method on AnnotationSource, the childrenReordered signal, drag-and-drop UI wiring, drop indicator styles, and a documentation link on the annotation help icon.

### DIFF
--- a/src/annotation/index.ts
+++ b/src/annotation/index.ts
@@ -1478,14 +1478,9 @@ export class AnnotationSource
     if (placement === "after") targetIndex += 1;
     ids.splice(targetIndex, 0, sourceId);
     // No-op if order is unchanged.
-    let changed = false;
-    let i = 0;
-    for (const id of annotationMap.keys()) {
-      if (id !== ids[i++]) {
-        changed = true;
-        break;
-      }
-    }
+    const changed = !Array.from(annotationMap.keys()).every(
+      (id, i) => id === ids[i],
+    );
     if (!changed) return false;
     const entries = ids.map(
       (id) => [id, annotationMap.get(id)!] as [AnnotationId, Annotation],
--- a/src/annotation/index.ts
+++ b/src/annotation/index.ts
@@ -1370,6 +1370,7 @@ export interface AnnotationSourceSignals {
   childUpdated: Signal<(annotation: Annotation) => void>;
   childCommitted: Signal<(annotationId: string) => void>;
   childDeleted: Signal<(annotationId: string) => void>;
+  childrenReordered: NullarySignal;
 }
 
 export class AnnotationSource
@@ -1383,6 +1384,7 @@ export class AnnotationSource
   childUpdated = new Signal<(annotation: Annotation) => void>();
   childCommitted = new Signal<(annotationId: string) => void>();
   childDeleted = new Signal<(annotationId: string) => void>();
+  childrenReordered = new NullarySignal();
 
   public pending = new Set<AnnotationId>();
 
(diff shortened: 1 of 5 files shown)


fcollman:
I think the test failure is a fluke and unrelated to my changes

src/annotation/index.ts
```diff
@@ -1454,6 +1456,49 @@ export class AnnotationSource
...
+    const ids = Array.from(annotationMap.keys());
+    const fromIndex = ids.indexOf(sourceId);
+    ids.splice(fromIndex, 1);
+    let targetIndex = ids.indexOf(targetId);
+    if (placement === "after") targetIndex += 1;
+    ids.splice(targetIndex, 0, sourceId);
+    // No-op if order is unchanged.
+    let changed = false;
+    let i = 0;
+    for (const id of annotationMap.keys()) {
```
chrisj: this loop and the outer variables can be replaced with
```ts
const changed = !annotationMap.keys().every((id, i) => {
  return id === ids[i];
});
```

src/ui/annotations.ts
```diff
@@ -504,9 +508,9 @@ export class AnnotationLayerView extends Tab {
 
     const helpIcon = makeIcon({
       title:
-        "The left icons allow you to select the type of the anotation. Color and other display settings are available in the 'Rendering' tab.",
+        "The left icons allow you to select the type of the annotation. Color and other display settings are available in the 'Rendering' tab. Click for documentation.",
       svg: svg_help,
-      clickable: false,
+      href: "https://neuroglancer-docs.web.app/user-guide/annotation.html",
```
chrisj: since this refers to the new documentation, let's make sure to merge that first

src/ui/annotations.ts
```diff
@@ -550,6 +554,153 @@ export class AnnotationLayerView extends Tab {
...
+      document.removeEventListener("mousemove", onMouseMove, true);
+      document.removeEventListener("mouseup", onMouseUp, true);
+      document.removeEventListener("keydown", onKeyDown, true);
+      dragState = undefined;
+      if (!commit || dropRow === undefined || dropPlacement === undefined) {
+        return;
+      }
+      const targetId = dropRow.dataset.annotationId;
+      if (targetId === undefined || targetId === sourceId) return;
+      const source = state.source;
```
chrisj: can be written as `const {source} = state;`

src/ui/annotations.ts
```diff
@@ -550,6 +554,153 @@ export class AnnotationLayerView extends Tab {
...
+      dragState = {
+        sourceId,
+        state,
+        sourceRow: row,
+        dropRow: undefined,
+        dropPlacement: undefined,
+      };
+      document.addEventListener("mousemove", onMouseMove, true);
+      document.addEventListener("mouseup", onMouseUp, true);
+      document.addEventListener("keydown", onKeyDown, true);
```
chrisj, line 701: We should add a `window.addEventListener("blur", onBlur, true);`

with
```ts
    const onBlur = () => {
      finish(false);
    };
```

This will cancel the re-order if the window loses focus which you can trigger with a cmd+tab on mac. More important for firefox

src/ui/annotations.ts
```diff
@@ -550,6 +554,153 @@ export class AnnotationLayerView extends Tab {
...
+
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "Escape") {
+        event.preventDefault();
+        finish(false);
+      }
+    };
+
+    const onReorderStart = (event: ActionEvent<MouseEvent>) => {
+      const row = (event.detail.target as HTMLElement | null)?.closest(
```
chrisj, line 676: this errors on firefox if you move outside the viewport, doesn't appear to cause any issue other than an unnecessary error message in the console.

Something like this fixes it:
```ts
      const eventTarget = event.target;
      const target =
        eventTarget instanceof Element
          ? (eventTarget.closest(
              ".neuroglancer-annotation-list-entry",
            ) as HTMLElement | null)
          : null;
```

chrisj (commented):
Think it is ready to merge after this mostly firefox related changes

seankmartin (approved):
Thanks, this looks good to me once Chris's comments about firefox fixes are in place

chrisj (approved):

PR #1018, opened 2026-05-28
auto hide controls

fcollman:
Based on a feature request from @stuarteberg  this allows the option for the uicontrols which are not accessed based on the current compilation of the GLSL code to be hidden from the user.  The principle use case is having check boxes of rendering modes, and unselecting those boxes will cause options be not relevant to the user.  Ideally this will work with #909 as well once it gets in, but if we adopt it this should be modified to also support the dropdown. 

@jbms and I discussed that long term we should move to some GLSL > AST parser, which will be required for moving to WebGPU anyway, but also will allow for more robust static analysis of the shader language. 

presently for example nested checkboxes will not disappear, even if the nested checkbox is not accessed becuase checkboxes are implemented as uniforms.  Also, if code paths are defined based on run-time evaluation of the UI control (such as the location of the slider) then this method will also fail.

### DIFF
--- a/src/layer/annotation/index.ts
+++ b/src/layer/annotation/index.ts
@@ -156,6 +156,7 @@ const LINKED_SEGMENTATION_LAYER_JSON_KEY = "linkedSegmentationLayer";
 const FILTER_BY_SEGMENTATION_JSON_KEY = "filterBySegmentation";
 const IGNORE_NULL_SEGMENT_FILTER_JSON_KEY = "ignoreNullSegmentFilter";
 const CODE_VISIBLE_KEY = "codeVisible";
+const HIDE_INACTIVE_CONTROLS_JSON_KEY = "hideInactiveControls";
 
 class LinkedSegmentationLayers extends RefCounted {
   changed = new NullarySignal();
@@ -401,6 +402,7 @@ const Base = UserLayerWithAnnotationsMixin(UserLayer);
 export class AnnotationUserLayer extends Base {
   localAnnotations: LocalAnnotationSource | undefined;
   codeVisible = new TrackableBoolean(true);
+  hideInactiveControls = new TrackableBoolean(false);
   private localAnnotationProperties: WatchableValue<AnnotationPropertySpec[]> =
     new WatchableValue([]);
   private localAnnotationRelationships: string[];
@@ -430,6 +432,7 @@ export class AnnotationUserLayer extends Base {
       this.specificationChanged.dispatch,
     );
     this.codeVisible.changed.add(this.specificationChanged.dispatch);
+    this.hideInactiveControls.changed.add(this.specificationChanged.dispatch);
     this.annotationDisplayState.ignoreNullSegmentFilter.changed.add(
       this.specificationChanged.dispatch,
     );
@@ -456,6 +459,9 @@ export class AnnotationUserLayer extends Base {
     super.restoreState(specification);
     this.linkedSegmentationLayers.restoreState(specification);
     this.codeVisible.restoreState(specification[CODE_VISIBLE_KEY]);
+    this.hideInactiveControls.restoreState(
+      specification[HIDE_INACTIVE_CONTROLS_JSON_KEY],
+    );
     this.localAnnotationsJson = specification[ANNOTATIONS_JSON_KEY];
     const properties = verifyOptionalObjectProperty(
       specification,
@@ -737,6 +743,7 @@ export class AnnotationUserLayer extends Base {
     x[CROSS_SECTION_RENDER_SCALE_JSON_KEY] =
       this.annotationCrossSectionRenderScaleTarget.toJSON();
     x[CODE_VISIBLE_KEY] = this.codeVisible.toJSON();
+    x[HIDE_INACTIVE_CONTROLS_JSON_KEY] = this.hideInactiveControls.toJSON();
     x[PROJECTION_RENDER_SCALE_JSON_KEY] =
       this.annotationProjectionRenderScaleTarget.toJSON();
     if (this.localAnnotations !== undefined) {
@@ -858,7 +865,10 @@ class RenderingOptionsTab extends Tab {
           layer.annotationDisplayState.shaderControls,
           this.layer.manager.root.display,
           this.layer,
-          { visibility: this.visibility },
+          {
+            visibility: this.visibility,
+            hideInactiveControls: layer.hideInactiveControls,
+          },
         ),
       ).element,
     );
--- a/src/layer/annotation/index.ts
+++ b/src/layer/annotation/index.ts
@@ -156,7 +156,7 @@ const LINKED_SEGMENTATION_LAYER_JSON_KEY = "linkedSegmentationLayer";
 const FILTER_BY_SEGMENTATION_JSON_KEY = "filterBySegmentation";
 const IGNORE_NULL_SEGMENT_FILTER_JSON_KEY = "ignoreNullSegmentFilter";
 const CODE_VISIBLE_KEY = "codeVisible";
-const HIDE_INACTIVE_CONTROLS_JSON_KEY = "hideInactiveControls";
+const hide_inactive_shader_controls_JSON_KEY = "hideInactiveShaderControls";
 
 class LinkedSegmentationLayers extends RefCounted {
   changed = new NullarySignal();
@@ -402,7 +402,7 @@ const Base = UserLayerWithAnnotationsMixin(UserLayer);
 export class AnnotationUserLayer extends Base {
   localAnnotations: LocalAnnotationSource | undefined;
   codeVisible = new TrackableBoolean(true);
-  hideInactiveControls = new TrackableBoolean(false);
+  hideInactiveShaderControls = new TrackableBoolean(false);
   private localAnnotationProperties: WatchableValue<AnnotationPropertySpec[]> =
     new WatchableValue([]);
   private localAnnotationRelationships: string[];
@@ -432,7 +432,7 @@ export class AnnotationUserLayer extends Base {
       this.specificationChanged.dispatch,
     );
     this.codeVisible.changed.add(this.specificationChanged.dispatch);
-    this.hideInactiveControls.changed.add(this.specificationChanged.dispatch);
+    this.hideInactiveShaderControls.changed.add(this.specificationChanged.dispatch);
     this.annotationDisplayState.ignoreNullSegmentFilter.changed.add(
       this.specificationChanged.dispatch,
     );
@@ -459,8 +459,8 @@ export class AnnotationUserLayer extends Base {
     super.restoreState(specification);
     this.linkedSegmentationLayers.restoreState(specification);
     this.codeVisible.restoreState(specification[CODE_VISIBLE_KEY]);
-    this.hideInactiveControls.restoreState(
-      specification[HIDE_INACTIVE_CONTROLS_JSON_KEY],
+    this.hideInactiveShaderControls.restoreState(
+      specification[hide_inactive_shader_controls_JSON_KEY],
     );
     this.localAnnotationsJson = specification[ANNOTATIONS_JSON_KEY];
     const properties = verifyOptionalObjectProperty(
@@ -743,7 +743,7 @@ export class AnnotationUserLayer extends Base {
     x[CROSS_SECTION_RENDER_SCALE_JSON_KEY] =
       this.annotationCrossSectionRenderScaleTarget.toJSON();
     x[CODE_VISIBLE_KEY] = this.codeVisible.toJSON();
-    x[HIDE_INACTIVE_CONTROLS_JSON_KEY] = this.hideInactiveControls.toJSON();
+    x[hide_inactive_shader_controls_JSON_KEY] = this.hideInactiveShaderControls.toJSON();
     x[PROJECTION_RENDER_SCALE_JSON_KEY] =
       this.annotationProjectionRenderScaleTarget.toJSON();
     if (this.localAnnotations !== undefined) {
@@ -867,7 +867,7 @@ class RenderingOptionsTab extends Tab {
           this.layer,
           {
             visibility: this.visibility,
-            hideInactiveControls: layer.hideInactiveControls,
+            hideInactiveShaderControls: layer.hideInactiveShaderControls,
           },
         ),
       ).element,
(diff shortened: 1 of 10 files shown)


seankmartin:
I will try take a more detailed look again, but overall I agree with later moving to an AST representation would be great, for now though relying on result of the compiling and linking the GLSL seems reasonable

src/webgl/shader_ui_controls.ts
```diff
@@ -1313,6 +1318,11 @@ export function getFallbackBuilderState(
   };
 }
 
+// JSON key under which `hideInactiveControls` is serialized inside the
+// shaderControlState object. The double-underscore prefix avoids collisions
+// with #uicontrol names (which by convention start with a letter).
+const HIDE_INACTIVE_JSON_KEY = "__hideInactiveControls";
```
chrisj: this is convenient but we should decide if this is a good convention for our state. Alternatively, we can make this a direct property on the layer instead.

chrisj:
[Example of the PR](https://neuroglancer--pr1018-rdsxtj9a.web.app/#!%7B%22dimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%2C%22position%22:%5B49166.4609375%2C49149.97265625%2C1694.5%5D%2C%22crossSectionScale%22:0.5856692901447939%2C%22projectionScale%22:131072%2C%22layers%22:%5B%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22tab%22:%22rendering%22%2C%22annotations%22:%5B%7B%22point%22:%5B49149.74609375%2C49037.05078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2279a0f82872520aa6471d37c0589fcff1931dc7da%22%7D%2C%7B%22point%22:%5B49220.8046875%2C49090.0859375%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22f5fda6876e7ddb4d49d5926e869c0483b615e15b%22%7D%2C%7B%22point%22:%5B49268.84765625%2C49172.140625%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22930b95f8c109ee320ef875551cf024d957f9e90d%22%7D%2C%7B%22point%22:%5B49172.765625%2C49222.171875%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%225b152a16ddf37e05a1ddd777c88cea0973cd4a06%22%7D%2C%7B%22point%22:%5B49079.69140625%2C49161.1328125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2280623dbccc502080b45f446b5ffaeb22eebca051%22%7D%2C%7B%22point%22:%5B49066.6796875%2C49083.078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%220071719479ad78f5de3299afe8bf44b1573df4e3%22%7D%5D%2C%22shader%22:%22#uicontrol%20vec3%20foo%20color%28%29%5Cn#uicontrol%20bool%20bar%20checkbox%28%29%5Cn%5Cnvoid%20main%28%29%20%7B%5Cn%20%20setColor%28defaultColor%28%29%29%3B%5Cn%20%20setColor%28vec3%281.0%2C%200.0%2C%200.0%29%29%3B%5Cn%7D%5Cn%22%2C%22shaderControls%22:%7B%22__hideInactiveControls%22:true%7D%2C%22name%22:%22annotation%22%7D%5D%2C%22selectedLayer%22:%7B%22size%22:568%2C%22visible%22:true%2C%22layer%22:%22annotation%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22size%22:568%2C%22visible%22:false%7D%7D)

I like this feature, the only ux improvement I would suggest is to indicate the number of hidden controls. I think it's worth letting the user know that things are being hidden. Something like this or even just (3)

Screenshot 2026-05-28 at 3 13 26 PM


I think it's worth adding support for detecting checkbox. Here is one possible solution solution:
https://github.com/chrisj/neuroglancer/commit/562a6888038e16a21a9e5127152d23515a2ba969

It only detects the presence of the string in the user shader (using regex word boundary).
If you define a new variable with that checkbox name, it will falsely assume it is used but there will also be a shader error. If you create an identical #define macro, it won't error but I think that's acceptable

fcollman:
I've incorporated @chrisj checkbox commit, and added a number of hidden elements to the UI

src/webgl/shader_ui_controls.ts
seankmartin: Yes I agree with Chris and I think being a property on the layer instead and then propagating that through to the shader controls would seem to make more sense to me than inside of the shader controls

src/webgl/shader_ui_controls.ts
```diff
@@ -1482,9 +1502,29 @@ export class ShaderControlState
         this.controls.value = result.controls;
       }
     }
+    // The active-controls set was derived from the previous shader and no
+    // longer matches the new control names; clear it so the UI shows
+    // everything until the next shader links.
+    this.lastReportedProgram = undefined;
+    if (this.activeControls.value !== undefined) {
+      this.activeControls.value = undefined;
+    }
```
seankmartin: I think this section can just have `this.lastReportedProgram = undefined`.
From what I can tell the `activeControls` reset is well intentioned but linking happens so fast that all you see is the controls flicker in and then out again. Resetting the program should be enough because then reportLinkedShader wont early return.

Maybe there's a small chance you see a stale result for a frame or two, but that should be much less noticable than all the controls flickering in and then hiding

src/webgl/shader_ui_controls.ts
```diff
@@ -1562,6 +1602,11 @@ export class ShaderControlState
     if (value === undefined) return;
     const { state } = this;
     verifyObject(value);
+    if (Object.prototype.hasOwnProperty.call(value, HIDE_INACTIVE_JSON_KEY)) {
+      this.hideInactiveControls.restoreState(value[HIDE_INACTIVE_JSON_KEY]);
+    } else {
+      this.hideInactiveControls.reset();
+    }
```
seankmartin: Moving the state control to the layer also helps remove this bit of awkwardness in the state management of the hide inactive json key here and in `reset` since on the layer state we'd follow a more common state pattern.

Small aside, I think could now just use [hasOwn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) (see [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty)) which we could apply on L1619 in the existing code.

src/widget/shader_controls.ts
```diff
@@ -176,28 +179,82 @@ export class ShaderControls extends Tab {
...
+
+    // Header row with the "Hide inactive controls" toggle. Built once and
+    // never torn down by `updateControls()`, so its UI state is stable
+    // across shader recompiles.
+    const header = document.createElement("label");
+    header.className = "neuroglancer-shader-controls-hide-inactive";
+    header.style.display = "flex";
+    header.style.alignItems = "center";
+    header.style.gap = "4px";
+    header.style.fontSize = "smaller";
```
seankmartin: This would be easier to later change if in CSS.

src/widget/shader_controls.ts
```diff
@@ -176,28 +179,82 @@ export class ShaderControls extends Tab {
     this.toolId = toolId;
     const { element } = this;
     element.style.display = "contents";
+
+    // Header row with the "Hide inactive controls" toggle. Built once and
+    // never torn down by `updateControls()`, so its UI state is stable
+    // across shader recompiles.
+    const header = document.createElement("label");
```
seankmartin: I think header is a bit of a confusing name here since really it's another control, just separated out from the main list of shader controls.

src/widget/shader_controls.ts
```diff
@@ -176,28 +179,82 @@ export class ShaderControls extends Tab {
...
+    header.style.fontSize = "smaller";
+    const checkbox = this.registerDisposer(
+      new TrackableBooleanCheckbox(state.hideInactiveControls, {
+        enabledTitle:
+          "Show all #uicontrols (including ones with no effect on the current render).",
+        disabledTitle:
+          "Hide #uicontrols whose uniforms are eliminated by the GLSL compiler" +
+          " for the current shader (e.g. controls only referenced inside an" +
+          " `if (checkbox)` branch that's currently false).",
+      }),
```
seankmartin: There is one part that is a bit unfortunate here, which is that the title would be better placed on the label element, not on the checkbox. That way hovering the text would also show this help message. Either the message could become static and describe the general function. Or stay dynamic and instead of using enabledTitle/disabledTitle here, listen to `hideInactiveControls.changed` directly and set `header.title` based on that

src/widget/shader_controls.ts
```diff
@@ -176,28 +179,82 @@ export class ShaderControls extends Tab {
...
-        this.registerCancellable(debounce(() => this.updateControls(), 0)),
-      ),
+      state.hideInactiveControls.changed.add(scheduleUpdate),
     );
     this.updateControls();
   }
 
   updateControls() {
-    const { element } = this;
+    const container = this.controlsContainer;
```
seankmartin: minor, but the pattern `const { controlsContainer} = this` is more commonly used in the code I think

seankmartin (commented):
Thanks for the changes! Left some comments, hope they help

src/widget/shader_controls.ts
fcollman: I have made it some static text on a tooltip

src/widget/shader_controls.ts
fcollman: i have reworded the comment

src/widget/shader_controls.ts
fcollman: I lifted this all to a css file.. not sure what the css philosphy is in terms of organizing css, happy to hear feedback on this.

src/webgl/shader_ui_controls.ts
fcollman: I have lifted the control to be tied to each layer

src/webgl/shader_ui_controls.ts
fcollman: fixed!

src/widget/shader_controls.ts
fcollman: changed

fcollman:
I think this test failure is unrelated and likely would work on retry, seems to be about OCDBT on windows.

src/widget/shader_controls.css
```diff
@@ -0,0 +1,22 @@
+/**
+ * @license
+ * Copyright 2024 Google Inc.
```
seankmartin: should be 2026

src/widget/shader_controls.ts
```diff
@@ -174,30 +185,82 @@ export class ShaderControls extends Tab {
...
+    hideInactiveControl.title =
+      "Hide #uicontrols that have no effect on the current render (their" +
+      " uniforms were eliminated by the GLSL compiler, e.g. controls only" +
+      " referenced inside an `if (checkbox)` branch that's currently false).";
+    const checkbox = this.registerDisposer(
+      new TrackableBooleanCheckbox(hideInactiveControls),
+    );
+    hideInactiveControl.appendChild(checkbox.element);
+    hideInactiveControl.appendChild(
+      document.createTextNode("Hide inactive controls"),
```
seankmartin: After trying this with layers where you can hide the shader code, I think "Hide inactive shader controls" would be better but I don't feel very strongly on it

seankmartin (approved):
Looks good thanks!

seankmartin:
One point now that we've moved the state representation with @fcollman and @chrisj, it would be good to add to the Python code the ability to control this new `hideInactiveControls` setting

src/widget/shader_controls.ts
fcollman: you just mean the text?

src/widget/shader_controls.css
fcollman: fixed

src/widget/shader_controls.ts
fcollman: i changed the text

fcollman:
> One point now that we've moved the state representation with @fcollman and @chrisj, it would be good to add to the Python code the ability to control this new `hideInactiveControls` setting

I have added this now

src/layer/annotation/index.ts
```diff
@@ -156,6 +156,7 @@ const LINKED_SEGMENTATION_LAYER_JSON_KEY = "linkedSegmentationLayer";
 const FILTER_BY_SEGMENTATION_JSON_KEY = "filterBySegmentation";
 const IGNORE_NULL_SEGMENT_FILTER_JSON_KEY = "ignoreNullSegmentFilter";
 const CODE_VISIBLE_KEY = "codeVisible";
+const HIDE_INACTIVE_CONTROLS_JSON_KEY = "hideInactiveControls";
```
chrisj: Almost merged this but.... last thing, I think we should also call this "hideInactiveShaderControls"

chrisj (changes requested):

src/layer/single_mesh/index.ts
```diff
@@ -130,7 +130,7 @@ export class SingleMeshUserLayer extends UserLayer {
     x[SHADER_JSON_KEY] = this.displayState.fragmentMain.toJSON();
     x[SHADER_CONTROLS_JSON_KEY] = this.displayState.shaderControlState.toJSON();
     x[CODE_VISIBLE_KEY] = this.codeVisible.toJSON();
-    x[HIDE_INACTIVE_CONTROLS_JSON_KEY] = this.hideInactiveControls.toJSON();
+    x[hide_inactive_shader_controls_JSON_KEY] = this.hideInactiveShaderControls.toJSON();
```
chrisj: this was formatted mixed-case

chrisj:
@fcollman this needs an npm run lint:fix

src/layer/single_mesh/index.ts
fcollman: fixed

src/layer/annotation/index.ts
fcollman: i have renamed this across all places using find/replace

PR #1020, opened 2026-06-01
fix(github-actions): bump Playwright packages to 1.60.0

chrisj:
Playwright < 1.60.0 can hang during browser install archive extraction on Node 24+

Refs: microsoft/playwright#40724


fcollman:
@chrisj why is this tagged with "help wanted" this seemed like it worked okay to me.

chrisj:
it needs attention as it is causing lots of actions to time out

fcollman:
oh agreed! i just think the "help wanted" tag means to me... I need someone else to help me with the content of the PR, when really we just need it to be rapidly reviewed and merged.

PR #1029, opened 2026-06-07
fix(datasource/deepzoom): remove extra slash when constructing tile path

### LINKED ISSUE #1028: bug? Neuroglancer unable to accurately use DZI descriptor file to locate image tiles
Testing DZI file load on Neuroglancer using the following data source URL: 
`deepzoom://s3://sea-ad-quantitative-neuropathology/middle-temporal-gyrus/H21.33.044/H21.33.044-A07-NEUN/downsampled/H21.33.044-A07-NEUN-primary/H21.33.044-A07-NEUN-primary.dzi`

The DZI file is named H21.33.044-A07-NEUN-primary.dzi and the folder containing tiles is named h21.33.044-a07-neun-primary_files as per the [original documentation from Microsoft](https://learn.microsoft.com/en-us/previous-versions/windows/silverlight/dotnet-windows-silverlight/cc645077(v=vs.95)). This directory structure is contained within a [public s3 bucket](https://sea-ad-quantitative-neuropathology.s3.amazonaws.com/index.html#middle-temporal-gyrus/H21.33.044/H21.33.044-A07-NEUN/downsampled/H21.33.044-A07-NEUN-primary/).

Neuroglancer will load the DZI file but is unable to populate any of the tiles. It looks like there is an extra "/" in the image tile paths directly prior to the file name - maybe this is is causing the issue ("//0_0.jpeg".) Please see the image and the following text:

`<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Key>middle-temporal-gyrus/H21.33.044/H21.33.044-A07-NEUN/downsampled/H21.33.044-A07-NEUN-primary/H21.33.044-A07-NEUN-primary_files/0//0_0.jpeg</Key><RequestId>3Z30CJFN9H81B433</RequestId><HostId>/PypFgp+emMVFd7KlDE3vlLvi+ruQ5JZP4EzSycml95VQxTFYB15BZYDfS1tZID3p+BFstLKZok=</HostId></Error>`

Image

chrisj:
resolves https://github.com/google/neuroglancer/issues/1028

Curious about thoughts on how this test is implemented, it's a bit verbose for a fix that deletes a single character

### DIFF
--- a/src/datasource/deepzoom/backend.ts
+++ b/src/datasource/deepzoom/backend.ts
@@ -81,7 +81,7 @@ export class DeepzoomImageTileSource extends WithParameters(
     const [x, y] = chunk.chunkGridPosition;
     const ox = x === 0 ? 0 : overlap;
     const oy = y === 0 ? 0 : overlap;
-    const path = `${this.tileKvStore.path}/${x}_${y}.${parameters.format}`;
+    const path = `${this.tileKvStore.path}${x}_${y}.${parameters.format}`;
     const response = await this.tileKvStore.store.read(path, {
       signal,
     });
--- a/tests/datasource/deepzoom.browser_test.ts
+++ b/tests/datasource/deepzoom.browser_test.ts
@@ -14,8 +14,45 @@
  * limitations under the License.
  */
 
-import "#src/datasource/deepzoom/register_default";
+import "#src/datasource/deepzoom/register_default.js";
 import "#src/sliceview/uncompressed_chunk_format.js";
+import { test } from "vitest";
+import { DeepzoomImageTileSource } from "#src/datasource/deepzoom/backend.js";
+import { ImageTileEncoding } from "#src/datasource/deepzoom/base.js";
 import { datasourceMetadataSnapshotTests } from "#tests/datasource/metadata_snapshot_test_util.js";
 
 datasourceMetadataSnapshotTests("deepzoom", ["14122_mPPC_BDA_s186.dzi"]);
+
+test("download reads tiles relative to trailing-slash directory paths", async ({
+  expect,
+}) => {
+  let readPath: string | undefined;
+  let readSignal: AbortSignal | undefined;
+  const source = {
+    parameters: {
+      url: "image_files/12/",
+      encoding: ImageTileEncoding.JPG,
+      format: "jpg",
+      overlap: 0,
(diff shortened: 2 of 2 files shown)


src/datasource/deepzoom/backend.ts
```diff
@@ -81,7 +81,7 @@ export class DeepzoomImageTileSource extends WithParameters(
     const [x, y] = chunk.chunkGridPosition;
     const ox = x === 0 ? 0 : overlap;
     const oy = y === 0 ? 0 : overlap;
-    const path = `${this.tileKvStore.path}/${x}_${y}.${parameters.format}`;
+    const path = `${this.tileKvStore.path}${x}_${y}.${parameters.format}`;
```
seankmartin, line 84: Overall looks good thanks! I separately noticed this one and was instead thinking to change `deepzoom/frontend.ts` to remove the trailing slash there on L191, so
```
url: `${this.url}/${array.length - 1 - index}`
//instead of
`url: `${this.url}/${array.length - 1 - index}/`
```

But I was wondering if it is more natural that the backend expects the path to end with a "/" or not. 

Either way this is good to me, but we've separately ran into a bug with this before in ngauth and now there is another one here. So I wonder if we could formalise somewhere the expectation, or have a helper to ensure consistency by either adding or removing trailing slashes depending on what is considered the standard.

seankmartin (approved):

src/datasource/deepzoom/backend.ts
chrisj: I removed the slash in backend only to follow the pattern in precomputed:
https://github.com/google/neuroglancer/blob/master/src/datasource/precomputed/backend.ts#L145

I think I prefer only attaching the trailing slash as needed when adding the next path segment so we can swap both if that sounds good

seankmartin: Sounds good to me, thanks!

jbms: I was trying to have the urls match the normalized data source / url pipeline syntax, where the normalized form for direcyory-based formats includes the trailing slash.

seankmartin:
@chrisj I merged this one since I think the fix is good, and we can always doc later about the urls using normalized form to include the trailing slash

PR #1032, opened 2026-06-09
fix: remove pointer cancel event in relative mouse drag

seankmartin:
This removes `pointercancel` listener in `startRelativeMouseDrag` on `cancel` otherwise it could remain registered indefinitely since it is added at the document level and keep triggering on any unrelated `pointercancel`.

For many cases this was pretty harmless, but it could lead to calling the `finishDragHandler` with stale data over and over again on unrelated `pointercancel`, such as when dragging panels around.


fcollman:
This seems straightforward to merge.

PR #1035, opened 2026-06-10
fix(sliceview): use different bounds to compute chunk positions

minnerbe:
When using the render backend, stacks sometimes don't have their origin at (0, 0). Those stacks would render normally on 2.40.1, but produced a blank display on 2.41.2.

Claude helped me figure out that this is caused by the clamping introduced by [a045804](https://github.com/google/neuroglancer/commit/a0458044c157f30c354723e76e3c280956b14ea2) in `updateFixedCurPositionInChunks`. There, `curPositionInChunks` (which is zeroed in the display dimensions) is clamped against `spec.{lower,upper}ChunkBound` (which is non-zero in these situations), leading to a negative chunk offset. This PR suggests to clamp against bounds computed from `nonDisplay{Lower,Upper}ClipBound`, which fixes the problem we had.

I'm happy for any feedback, especially if this _conceptually_ isn't the right fix.

### DIFF
--- a/src/sliceview/base.ts
+++ b/src/sliceview/base.ts
@@ -184,8 +184,7 @@ function updateFixedCurPositionInChunks<
 ): boolean {
   const { curPositionInChunks, fixedPositionWithinChunk } = tsource;
   const { nonDisplayLowerClipBound, nonDisplayUpperClipBound } = tsource;
-  const { rank, chunkDataSize, lowerChunkBound, upperChunkBound } =
-    tsource.source.spec;
+  const { rank, chunkDataSize } = tsource.source.spec;
   if (
     !getChunkPositionFromCombinedGlobalLocalPositions(
       curPositionInChunks,
@@ -219,12 +218,19 @@ function updateFixedCurPositionInChunks<
       return false;
     }
     const chunkSize = chunkDataSize[chunkDim];
-    // Given that clip bounds are already tested above, clamp chunk index to its
-    // bounds, to ensure floating-point imprecision does not result in an
-    // out-of-bounds index.
+    // Given that clip bounds are already tested above, clamp the chunk index to
+    // the range implied by those clip bounds, to ensure floating-point
+    // imprecision does not result in an out-of-bounds index. The clip bounds
+    // are used because they are expressed in the same coordinate frame as `x`.
+    const lowerChunkLimit = Math.floor(
+      nonDisplayLowerClipBound[chunkDim] / chunkSize,
+    );
+    const upperChunkLimit = Math.ceil(
+      nonDisplayUpperClipBound[chunkDim] / chunkSize,
+    );
     const chunk = (curPositionInChunks[chunkDim] = Math.min(
-      upperChunkBound[chunkDim] - 1,
-      Math.max(lowerChunkBound[chunkDim], Math.floor(x / chunkSize)),
+      upperChunkLimit - 1,
+      Math.max(lowerChunkLimit, Math.floor(x / chunkSize)),
     ));
     fixedPositionWithinChunk[chunkDim] = x - chunk * chunkSize;
   }
--- a/src/sliceview/base.ts
+++ b/src/sliceview/base.ts
@@ -182,9 +182,14 @@ function updateFixedCurPositionInChunks<
   globalPosition: Float32Array,
   localPosition: Float32Array,
 ): boolean {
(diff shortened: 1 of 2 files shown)


chrisj:
@minnerbe do you have a publicly shareable example? We should add a regression test for this. Still need to analyze the issue.

minnerbe:
Thanks for the fast reply, @chrisj! We currently don't have a public-facing render instance that could serve example data. If necessary we could set up a temporary instance, though.

It also seems that chunked data like zarr won't trigger this behavior since origins other than (0, 0) are represented by coordinate transformations. This is different in render, where they are represented by a nonzero `lowerVoxelBound`, causing wrong chunk indices to be computed.

I added a first version of a regression test that sets up the same state as in the example I tested. It fails before the fix and passes after. Unfortunately, testing lower (`updateFixedCurPositionInChunks`, not exposed) or higher (`getVolumetricTransformedSources`, needs webgl) in the call stack wasn't possible.

minnerbe:
In the last week, @trautmane kindly put together a minimal render webserver with a small example dataset. You can see a neuroglancer view [here](https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B4e-9%2C%22m%22%5D%2C%22y%22:%5B4e-9%2C%22m%22%5D%2C%22z%22:%5B4e-8%2C%22m%22%5D%7D%2C%22position%22:%5B-16276.2578125%2C-35086.19921875%2C1.5%5D%2C%22crossSectionScale%22:16%2C%22projectionScale%22:16384%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22render://https://render-test.janelia.org/test/neuroglancer/two_z_with_four_tiles%22%2C%22tab%22:%22source%22%2C%22name%22:%22two_z_with_four_tiles%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22two_z_with_four_tiles%22%7D%2C%22layout%22:%22xy%22%7D). This example fails to show in the current ng deployment, but renders in our internal ng instance that includes the fix from this PR (see screenshot below).

image

src/sliceview/base.ts
```diff
@@ -219,12 +218,19 @@ function updateFixedCurPositionInChunks<
...
+    // the range implied by those clip bounds, to ensure floating-point
+    // imprecision does not result in an out-of-bounds index. The clip bounds
+    // are used because they are expressed in the same coordinate frame as `x`.
+    const lowerChunkLimit = Math.floor(
+      nonDisplayLowerClipBound[chunkDim] / chunkSize,
+    );
+    const upperChunkLimit = Math.ceil(
+      nonDisplayUpperClipBound[chunkDim] / chunkSize,
+    );
     const chunk = (curPositionInChunks[chunkDim] = Math.min(
```
seankmartin, line 241: Thanks for raising this and making the example which helps a lot. I am a bit concerned this might not be the right fix though. I understand from the test point of view the fix. But in practice I think the non display clip bounds end up being positive and negative infinity for the dims being clamped here and then this reduces to `chunk = curPositionInChunks[chunkDim] = Math.floor(x / chunkSize)`. So maybe we instead just need to detect cases to not use clip bounds, or try to fix these bounds for the issue you are encountering? I'm also struggling a bit with this conceptually because I thought the chunk index would be for display dimensions so I don't really understand why we'd clamp against the non display bounds.

minnerbe: Sorry for the delay! I agree, it's a bit hard to wrap my head around this conceptually. As far as I understand right now, `updateFixedCurPositionInChunks`' responsibility is to process the non-display dims only. The display dims have a `curPositionInChunks[chunkDim] = 0` placeholder (the display rows of `fixedLayerToChunkTransform` are zeroed in frontend.ts:1033) that the pre-[a045804](https://github.com/google/neuroglancer/commit/a0458044c157f30c354723e76e3c280956b14ea2) code preserved, but the current code might clamp to a non-zero value, which causes `fixedPositionWithinChunk` to wrap. That only came up in the render backend because it sets a non-zero `lowerVoxelBound`.

Having understood this, maybe a better fix would be to just iterate over the non-display dimensions in the first place? This would allow us to clamp against `{lower,upper}ChunkBound` again. I've pushed that for the sake of discussion (I'm happy to squash/rebase later).

Issue #1038, opened 2026-06-11
bug: skeleton lines only mode still draws points


seankmartin:
<img width="1857" height="934" alt="Image" src="https://github.com/user-attachments/assets/8c35938f-c10d-4107-9ec5-998acf43785a" />

I think we need to thread the drawing mode through, or explicitly set the shader program for node drawing to `null`. The intended node shader program gating is not working properly.

chrisj:
resolved with: https://github.com/google/neuroglancer/pull/1039

PR #1039, opened 2026-06-11
refactor: clarify purpose of skeleton node drawing in lines only rendering mode

seankmartin:
Clarifies the intent of node drawing in skeletons, as the `nodeShader` is never null at the point in the code it was being checked.

This originally aimed to change the lines rendering. But given the double node drawing fix it should be fine to leave it instead.


seankmartin:
@chrisj @jbms @fcollman I updated this after our meeting to just be a refactor, thank you for the comments on it

fcollman:
@seankmartin this also needs the rebase.

seankmartin:
@jbms @chrisj based on our meeting I think if we merge this then I'll close https://github.com/google/neuroglancer/issues/1038 about the points rendering in lines mode. This PR aims to clarify in the code that we always draw nodes on purpose. The previous check about the shader being null is never hit, as we check earlier than this and early return on [L583](https://github.com/MetaCell/neuroglancer/blob/0335624796627cc2ac4a4de925c129ca842a52e0/src/skeleton/frontend.ts#L583).

```
    if (edgeShader === null || nodeShader === null) {
      // Shader error, skip drawing.
      return;
    }
```

chrisj (approved):

PR #1041, opened 2026-06-12
fix: protect against parsing omero color if none provided

seankmartin:
I have extracted out the color parsing guard fix from @r2dliu in #886 since I think that got a bit held up by the automated tests and it now also has conflicts to due a different fix for the omero parsing in v0.5 onwards.


seankmartin:
@jbms @chrisj any issues if I merge this one?

jbms:
Looks good go ahead

PR #1043, opened 2026-06-13
making selecting number input select on focus

fcollman:
This is a simple PR that makes it so that when you click into a numeric property input box, it will by default to selecting the entire value rather than simply putting the cursor into the box.  The most common use case for this UI is for people to enter a number for the first time, making it unlikely that they will want to be editing an existing digit on a number.  Once can simply click again to get to the digit/space you want to edit if you want to make a more targeted correction to the number field, but I think that is a much smaller set of use cases and so this is a preferable behavior.

### DIFF
--- a/src/ui/annotation_schema_tab.ts
+++ b/src/ui/annotation_schema_tab.ts
@@ -915,9 +915,11 @@ class AnnotationUIProperty extends RefCounted {
       input = this.createNumberInput(inputValue as number, config, readonly);
     } else if (config.type === "text") {
       input = createTextAreaElement(inputValue as string, readonly);
+      input.addEventListener("focus", () => input!.select());
     } else {
       input = document.createElement("input");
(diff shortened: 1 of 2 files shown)


seankmartin (approved):

chrisj:
@fcollman would you also want it to auto select for the enum label fields?

We could generalize this since we do this for several other inputs.

Some that don't have it are the settings limits and render -> color seed but I don't see any reason not to make it consistent.

fcollman:
I don't think the settings limits are the same.. often for example memory i just want to change the first few digits, and not have to type all those zeros. I also think we can do that case by case and don't need to stack all those updates into a single PR.

fcollman:
I do think the schema enum defaults are the same though, so i'll add that.

chrisj (approved):

PR #1044, opened 2026-06-13
Design draft: Annotation prop input enhancements

fcollman:
This is an design attempt to reconcile the functionality of #650 (which was an update to #136 ) with the new structured annotation schema introduced by #809.

This PR is a functional design PR, where'd I think it would be good to focus on the UI design and debate the conceptual implementation, independent of the actual implementation before we worry too much about the code. 

The use case for this design is that a user wants to rapidly iterate through a set of annotations/locations and annotate that they belong to a certain class.  Clicking into the annotation property box to label each class is un-ergonomic.  

To accomplish this, this PR proposes adding tool keybinds to the boolean and enum fields of the schema tab.

image

So that users can quickly cause the selected annotation to change the enum value to the bound hotkey or toggle the boolean value. 

The PR (like #650) adds hotkey binds of alt-up and alt-down to allow users to navigate the list.

Relatedly, once a user has tagged a bunch of annotations with this metadata, they often want to review a subset of them.  For example, they may tagged a subset with "uncertain", or simply want to show a colleague a subset of examples that they found that were put into some relatively rare class.  

This is related to a more general need to be able to see the properties in the list view, and be able to sort the list by those properties, similar to how the segment properties currently works. 

To accomplish this the PR lets users add numerical properties to the annotation list view.  It also provides up and down arrow buttons to let users sort the list.  This doesn't actually change the order of annotations in the state, so that unsorting all columns leaves the list in its original order. The up/down arrow actions move the selection in the currently displayed list and not the order of annotations in the state. 

image

image

I think this follows the spirit of the UI where the description field shows up in the annotation list below the annotation, so that users can find the annotations that they left particular notes on.  With the structured schema now incorporated, users need a way to find those annotations. 

The workflow setup in #650 would be accomplished by adding a boolean property for each tag that you wanted to enter, and assigning each a keybind.  This is more general in that you could either do that, or implement a "forced-choice" workflow with an enum, such that selecting one category implicitly deselects the other categories, depending on what is appropriate for your use-case.

I thought about, but didn't add, a more general query box like in segment properties that would allow users to search/filter through annotations by combining filters on columns, but that is something we could debate.

I have not considered deeply how to reconcile this with other non-local annotation sources which might be dynamically loading annotations into view.  We currently don't show any list of annotations in that case, or way to visualize their properties.  Large annotations lists will be dynamically adding/removing data from the view and the list will almost never be authoritative or complete.

### DIFF
--- a/src/layer/annotation/index.ts
+++ b/src/layer/annotation/index.ts
@@ -108,6 +108,8 @@ import { Tab } from "#src/widget/tab_view.js";
 const POINTS_JSON_KEY = "points";
 const ANNOTATIONS_JSON_KEY = "annotations";
 const ANNOTATION_PROPERTIES_JSON_KEY = "annotationProperties";
+const ANNOTATION_LIST_COLUMNS_KEY = "annotationListColumns";
+const ANNOTATION_LIST_SORT_KEY = "annotationListSort";
 const ANNOTATION_RELATIONSHIPS_JSON_KEY = "annotationRelationships";
 const CROSS_SECTION_RENDER_SCALE_JSON_KEY = "crossSectionAnnotationSpacing";
 const PROJECTION_RENDER_SCALE_JSON_KEY = "projectionAnnotationSpacing";
@@ -708,6 +710,26 @@ export class AnnotationUserLayer extends Base {
     this.annotationDisplayState.shaderControls.restoreState(
       specification[SHADER_CONTROLS_JSON_KEY],
     );
+    const shownColumns = verifyOptionalObjectProperty(
+      specification,
+      ANNOTATION_LIST_COLUMNS_KEY,
+      verifyStringArray,
+    );
+    if (shownColumns !== undefined) {
+      this.annotationListShownColumns.value = shownColumns;
+    }
+    const sortJson = verifyOptionalObjectProperty(
+      specification,
+      ANNOTATION_LIST_SORT_KEY,
+      verifyObject,
+    );
+    if (sortJson !== undefined) {
+      const propertyId = verifyString(sortJson["propertyId"]);
+      const order = verifyString(sortJson["order"]);
+      if (order === "asc" || order === "desc") {
+        this.annotationListSortState.value = { propertyId, order };
+      }
+    }
   }
 
   getLegacyDataSourceSpecifications(
@@ -983,6 +1005,14 @@ export class AnnotationUserLayer extends Base {
     x[SHADER_CONTROLS_JSON_KEY] =
       this.annotationDisplayState.shaderControls.toJSON();
     Object.assign(x, this.linkedSegmentationLayers.toJSON());
+    const shownColumns = this.annotationListShownColumns.value;
+    if (shownColumns.length > 0) {
+      x[ANNOTATION_LIST_COLUMNS_KEY] = shownColumns;
+    }
+    const sortState = this.annotationListSortState.value;
+    if (sortState !== null) {
+      x[ANNOTATION_LIST_SORT_KEY] = sortState;
+    }
     return x;
   }
 
--- a/src/layer/annotation/index.ts
+++ b/src/layer/annotation/index.ts
@@ -19,10 +19,14 @@ import "#src/layer/annotation/style.css";
 import type { AnnotationDisplayState } from "#src/annotation/annotation_layer_state.js";
 import { AnnotationLayerState } from "#src/annotation/annotation_layer_state.js";
 import { MultiscaleAnnotationSource } from "#src/annotation/frontend_source.js";
-import type { AnnotationPropertySpec } from "#src/annotation/index.js";
+import type {
+  AnnotationNumericPropertySpec,
+  AnnotationPropertySpec,
+} from "#src/annotation/index.js";
 import {
   annotationPropertySpecsToJson,
   AnnotationType,
+  isAnnotationNumericPropertySpec,
   LocalAnnotationSource,
   parseAnnotationPropertySpecs,
 } from "#src/annotation/index.js";
@@ -54,12 +58,18 @@ import {
   WatchableValue,
   observeWatchable,
 } from "#src/trackable_value.js";
+import {
+  setEnumPropertyToolJson,
+  toggleBoolPropertyToolJson,
+} from "#src/ui/annotation_properties.js";
 import { AnnotationSchemaTab } from "#src/ui/annotation_schema_tab.js";
 import type {
   AnnotationLayerView,
   MergedAnnotationStates,
 } from "#src/ui/annotations.js";
 import { UserLayerWithAnnotationsMixin } from "#src/ui/annotations.js";
+import type { ToolActivation } from "#src/ui/tool.js";
+import { LayerTool, registerTool, unregisterTool } from "#src/ui/tool.js";
 import { animationFrameDebounce } from "#src/util/animation_frame_debounce.js";
 import type { Borrowed, Owned } from "#src/util/disposable.js";
 import { RefCounted } from "#src/util/disposable.js";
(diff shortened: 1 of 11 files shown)


syamace-metacell:
@fcollman, from a design perspective (taking into context #1048 as well)

Adding keybinds to annotations
- The keybinds works well for rapid annotation and it aligns with existing patterns on Neuroglancer.

Adding annotation properties in the annotation list:
- Adding annotation properties in the list makes sense, as well especially with the ability to sort/filter.
- Even though we’re introducing new patterns with the ability to show/hide certain properties in the list, this would be useful and needed if we were to display annotation properties in the list.

This PR's design approach is simple and nice, with some additions (stated below), it would target the use cases while keeping the introduction of new patterns to the minimal. This version make it seem like it's just a small addition on top of exisitng patterns.

These are additions I suggest to make it more aligned with the rest of the UI:
1. Query box (like seg. & skeleton tabs)
2. On hover of each annotation row, filter icons appears for each of the value (keeping the similar pattern on seg. tab for 'filter by label')

Sort minor UI feedback
- Sort UI needs to be aligned with what's on seg. tab with the use of 1 arrow, instead of 2.

chrisj:
[minimal example](https://neuroglancer--pr1044-7yjtbv3c.web.app/#!%7B%22dimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%2C%22position%22:%5B49161.05078125%2C49127.3359375%2C1694.5%5D%2C%22crossSectionScale%22:0.5856692901447939%2C%22projectionScale%22:131072%2C%22layers%22:%5B%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B1.8e-8%2C%22m%22%5D%2C%22y%22:%5B1.8e-8%2C%22m%22%5D%2C%22z%22:%5B4.5e-8%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePoint%22%2C%22toolBindings%22:%7B%22A%22:%22toggleBoolProperty_proofread%22%2C%22R%22:%22setEnumProperty_color_0%22%2C%22G%22:%22setEnumProperty_color_1%22%2C%22B%22:%22setEnumProperty_color_2%22%7D%2C%22tab%22:%22annotations%22%2C%22annotations%22:%5B%7B%22point%22:%5B49149.74609375%2C49037.05078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2279a0f82872520aa6471d37c0589fcff1931dc7da%22%2C%22props%22:%5Bfalse%2C0%2C5%5D%7D%2C%7B%22point%22:%5B49220.8046875%2C49090.0859375%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22f5fda6876e7ddb4d49d5926e869c0483b615e15b%22%2C%22props%22:%5Btrue%2C1%2C40%5D%7D%2C%7B%22point%22:%5B49268.84765625%2C49172.140625%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%22930b95f8c109ee320ef875551cf024d957f9e90d%22%2C%22props%22:%5Btrue%2C0%2C20%5D%7D%2C%7B%22point%22:%5B49172.765625%2C49222.171875%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%225b152a16ddf37e05a1ddd777c88cea0973cd4a06%22%2C%22props%22:%5Btrue%2C2%2C9%5D%7D%2C%7B%22point%22:%5B49079.69140625%2C49161.1328125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%2280623dbccc502080b45f446b5ffaeb22eebca051%22%2C%22props%22:%5Bfalse%2C2%2C10%5D%7D%2C%7B%22point%22:%5B49066.6796875%2C49083.078125%2C1694.5%5D%2C%22type%22:%22point%22%2C%22id%22:%220071719479ad78f5de3299afe8bf44b1573df4e3%22%2C%22props%22:%5Bfalse%2C0%2C4%5D%7D%5D%2C%22annotationProperties%22:%5B%7B%22id%22:%22proofread%22%2C%22description%22:%22%22%2C%22type%22:%22bool%22%2C%22default%22:true%7D%2C%7B%22id%22:%22color%22%2C%22description%22:%22%22%2C%22type%22:%22uint8%22%2C%22enum_labels%22:%5B%22red%22%2C%22green%22%2C%22blue%22%5D%2C%22enum_values%22:%5B0%2C1%2C2%5D%7D%2C%7B%22id%22:%22size%22%2C%22description%22:%22%22%2C%22type%22:%22float32%22%2C%22default%22:10%7D%5D%2C%22shader%22:%22vec3%20red%20=%20vec3%281.0%2C%200.0%2C%200.0%29%3B%5Cnvec3%20green%20=%20vec3%280.0%2C%201.0%2C%200.0%29%3B%5Cnvec3%20blue%20=%20vec3%280.0%2C%200.0%2C%201.0%29%3B%5Cnvoid%20main%28%29%20%7B%5Cn%20%20if%20%28prop_color%28%29%20==%200u%29%20%7B%5Cn%20%20%20%20setColor%28red%29%3B%5Cn%20%20%7D%20else%20if%20%28prop_color%28%29%20==%201u%29%20%7B%5Cn%20%20%20%20setColor%28green%29%3B%5Cn%20%20%7D%20if%20%28prop_color%28%29%20==%202u%29%20%7B%5Cn%20%20%20%20setColor%28blue%29%3B%5Cn%20%20%7D%5Cn%20%20setPointMarkerSize%28prop_size%28%29%29%3B%5Cn%7D%5Cn%22%2C%22shaderControls%22:%7B%22mode%22:%22large%22%7D%2C%22annotationListColumns%22:%5B%22color%22%2C%22proofread%22%2C%22size%22%5D%2C%22name%22:%22annotation%22%7D%5D%2C%22selectedLayer%22:%7B%22size%22:568%2C%22visible%22:true%2C%22layer%22:%22annotation%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22size%22:568%2C%22layers%22:%7B%22annotation%22:%7B%22annotationId%22:%22f5fda6876e7ddb4d49d5926e869c0483b615e15b%22%2C%22annotationSource%22:0%2C%22annotationSubsource%22:%22default%22%7D%7D%7D%7D)

shift+A toggles proofread status
shift+R/G/B toggles color

fcollman:
> 1. Query box (like seg. & skeleton tabs)
What would the query box implement? the same thing as i have in #1048 ? a string representation of a query/sort filter?

> 2. On hover of each annotation row, filter icons appears for each of the value (keeping the similar pattern on seg. tab for 'filter by label')
what would you do for numerical values?

I see from your comment on #1048 that you might be in a similar conflicted place to where i was.. where i felt like not having numerical filters was too much of a concern, and then i wanted to align the UI as close as possible to the segment properties.... but yeah in the end it's a big change and makes a fairly complicated UI.  

In terms of sequencing, it seems like maybe this would be a 

1. keybinds on enums/booleans is uncontroversial and everyone so far likes the design that is in these two PRs. This can go in as it's own piece i think (along with hotkeys to move up/down the annotation list).

2. We need to have the ability to list and sort annotation properties in the table view, with some mechanism to add/remove what properties are listed there. We need to align on the json representation of what properties are shown there and what the sorting mechanism is.  It seems like the display and sort part of the design people like (with the exception that the sorting buttons have to change to a single arrow tri-state mode like i have in #1048 )

If we pre-align around the idea that there is a query box and a string, and that is what goes into the json state and is parsed into setting the UI, and conversely changing things in the UI edits that string, then we could punt on what the UI is for setting all aspects of that query string.  For example, the UI doesn't need to allow people to set numerical filters immediately, or necessarily set filters on enum properties immediately, or we could decide to change it. 

We could put the query box into this PR, so that people would have the functionality and we would have the code infrastructure to inter-operate with it and the annotation list. 

3. we could add enhancements to allow UI setting of the query string in a way that is visually aligned between segment properties and the skeleton tab, including in/out filters on categorical variables and numerical filters.

4. we can add more default annotation properties to annotations (length, volume, duration, delta_{dim}, extent_{dim}) that are derived dynamically from the annotation data rather then written explicitly by the user.

syamace-metacell:
> 1. Query box (like seg. & skeleton tabs)
> What would the query box implement? the same thing as i have in https://github.com/google/neuroglancer/pull/1048 ? a string representation of a query/sort filter?

Yes exactly. Whats on #1048 is good, it introduces some new type of queries like the boolean and strings of the numerical filters but don't think it'll hurt to add them.

> 2. On hover of each annotation row, filter icons appears for each of the value (keeping the similar pattern on seg. tab for 'filter by label')
> what would you do for numerical values?

Right now, the same for simplicity, filter icon will only filter by that 1 numerical value. For numerical range, it would have to be done through the query box.

seankmartin:
Thanks for making this! Overall I agree the most with:

> If we pre-align around the idea that there is a query box and a string, and that is what goes into the json state and is parsed into setting the UI, and conversely changing things in the UI edits that string, then we could punt on what the UI is for setting all aspects of that query string. For example, the UI doesn't need to allow people to set numerical filters immediately, or necessarily set filters on enum properties immediately, or we could decide to change it.

I would need to think on it a bit more for exacts. But it feels a little bit similar to a leading drive in the schema UI creation. Neuroglancer annotation properties are the source of truth. The UI can do whatever it likes so to speak, but it can only ever make changes through that property API. Here, if we can manage to align on the core state representation of these annotation properties usage improvements then we can be a bit looser with the UI because it could always update in the future without us worrying about supporting old state representations and hopefully without needing to change the state representation.

fcollman:
closing this as we have started down an implementation path with #1063

PR #1045, opened 2026-06-13
fix: pin down crypotgraphy for x86 mac os x test install

fcollman:
cryptography 49.0.0 released yesterday dropped macOS x86_64 wheels (48.x had universal2; 49.x is arm64-only + sdist). google-auth depends on cryptography with no conditional pins, so it is pulled in. 

This forced the python test build to source-build cryptography it under arm64→x86_64 cross-compilation, which fails due to SSL libraries being installed.  

Support for intel macs is being phased out by apple, so I think pinning down the test for now makes sense and let users decide if they want to try to compile cryptography themselves or pin down the version.  This also allows neuroglancer tests to pass. 

Longer term it probably makes sense to drop support entirely.


fcollman:
actions are failing due to a quota on a firebase channel, a different unrelated problem.

.github/workflows/build.yml
```diff
@@ -220,6 +220,14 @@ jobs:
...
           CIBW_BEFORE_ALL_LINUX: /project/python/build_tools/cibuildwheel_linux_cache_setup.sh /host${{ steps.uv-cache.outputs.dir }}
           CIBW_BUILD: ${{ matrix.cibw_build }}
+          # cryptography>=49 dropped macOS x86_64 wheels (48.x shipped universal2),
+          # forcing a source build that can't cross-compile openssl on the arm64
+          # runner. A uv constraint caps the transitive google-auth -> cryptography
+          # resolve during the wheel install (CIBW_TEST_REQUIRES can't, since it is
+          # installed separately). Scoped to the x86_64 test env via UV_CONSTRAINT;
+          # empty for other entries (uv treats empty as unset) and never passed into
+          # the Linux container, so all other jobs are unaffected.
+          UV_CONSTRAINT: ${{ matrix.wheel_identifier == 'macos_x86_64' && format('{0}/python/build_tools/macos-x86_64-test-constraints.txt', github.workspace) || '' }}
```
seankmartin: Thanks for the change, I think I prefer this constraint inline as opposed to in another file. However I see your first commit used this strategy of having the constraint inline (with `CIBW_TEST_REQUIRES` of course). Was there a requirement to have the constraint in another file?

chrisj:
used an intel mac recently and noticed there was already widespread lack of support for it. Apple is dropping support in macos 27

chrisj (approved):

.github/workflows/build.yml
fcollman: This is just how UV_CONSTRAINT works, it takes reference to a list of files and not inline values.
https://docs.astral.sh/uv/reference/environment/#uv_constraint

fcollman: CIBW_TEST_REQUIRES didn't work because that environment variable wasn't passed down to the environment where it was building the wheel.

jbms: How about we just ensure the uv lock file itself specifies the desired version, maybe by adding an extra dependency group with a platform-specific constraint?

fcollman: so one argument is that the reason this is failing is because it's an arm64 runner trying to cross compile a wheel for x86 in the CI process.  An actual user on actual x64 system might be able to compile the cryptography package from source, we just can't do it in the CI.  If we add it as a constraint then we are effectively capping the version that intel mac users will install, and that could easily wind up constraining how they install things if something starts depending on the newest patch of cryptography. 

if we constrain it here just for the purposes of our tests we will limit the scope of this constraint to our own CI build process.

fcollman: perhaps more broadly, either way we do this, at some point it seems likely that google auth will start increasing it's minimum pin on cryptography (currently at   "cryptography >= 38.0.3") but the problem emerges at v49 . 

When this happens our intel mac test will fail entirely.  

A reasonable alternative is to just drop testing of intel based mac entirely and assume that either it works because we have tested in elsewhere, or the reasons it is not working is due to dropped support of some kind and we aren't going to support it alone.

fcollman: I have pushed this commit/version just to make sure that it indeed works in CI and people can see what it looks like.

fcollman: to be clear, this builds a wheel for intel based macs, but doesn't actually fully test the wheel, so intel based macs will be on their own to resolve any dependency problems that emerge from dropped support for intel based macs that will undoubtably arise as time goes on.

seankmartin (approved):
Thanks @fcollman for the changes, this sounds like a reasonable strategy to me to drop it from the test matrix

jbms:
I think we could also specify the necessary constraints using this option: https://docs.astral.sh/uv/reference/settings/#build-constraint-dependencies

That will not imply a transitive constraint that applies to other packages that depend on neuroglancer; it only affects the uv.lock file that is used for development and CI purposes.

fcollman:
> I think we could also specify the necessary constraints using this option: https://docs.astral.sh/uv/reference/settings/#build-constraint-dependencies
> 
> That will not imply a transitive constraint that applies to other packages that depend on neuroglancer; it only affects the uv.lock file that is used for development and CI purposes.

This works, but would constrain our testing to pinned down versions for all Darwin wheels,  the uv.lock composition happens on the runner with is arm64, and then it tries to cross compile the x86_64. If we set the uv build constraint to be x86_64 specific it won't trigger, and then the wheel testing will still fail as it tries to pull it in.

just for demonstration i'll push the change like that and check that it still fails as I am anticipating.

fcollman:
@jbms I think this is what you were suggesting and you can see that it still fails on x86_64 due to trying to pull in cryptography version 49.

fcollman:
so we could..

1. mark the build constraint for all darwin builds to cryptography<49, which would open us up to Darwin failing tests if/when google auth bumps up their min dependency to >=49, or wheels not building for some other reason on the most recent current dependency constraints and us not having any signal about that.  I don't think this is a great option because it's hurting our robustness checking for osx wheels because of  osx x86_64 systems which are clearly a falling minority of systems.

2. revert to the last commit where we just stop testing osx x86_64 builds.  This is probably where we will end up long run, or else just dropping building osx x86_64 builds entirely. 

3. revert back to my second commit, where we dynamically inject in the global uv constraint in the test matrix when it's building the  osx x86_64 wheel using the UV_CONSTRAINT mechanism, that requires the extra file. 
I personally think this is the best outcome, though agreed, it makes another file and feels messy.. but this is i think a bit. messy.  It will only work as long as cryptography<49 is allowed by the other constraints of neuroglancer, which feels like it won't last forever.

jbms:
I see the constraint is mentioned in the lock file but the lock file does not actually specify a different resolved version, so somehow it didn't work.

fcollman:
the lock file did go down to 46 but it didn't matter because i realize that cibuildwheel uses uv pip install to setup environment.

fcollman:
i have reverted back to my second commit... i don't think there is anything wrong with it, even if it's not the most elegant solution in terms of files.

chrisj:
I'm pushing this as if we want to switch to another approach, we can always do that later but I think it's important to get CI running again as long as it is done in a safe manner

PR #1058, opened 2026-06-22
fix: correct screenshot ui styling by using new framed dialog class

seankmartin:
The main benefit of this is fixing some styling in the screenshot menu, especially around hover states and pointers. The screenshot menu in the UI previously used the generic overlay class, but since framed dialog was introduced we can use this instead and be more consistent. For example, old:

https://github.com/user-attachments/assets/36fe22cc-166d-451e-bc94-82885c0cdfaf

And new:

https://github.com/user-attachments/assets/25cf1ba5-8803-4f53-b702-d90e4bb198af

This is related to https://github.com/google/neuroglancer/pull/805 but is extracting a subset of that PR which is focused just around styling fixes and leveraging the `FramedDialog` for the screenshot menu to help with that.

### DIFF
--- a/src/overlay.css
+++ b/src/overlay.css
@@ -75,8 +75,50 @@
   margin-top: 1px;
 }
 
-.neuroglancer-framed-dialog-close-button {
-  cursor: pointer;
+.neuroglancer-framed-dialog:focus-visible {
+  outline: none;
+}
+
+.neuroglancer-framed-dialog-header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0.5rem 1.5rem;
+  border-bottom: 1px solid #ccc;
+  height: 2rem;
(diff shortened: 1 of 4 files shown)


fcollman:
Let's rebase this so we can get the checks passing!

seankmartin:
Thanks @fcollman good spot! Updated

seankmartin:
@jbms @chrisj ok if I merge this one? The changes are self contained to the the framed dialog class and the screenshot UI, which now uses the framed dialog. The annotation schema tab overlay for editing an annotation property description is the only other thing which uses the framed dialog. 

The real aim of this PR is to clean up some UI inconsistencies that the screenshot overlay had, and in the process we slightly generalise the framed dialog class which should make it easier to use in the future.

chrisj (approved):
@seankmartin go ahead!

PR #1063, opened 2026-06-25
Annotation schema hotkeys

fcollman:
This adds layer tools to bind hotkeys to boolean and enum values in the local annotation schema, to make adding structured annotations more ergonomic.  It also adds a layer tool to iterate through the annotation list. 

image
image

This is step 1 of moving toward something like #1044 or #1048.


config/custom-keybinds.json
```diff
@@ -0,0 +1,41 @@
+{
+  "keym": {
```
chrisj: this file should be removed

fcollman: ah whoops.. good catch

src/layer/annotation/index.ts
```diff
@@ -398,7 +412,150 @@ class LinkedSegmentationLayersWidget extends RefCounted {
   }
 }
 
+// A keybindable tool that sets the currently-selected (iterated) annotation's
```
seankmartin: I think it's just currently selected, not sure the (iterated) part matters

src/layer/annotation/index.ts
```diff
@@ -398,7 +412,150 @@ class LinkedSegmentationLayersWidget extends RefCounted {
...
+// enum property to a specific value, then advances to the next annotation so
```
seankmartin: It doesn't advance automatically (which is correct imo)

src/layer/annotation/index.ts
```diff
@@ -398,7 +412,150 @@ class LinkedSegmentationLayersWidget extends RefCounted {
...
+    return setEnumPropertyToolJson(this.propertyIdentifier, this.enumValue);
+  }
+
+  get description() {
+    return `set ${this.propertyIdentifier} = ${this.enumLabel}`;
+  }
+}
+
+// A keybindable tool that toggles a boolean annotation property on the
+// currently-selected (iterated) annotation.
```
seankmartin: Same here don't think iterated is needed as comment

src/layer/annotation/index.ts
```diff
@@ -407,6 +564,13 @@ export class AnnotationUserLayer extends Base {
     new WatchableValue([]);
   private localAnnotationRelationships: string[];
   private localAnnotationsJson: any = undefined;
+  // Tool ids currently registered for enum property values, keyed by tool id.
+  private registeredEnumTools = new Map<
+    string,
+    { propertyIdentifier: string; enumValue: number }
+  >();
+  // Tool ids currently registered for bool property toggles, keyed by tool id.
+  private registeredBoolTools = new Set<string>();
```
seankmartin: It's not a blocker, but somehow this feels like this should be registeredPropertyTools and then within that the tool can describe what it is, so there's just one map. Then later we can have number tools or color tools without needing to track each separately. If that causes complications then would suggest to rename to `registeredEnumPropertyTools` and `registeredBoolPropertyTools`.

src/layer/annotation/index.ts
```diff
@@ -398,7 +412,150 @@ class LinkedSegmentationLayersWidget extends RefCounted {
...
+// currently-selected (iterated) annotation.
+class ToggleBoolPropertyTool extends LayerTool<AnnotationUserLayer> {
+  constructor(
+    public propertyIdentifier: string,
+    layer: AnnotationUserLayer,
+  ) {
+    super(layer, true);
+  }
+
+  activate(activation: ToolActivation<this>) {
```
seankmartin: Although I'm not a huge fan of suggesting to reduce duplication too early, the body of this activate is almost exactly the same as the enum property tool activate. A helper function to update a property would seem useful to make and use across both.

src/layer/annotation/index.ts
```diff
@@ -455,21 +619,104 @@ export class AnnotationUserLayer extends Base {
...
+        this.syncEnumPropertyTools();
+      }),
+    );
+  }
+
+  // Keep one keybindable tool registered per (enum property × enum value) and
+  // per bool property so each can be bound to a hotkey in the schema editor.
+  // Registers newly-added options and tears down removed ones, including any
+  // live keybindings in this layer's tool binder.
+  private syncEnumPropertyTools(
```
seankmartin: This also syncs bools so the name is a bit confusing. Could it be `syncPropertyTools`? Then we'd also update the comment here.

This also further suggests combining the map and set of registered enum and bool tools could be the right direction as it would seem to simplify the cleanup step here which currently manages two separate registration paths.

seankmartin: I've not dug into why this happens, but my suspicion would be from here. If you bind a key to an enum, and then change the value of that enum in the schema, you can no longer bind a new key to that enum tool.

src/ui/annotation_schema_tab.ts
```diff
@@ -730,6 +743,20 @@ class AnnotationUIProperty extends RefCounted {
     enumRow.appendChild(valueInput);
 
     if (!this.readonly) {
+      // Keybind affordance: bind a key so that pressing it while iterating the
+      // annotation list sets the selected annotation's property to this enum
+      // value (and advances to the next annotation).
```
seankmartin: It doesn't advance and I don't think the comment here is needed

src/ui/annotation_schema_tab.ts
```diff
@@ -730,6 +743,20 @@ class AnnotationUIProperty extends RefCounted {
...
+      const toolButton = makeToolButton(
+        this,
+        this.parentView.layer.toolBinder,
+        {
+          toolJson: setEnumPropertyToolJson(oldProperty.identifier, value),
+          title: `Bind a key to set the selected annotation's ${oldProperty.identifier} to "${label}"`,
```
seankmartin: in makeToolButton the title is set as the title of the labelElement, and this has no label. Personally open to makeToolButton changing to set it as a mix of the title and ` ToolBindingWidget`'s title which is `click -> bind key, dbclick -> unbind`

src/ui/annotation_schema_tab.ts
```diff
@@ -545,6 +548,16 @@ class AnnotationUIProperty extends RefCounted {
...
       this.registerEventListener(checkbox, "change", changeFunction);
+      const toolButton = makeToolButton(
+        this,
+        this.parentView.layer.toolBinder,
+        {
+          toolJson: toggleBoolPropertyToolJson(oldProperty.identifier),
+          title: `Bind a key to toggle the selected annotation's ${oldProperty.identifier}`,
+        },
+      );
+      toolButton.classList.add("neuroglancer-annotation-schema-enum-keybind");
```
seankmartin: Would suggest just `neuroglancer-annotation-schema-keybind`

src/ui/annotation_schema_tab.ts
```diff
@@ -730,6 +743,20 @@ class AnnotationUIProperty extends RefCounted {
...
+      // value (and advances to the next annotation).
+      const toolButton = makeToolButton(
+        this,
+        this.parentView.layer.toolBinder,
+        {
+          toolJson: setEnumPropertyToolJson(oldProperty.identifier, value),
+          title: `Bind a key to set the selected annotation's ${oldProperty.identifier} to "${label}"`,
+        },
+      );
+      toolButton.classList.add("neuroglancer-annotation-schema-enum-keybind");
```
seankmartin: Would suggest just `neuroglancer-annotation-schema-enum-keybind`

src/ui/annotations.ts
```diff
@@ -516,6 +543,24 @@ export class AnnotationLayerView extends Tab {
     mutableControls.appendChild(helpIcon);
 
     toolbox.appendChild(mutableControls);
+
+    const navRow = document.createElement("div");
```
seankmartin: While precomputed annotations have no list, I think this should only be visible on local annotations (or I guess if another type of annotation has a list on those too, think the readonly style check used to show/hide the annotations type selector should do)

src/ui/annotations.ts
```diff
@@ -2004,6 +2049,62 @@ export function UserLayerWithAnnotationsMixin<
...
+        (x) =>
+          x.sourceIndex === layerSelectionState.annotationSourceIndex &&
+          (layerSelectionState.annotationSubsource === undefined ||
+            x.subsourceId === layerSelectionState.annotationSubsource),
+      );
+      if (annotationLayerState === undefined) return undefined;
+      return { annotationLayerState, annotationId };
+    }
+
+    changeSelectedIndex(offset: number) {
```
seankmartin: Minor, but not sure on the name of this func. Maybe `shiftSelectedIndexBy` or something like that to indicate what the offset is doing

src/ui/annotations.ts
```diff
@@ -2004,6 +2049,62 @@ export function UserLayerWithAnnotationsMixin<
...
+          // this only happens if offset is negative
+          annotationLayerStateIndex -= 1;
+        } else if (index === annotations.length) {
+          // this only happens if offset is positive
+          annotationLayerStateIndex += 1;
+        } else {
+          const annotation = annotations[index];
+          this.selectAnnotation(annotationLayerState, annotation.id, true);
+          moveToAnnotation(this, annotation, annotationLayerState);
+          return;
```
seankmartin: I think this will likely hit an error at the boundaries if offset is not +1 or -1 because index will be negative but not -1 or postive but not the length of the array. Would suggest to update the type indication in the function away from number to `-1 | 1`. I also think if possible, refactoring this endless while loop to be clearer would be good. I think it's primarily trying to handle that we don't know which layer will have the annotation, and dealing with boundary effects and empty arrays but it feels a like it could be a bit clearer

seankmartin (commented):
Overall I think this is great, I particularly like the next/previous annotation tool. I have some ideas about the enum property tool, which is that, while I see very much the value in being able to bind each property to a key - I must admit it feels very confusing to me. Especially since I can't bind to numbers. Like if I could bind to option 1, 2, and 3 using the numbers 1, 2, and 3 I think that would be easier to remember my binds.

What I feel would be nicer in my opinion is that I also have an option to bind to the enum property as a whole, not just at each key. Then within that single enum property tool I have the ability to select the property by scroll wheel (if hold tool) or by a second keypress, e.g. numbers or tab (if toggle tool). If it were a toggle tool I could also leave it active while moving through annotations via the next/back tool ideally. This also feels like it would scale to more properties better, as my interaction within each property then becomes very similar and I only need to know the binds to choose the property and not binds per property per value.

I do understand though that you likely have an established workflow in mind, I just wanted to bring this up in case it is useful.

Outside of that I have left some comments about the implementation itself which hopefully help.

src/ui/annotations.ts
chrisj: I agree we should clean it up/refactor, I wrote this a while back. I think it might be worth solving for the case where offset can be any number, I definitely was assuming it would only be -1 or 1 when I wrote it.

> this endless while loop... which layer will have the annotation

The loop here is dealing with multiple layer states

Something like this is clearer
```ts
const entries = this.annotationStates.states.flatMap((annotationLayerState) =>
  Array.from(annotationLayerState.source, (annotation) => ({
    annotation,
    annotationLayerState,
  })),
);
const selectedIndex = entries.findIndex(
  (entry) =>
    entry.annotationLayerState === annotationLayerState &&
    entry.annotation.id === annotationId,
);
 ```

**Comment by fcollman** [comment]
> Overall I think this is great, I particularly like the next/previous annotation tool. I have some ideas about the enum property tool, which is that, while I see very much the value in being able to bind each property to a key - I must admit it feels very confusing to me. Especially since I can't bind to numbers. Like if I could bind to option 1, 2, and 3 using the numbers 1, 2, and 3 I think that would be easier to remember my binds.
> 
> What I feel would be nicer in my opinion is that I also have an option to bind to the enum property as a whole, not just at each key. Then within that single enum property tool I have the ability to select the property by scroll wheel (if hold tool) or by a second keypress, e.g. numbers or tab (if toggle tool). If it were a toggle tool I could also leave it active while moving through annotations via the next/back tool ideally. This also feels like it would scale to more properties better, as my interaction within each property then becomes very similar and I only need to know the binds to choose the property and not binds per property per value.
> 
> I do understand though that you likely have an established workflow in mind, I just wanted to bring this up in case it is useful.
> 
> Outside of that I have left some comments about the implementation itself which hopefully help.

The most common workflow that we have used in the past was that a person has a set of annotations that are laid down already, and now they want to go through them and provide categorical or "tag" labels to each of them. 

As an example, there are a set of cell body locations automatically detected but we don't know what cell type they are, and the user goes through and makes a call as to which cell-type (from some list it is).  

As another example, the user has a list of synapses and they want to annotate whether those synapse have some property (pre-synaptic mitochondria are present, the synapse is on a spine/shaft/soma, the synapse is perforated, etc).  In each case, there is typically just one property they are setting and so minimizing the keystrokes they have to use to set the category is the most important ergonomic factor.  

The auto-advance is something that I took out, but I had it earlier for this same reason,  in that for these kinds of workflows because it's just one less button to hit when doing a single category kind of task.  I understood though that in the context of a more complex schema this would be undesirable. I forgot to remove the comments about the feature, I will do that. 

I'd like to find a way to get the same efficiency of labelling ergonomics we have on https://spelunker.cave-explorer.org/ with the tags feature there. 

I agree the number keys would be ideal, but the tool-bind feature doesn't allow us to map those.  Do we want to change that?

**Comment by chrisj** [comment]
@jbms can confirm the intent but it doesn't seem like shift+number has any behavior now but it was possibly left out because numbers are currently associated with layers

plain number toggles visibility on that layer and ctrl + number switches the side panel to that layer

with the demo Forrest showed, I think numeric binds are fine to use for whatever when a tool is active

**Comment by seankmartin** [comment]
Thanks Forrest, that extra context helps.

> As an example, there are a set of cell body locations automatically detected but we don't know what cell type they are, and the user goes through and makes a call as to which cell-type (from some list it is).

Here I still think that a top level enum property tool could help. I could imagine this tool reminding me what the binds are in the tool status message a bit more clearly. But either way, having the ability to bind per property is still useful and doesn't prevent the potential of having this single tool for an enum property.

> As another example, the user has a list of synapses and they want to annotate whether those synapse have some property (pre-synaptic mitochondria are present, the synapse is on a spine/shaft/soma, the synapse is perforated, etc). In each case, there is typically just one property they are setting and so minimizing the keystrokes they have to use to set the category is the most important ergonomic factor.

Yes, here the ability to bind to one property makes a lot of sense.

> I'd like to find a way to get the same efficiency of labelling ergonomics we have on https://spelunker.cave-explorer.org/ with the tags feature there.

From trying it out the main difference to me with the tags is that I can have multiple tags at the same time. I think that's what I'm trying to get at with the idea of a top level enum tool. If you want to mark multiple different types of enum properties on a single annotation, it feels like the amount of binds you'd need for that doesn't scale well. Again, that doesn't block this PR at all in my opinion. Moreso that I think it could be useful to be able to bind to a single property of an enum, or the whole enum.

> I agree the number keys would be ideal, but the tool-bind feature doesn't allow us to map those. Do we want to change that?

I'd be open to changing it, and as @chrisj said it's probably because numbers are associated with layers but @jbms might have a different reason. To make it simple, I'd suggest we don't bundle that into this PR. I just noticed from using this PR it would seem nice if I could bind tools to numbers, not just a - z. Or possibly even keys outside of that, like left and right or up and down for prev/next annotation (I know these can conflict with neuroglancer binds, so it would be saying that tool binds take precedence. It could be that you get a message that it's bound to action "xyz" are you sure. And if the user wants to override the bind with the tool we assume they have good reason to. It does mean that in particular states a bind might not do what you expect though. But I'd also at some point like to discuss customisable keybinds, at which point this comes up anyway)

**Comment by fcollman** [comment]
I don't think this is polished yet in terms of code, but i wondered @seankmartin if this aligns with what you were thinking in terms of functionality, where you have a enum level tool and then the number keys map to the individual options.

**Comment by seankmartin** [comment]
> I don't think this is polished yet in terms of code, but i wondered @seankmartin if this aligns with what you were thinking in terms of functionality, where you have a enum level tool and then the number keys map to the individual options.

@fcollman yes that's exactly what I was thinking, to me this looks great, and a lot more intuitive. And it's great that I only have to set one bind per property. It's a nice touch that it eliminates holding shift + key to go through the annotations while the tool is open. I'm of course biased since it's how I was thinking about it, but I feel like I can drop into this system more easily and understand it. So I'm unsure how it feels to you and @chrisj who have been using a slightly different system.

This does raise some slightly natural questions like what happens past 10 options:
image
Ehh maybe modifier + num for the next 30 and after that well too bad. Or maybe you can do "number entry" + enter to select values when there is more than 10, like a vim style count prefix on going to a line etc. Perhaps you can indicate the important ones if you want to trim the list. I genuinely don't feel strongly here, but bring it up in case it's relevant.

File: src/layer/annotation/index.ts (hunk 4 above)
Comment (fcollman) [reply src/layer/annotation/index.ts]: removed

File: src/layer/annotation/index.ts (hunk 3 above)
Comment (fcollman) [reply src/layer/annotation/index.ts]: removed

File: src/layer/annotation/index.ts (hunk 5 above)
Comment (fcollman) [reply src/layer/annotation/index.ts]: i think i have consolidated this now

File: src/ui/annotation_schema_tab.ts (hunk 10 above)
Comment (fcollman) [reply src/ui/annotation_schema_tab.ts]: changed

File: src/ui/annotation_schema_tab.ts (hunk 11 above)
Comment (fcollman) [reply src/ui/annotation_schema_tab.ts]: changed

File: src/layer/annotation/index.ts (hunk 7 above)
Comment (fcollman) [reply src/layer/annotation/index.ts]: confirmed this was a bug and fixed

File: src/ui/annotation_schema_tab.ts (hunk 8 above)
Comment (fcollman) [reply src/ui/annotation_schema_tab.ts]: removed

File: src/ui/annotations.ts (hunk 14 above)
Comment (fcollman) [reply src/ui/annotations.ts]: I have incorporated Chris's suggestion

**Comment by fcollman** [comment]
I tried to improve the styling/readability a bit and I have the enum tool switch to numeric input when it is more than 10 items

image

**Comment by seankmartin** [comment]
> I tried to improve the styling/readability a bit and I have the enum tool switch to numeric input when it is more than 10 items
> 
> image

This looks nice, the number input itself seems to cause some issues when interacting with the viewer while having the tool open. I didn't check the code but my guess is that when the number input is there the binds get tied to the input element instead of the tool, so losing focus on the input element loses the binds

**Comment by fcollman** [comment]
> > I tried to improve the styling/readability a bit and I have the enum tool switch to numeric input when it is more than 10 items
> > image
> 
> This looks nice, the number input itself seems to cause some issues when interacting with the viewer while having the tool open. I didn't check the code but my guess is that when the number input is there the binds get tied to the input element instead of the tool, so losing focus on the input element loses the binds

I have added some code to the numeric entry to steal focus back when the user stops interacting with the data viewer, which I think is the most common case.  It's still not a perfect situation in that it prevents the use of other hotkeys when using the numeric entry keys.  

It was @chrisj suggestion to use a generic entry box, but I think this is maybe a reason to go away from that?  really only the numeric keys should be routed to the numeric input tool when it's activated leaving the other letter keys to remain having their normal functions, and this is now how a standard input field works right now where all keypresses go to it when it has focus.

**Comment by seankmartin** [comment]
> I have added some code to the numeric entry to steal focus back when the user stops interacting with the data viewer, which I think is the most common case. It's still not a perfect situation in that it prevents the use of other hotkeys when using the numeric entry keys.
> 
> It was @chrisj suggestion to use a generic entry box, but I think this is maybe a reason to go away from that? really only the numeric keys should be routed to the numeric input tool when it's activated leaving the other letter keys to remain having their normal functions, and this is now how a standard input field works right now where all keypresses go to it when it has focus.

Thanks for the change, this helps, but I'd be inclined to move away from a regular input element as I think it will be more robust and easier to reason about if we don't link what's focused on to what the tool can do

**Comment by chrisj** [comment]
I agree, I don't think it is a requirement but when it can fit, using the standard html elements is good practice

**Comment by fcollman** [comment]
okay i have switched it over to a custom buffer input type which can intercept subsets of keys based on the data type, preserving the rest of the keybinds for their default behavior.  It looks similar to existing entry box. 
image

I also cleaned up the use cases where people have not yet bound previous/next tool keybinds and fixed the bug with the 10th option still showing 0 as the label when there are 11+ options.

**Comment by fcollman** [comment]
i think this test failure is a fluke and could just use a redo

**Comment by chrisj** [comment]
I would like to find a way to have a common way to represent keybinds for tools

Clearly the existing tool bind that yours doesn't use, is not flexible enough for this new tool. Trying to think of a way to generalize it which might mean having the existing UI for all tools look more like this one.

Screenshot 2026-07-29 at 5 41 50 PM

**Comment by seankmartin** [comment]
> I would like to find a way to have a common way to represent keybinds for tools
> 
> Clearly the existing tool bind that yours doesn't use, is not flexible enough for this new tool. Trying to think of a way to generalize it which might mean having the existing UI for all tools look more like this one.
> 
> Screenshot 2026-07-29 at 5 41 50 PM

We've had a similar problem with the edit tool for CATMAID skeletons ever since we brought the merge/split/move etc all into one single tool. The current binds indicator system is not really flexible enough for this, as it shows too many binds, similar to what is here. In addition in the skeleton edit tool, we want the binds shown inside the tool to be conditional. For now we actually just hid the binds indicator, and show our own custom binds indicator - which I imagine is likely similar to what's here (and of course we don't plan to have that forever, it's a workaround). I think these together do motivate exactly as you say, a more flexible system tools can use for tool bind indication.

**Comment by fcollman** [comment]
This makes a lot of sense to me, how do you want to handle this PR?

I can imagine 3 plans..

1) leave the generalization infrastructure for a followup pr
2) develop the generalization infrastructure in this PR but only implement it for this tool, leaving any refactoring of other tools for later
3) develop the generalization interface and replace any custom code in other tools in 1 PR. 

Perhaps if CATMAID is the only other example which isn't yet merged, then option 3 doesn't make much sense.

**Comment by fcollman** [comment]
> I would like to find a way to have a common way to represent keybinds for tools
> 
> Clearly the existing tool bind that yours doesn't use, is not flexible enough for this new tool. Trying to think of a way to generalize it which might mean having the existing UI for all tools look more like this one.
> 
> Screenshot 2026-07-29 at 5 41 50 PM

I would want the showing of those hints to be optional, or have prettier UI elements than that list of text. Perhaps chip buttons that you can click with the mouse and/or press the key that is advertised over them.

**Comment by fcollman** [comment]
> I would like to find a way to have a common way to represent keybinds for tools
> 
> Clearly the existing tool bind that yours doesn't use, is not flexible enough for this new tool. Trying to think of a way to generalize it which might mean having the existing UI for all tools look more like this one.
> 
> Screenshot 2026-07-29 at 5 41 50 PM

I would want to be able suppress the capturing of all the numerical keys in the UI, as calling out those keys individually doesn't make any sense.

**Comment by seankmartin** [comment]
> This makes a lot of sense to me, how do you want to handle this PR?
> 
> I can imagine 3 plans..
> 
> 1. leave the generalization infrastructure for a followup pr
> 2. develop the generalization infrastructure in this PR but only implement it for this tool, leaving any refactoring of other tools for later
> 3. develop the generalization interface and replace any custom code in other tools in 1 PR.
> 
> Perhaps if CATMAID is the only other example which isn't yet merged, then option 3 doesn't make much sense.

You might have already started down a path. Agreed don't think we need 3.

I'd lean 2 honestly, I think this is a pretty motivating example that could help drive the start of it. If it gets too complex we could always extract as the basis for 1. Otherwise I think 1 might just float around a little to be honest.

**Comment by fcollman** [comment]
@seankmartin @chrisj I did a v0 of trying to generalize it, but I don't think it looks that clean/useful right now.

I'm not sure i'm happy about it, so i've made a PR onto this base branch so we review/see the changes independently, but i"d be happy to keep the discussion here. 

https://github.com/AllenInstitute/neuroglancer/pull/2/changes

**Comment by chrisj** [comment]
@seankmartin I think this is ready for review. The only general tool keybind change is an option to disable the default display.

File: src/layer/annotation/index.ts [hunk 15]
```diff
@@ -399,11 +409,42 @@ class LinkedSegmentationLayersWidget extends RefCounted {
 }
 
 const Base = UserLayerWithAnnotationsMixin(UserLayer);
+
+// A keybindable tool that selects the previous annotation in the iteration order.
```
seankmartin: nit: not very useful comment seems clear from class name

src/layer/annotation/index.ts
```diff
@@ -399,11 +409,42 @@ class LinkedSegmentationLayersWidget extends RefCounted {
...
+  }
+  toJSON() {
+    return SELECT_PREVIOUS_ANNOTATION_TOOL_ID;
+  }
+  get description() {
+    return "select previous annotation";
+  }
+}
+
+// A keybindable tool that selects the next annotation in the iteration order.
```
seankmartin: nit: not very useful comment seems clear from class name

src/layer/annotation/index.ts
```diff
@@ -399,11 +409,42 @@ class LinkedSegmentationLayersWidget extends RefCounted {
...
+  }
+}
+
 export class AnnotationUserLayer extends Base {
   localAnnotations: LocalAnnotationSource | undefined;
   codeVisible = new TrackableBoolean(true);
   hideInactiveShaderControls = new TrackableBoolean(false);
-  private localAnnotationProperties: WatchableValue<AnnotationPropertySpec[]> =
+  // Accessed by the property keybind tools (e.g. EnumClassificationTool) to
+  // observe schema changes while active.
```
seankmartin: comment seems misleading, suggest to remove

src/layer/annotation/tool_state.ts
```diff
@@ -0,0 +1,56 @@
+/**
```
seankmartin, line 1: The name of this file seems a bit narrow for what it does - would currently lean to have this for just IDs and move the helper functions and property tool interface to property_tools.ts

Alternatively, could change the file name to indicate the broader functionality.

src/ui/annotation_schema_tab.ts
```diff
@@ -331,7 +361,7 @@ class AnnotationUIProperty extends RefCounted {
       const rawValue = (event.target as HTMLInputElement).value;
       // Replace dash and spaces with underscores
       let sanitizedValue = rawValue.replace(/-/g, "_");
-      sanitizedValue = rawValue.replace(/\s+/g, "_");
+      sanitizedValue = sanitizedValue.replace(/\s+/g, "_");
```
seankmartin, line 364: Thanks for spotting that one!

seankmartin (approved):
Looks great thanks @chrisj and @fcollman! I left some minor comments, which I think would be good to do but approving as I agree on the whole so imo can be merged after that

seankmartin:
@chrisj since this one was opened by @fcollman, I know you also worked on it though. Would you prefer a review from @jbms ideally if possible before merging? Or are you happy to be the second reviewer on this one?

PR #1067, opened 2026-07-06
feat: make the tool status be the bottom element with new messages at top

seankmartin:
In #991 we use a tool which is on the screen very commonly, and the status message system to inform about edits made. The kind of thing like - if a merge is performed correctly, or if the merge fails, or if you have the wrong permissions to move a node.

When this happens the tool message visually shifts around because the status messages are getting appended to the bottom of the list. This PR aims to change this behaviour, so tools are always appended at the bottom of the element list, and messages instead get prepended and put as the first element in the list.

So while you could previously get this where the tools are in between messages, or the message shows below the tool and moves the tool up visually
image

Now you always get it like:
image
with the tool fixed at the bottom and messages above.

I'm making this PR as a general one because I think it could be helpful for review purposes of #991 if these kind of general changes are not baked inside that PR and can be more easily discussed in smaller pieces where possible. And because if other tools are ever using the same system, it could be helpful.

### DIFF
--- a/src/status.ts
+++ b/src/status.ts
@@ -77,14 +77,14 @@ export class StatusMessage {
       event.preventDefault();
     }
   };
-  constructor(delay: Delay = false, modal = false) {
+  constructor(delay: Delay = false, modal = false, insertAtBottom = false) {
     const element = document.createElement("li");
     this.element = element;
     element.addEventListener("mousedown", this.handleMouseDown);
     if (delay === true) {
       delay = DEFAULT_STATUS_DELAY;
     }
-    this.setModal(modal);
+    this.setModal(modal, insertAtBottom);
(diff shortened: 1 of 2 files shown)


src/status.ts
```diff
@@ -77,14 +77,14 @@ export class StatusMessage {
       event.preventDefault();
     }
   };
-  constructor(delay: Delay = false, modal = false) {
+  constructor(delay: Delay = false, modal = false, insertAtBottom = false) {
```
seankmartin, line 80: Depending on how widely used this is, it may be better to instead refactor to an options object

chrisj (approved):
@seankmartin I'm good with this as is if you want to go ahead

PR #1070, opened 2026-07-10
Feature/annotation measurements

chourroutm:
I have added measurements for the different types of annotations:

**Line:**
- Length
- Projected lengths on x/y/z 

image

**Bounding box:**
- Edge lengths
- Volume

image

**Polyline:**
- Total length
- Angles 

image


chrisj:
@chourroutm this is a valuable feature and I will give you more implementation feedback later

I worked on a similar feature and it stalled out but I recommend taking a look at the implementation: https://github.com/google/neuroglancer/pull/694. It added "default annotation properties" and I think that is a good idea for your implementation as well.


It would be great to get some feedback from scientists/users but I'm wondering what is the right data to show and perhaps there should be an expanded and collapsed version of the metadata.
Screenshot 2026-07-16 at 12 26 46 PM

[example](https://neuroglancer--pr1070-w3aa2kbp.web.app/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B17213.19140625%2C19861.625%2C20697.5%5D%2C%22crossSectionScale%22:1%2C%22projectionScale%22:65536%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22gs://neuroglancer-janelia-flyem-hemibrain/emdata/clahe_yz/jpeg/%7Cneuroglancer-precomputed:%22%2C%22tab%22:%22annotations%22%2C%22name%22:%22jpeg%22%7D%2C%7B%22type%22:%22annotation%22%2C%22source%22:%7B%22url%22:%22local://annotations%22%2C%22transform%22:%7B%22outputDimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%7D%7D%2C%22tool%22:%22annotatePolyline%22%2C%22tab%22:%22annotations%22%2C%22annotations%22:%5B%7B%22points%22:%5B%5B17018.5%2C19780.5%2C20697.5%5D%2C%5B17334.5%2C19594.5%2C20697.5%5D%2C%5B17362.5%2C19878.5%2C20697.5%5D%2C%5B17234.5%2C20070.5%2C20697.5%5D%2C%5B17022.5%2C19968.5%2C20697.5%5D%2C%5B17234.5%2C19803.5%2C20697.5%5D%5D%2C%22type%22:%22polyline%22%2C%22id%22:%223d7ab1fec2e4fc4091e5239339431700288f19ca%22%7D%5D%2C%22name%22:%22annotation%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22annotation%22%7D%2C%22layout%22:%22xy%22%2C%22selection%22:%7B%22layers%22:%7B%22annotation%22:%7B%22annotationId%22:%223d7ab1fec2e4fc4091e5239339431700288f19ca%22%2C%22annotationSource%22:0%2C%22annotationSubsource%22:%22default%22%7D%7D%7D%7D)

chourroutm:
Thanks, @chrisj! Should I change something in my implementation?

src/ui/annotations.ts
```diff
@@ -1642,6 +1648,108 @@ class PlacePolylineTool extends MultiStepAnnotationTool {
   }
 }
 
+/**
+ * Builds a per-annotation-dimension physical scale array (nanometers per unit)
+ * for the given annotation layer. A dimension whose unit is not length (meters)
+ * contributes 0 so it does not affect the measured length. Returns undefined if
+ * the layer transform is not yet resolved.
+ */
+function getAnnotationPhysicalScales(
```
Le0C, line 1657: I wrote a proposed test for this function here: https://github.com/Le0C/neuroglancer/blob/24bb4640/src/ui/annotations.browser_test.ts

src/util/ruler_length.ts
```diff
@@ -0,0 +1,60 @@
...
+): number {
+  if (points.length < 2) return 0;
+  let total = 0;
+  for (let i = 0; i + 1 < points.length; ++i) {
+    const a = points[i];
+    const b = points[i + 1];
+    const rank = Math.min(a.length, b.length);
+    let sumSquares = 0;
+    for (let d = 0; d < rank; ++d) {
+      const scale = d < scaleNm.length ? scaleNm[d] : 1;
```
Le0C, line 53: If a caller ever passes a `scaleNm` shorter than the point rank, the dimensions past the end would be treated as 1nm per unit, so this would mix nanometre-scaled and unit-scaled deltas. This `scale` variable is also in `axis_extents.ts` and `angle_measurement.ts`

- `scaleNm[d]` is `undefined`
- `undefined === 0` is false, so the guard "skip non-length axes" (used in two places in annotations.ts) would not fire. If I understand correctly you would get a "Length (axis)" row  computed at 1nm/unit rather than no row.

I don't think anything in the current code would run into this, but I think a test to pin the desired behaviour would be good.

Test here: https://github.com/Le0C/neuroglancer/blob/24bb4640/src/util/ruler_length.spec.ts#L70-L83

src/util/ruler_length.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  // C3: a duplicate consecutive point contributes 0
+  it("treats a zero-length segment as contributing nothing", () => {
+    const points = [p(1, 1, 1), p(1, 1, 1), p(1, 1, 4)];
+    expect(computeRulerPhysicalLength(points, [1, 1, 1])).toBeCloseTo(3, 6);
+  });
+
+  // C4: anisotropic per-axis scale
+  it("applies the per-axis physical scale", () => {
```
Le0C, line 53: `axis_extents.spec.ts:53` pins the behaviour of a zero-scale (non-length) axis. `ruler_length.spec.ts` and `angle_measurement.spec.ts` have no equivalent case currently.

In `getAnnotationPhysicalScales` the scale is left at 0 for any dimension which is not `"m"`. If ever it didn't do that, the extra axis would get pulled in, so a length would take on part of its span through the norm and an angle would move depending on which direction the two arms lean along it

Added a test for [length](https://github.com/Le0C/neuroglancer/blob/24bb4640/src/util/ruler_length.spec.ts#L61-L68) & [angle](https://github.com/Le0C/neuroglancer/blob/24bb4640/src/util/angle_measurement.spec.ts#L99-L109) here

Le0C (commented):
Hello 👋  this is a good feature, thanks for building it. I have built a similar feature in the fork I maintain. I was able to replace all of the measurement maths in my fork with your helpers, without any modification, and all of your tests passed against my forks code.

I have forked your branch to add some suggested tests, in three places:

- `getAnnotationPhysicalScales`
- `src/util/ruler_length.ts` 
- `src/util/ruler_length.spec.ts`

That fork is here: https://github.com/chourroutm/neuroglancer/pull/1

Issue #1075, opened 2026-07-14
OME-Zarr source: color metadata is ignored when there's only one channel


m-albert:
Hi! I wanted to report the following problem which I believe is a bug:

When loading OME-Zarr data with a single channel, Neuroglancer displays the channel in white instead of using the color from the OMERO metadata. The same metadata works correctly when the dataset contains multiple channels.

Reproduce using the example dataset: https://ssbd.riken.jp/100118-dcacbb41/zarr/v0.4/Fig3a_FIB-SEM_synapse.zarr. Here although the omero channel in .zattrs indicates "808080" for the colormap, neuroglancer loads the dataset in white.

The likely cause is in src/layer/multi_channel_setup.ts, where single-channel layers are assigned white before the metadata color is checked.

Thanks a lot in advance for looking into this!

seankmartin:
Thanks for making this one @m-albert, really appreciate the very clear bug report. I've pushed what I think should fix in #1076, please let me know if not and I'll take another look

m-albert:
@seankmartin Oh wow that was a quick fix :) Thank you so much! 

I tested the PR branch and the channel coloring works as expected 🚀

PR #1076, opened 2026-07-16
fix: single layer multi-channel setup uses omero color

seankmartin:
This changes the color precedence to be omero channel color, then white if only one channel, then default color array (loops through RGBW). Previously it was mistakenly choosing white if only one channel over the color in the metadata.

Should address https://github.com/google/neuroglancer/issues/1075


m-albert:
Thank you @seankmartin, this is great 🙏🚀

I tested the PR branch and channel coloring seems to work as expected!

PR #1078, opened 2026-07-17
perf(volume_rendering): faster volume rendering depth check in fragment shader

seankmartin:
### Summary
Improves fragment shader performance of depth testing in volume rendering, adds a new test for volume rendering occlusion, and updates an existing volume rendering test to check both perspective and orthographic projections.

### Motivation
During volume rendering, we currently compute the depth along the ray at each ray march step in the fragment shader. This computed depth is compared to the depth value in the depth buffer, to see if the ray has passed behind an opaque object. This allows opaque objects to block volume rendering rays. Without this feature, you visually see an odd mixing between opaque objects and volume rendering. However, there are two parts of this which could be more efficient.

### Changes
1. **Finding the current ray sample position in clip space.**
*Previously* - Apply the `uModelViewProjectionMatrix` to the ray position at each ray march step in model space to get that position in clip space. In other words, interpolate in model space and then transform to clip space.
*Here* - Apply the `uModelViewProjectionMatrix` to the near position of the ray and the far position of ray (akin to ray entry and exit points) and then at each ray march step, interpolate along the vector in clip space to find the position of the current sample. In other words, transform the ray start and end points into clip space and then interpolate in clip space.
*Reason* - This interpolation is more efficient than a matrix multiplication each step. We can do this because `MVP·mix(near, far, t) = MVP·((1-t)·near + t·far) = (1-t)·MVP·near + t·MVP·far = mix(MVP·near, MVP·far, t)` due to additivity and homogeneity of degree 1 properties of the matrix multiplication here. This works as long as we do it before perspective divide.
2. **Extracting the depth from the buffer for the current ray sample.**
*Previously* - Map the position of the current sample in clip space to UV coordinates and then sample from the depth buffer at that UV value.
*Here* - Before starting the ray march, sample from the depth buffer once using the clip space position of the position passed from the vertex shader to the fragment shader.
*Reason* - Since positions along a ray project back to the same fragment location, we know that the UV value must be constant along the ray. So we don't need to resample from the depth buffer texture at each ray march step.

### DIFF
--- a/python/tests/volume_rendering_test.py
+++ b/python/tests/volume_rendering_test.py
@@ -20,7 +20,7 @@ import numpy as np
 import pytest
 
 
-def _setup_viewer_with_volume_and_mesh(webdriver, orthographic):
+def _setup_viewer_with_volume_and_mesh(webdriver, orthographic, mesh_opacity=1.0):
     shape = (20, 20, 20)
 
     # Image data in the far half of the volume (high z = far from default camera,
@@ -60,6 +60,7 @@ void main() {
                 segments=[1],
                 segment_default_color="#ff0000",
                 hover_highlight=False,
+                object_alpha=mesh_opacity,
             ),
         )
         s.layout = "3d"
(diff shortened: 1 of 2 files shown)


fcollman:
does this fit into the sub-system framework @jbms @chrisj ?

jbms (approved):

PR #1081, opened 2026-07-22
segment coloring with batching

chrisj:


fcollman:
@chrisj  is this still a draft? it's not getting reviewed I think because it is marked draft.

PR #1084, opened 2026-07-27
fix: recompute default position when dimension order changes

AKnassa:
### What this does

When you load a saved view that has no position saved in it, Neuroglancer picks a sensible starting point for you — the middle of the data. This makes sure it works that out fresh for the view you're actually loading, instead of reusing the one it picked for the previous view.

### Why

If the new view lists its dimensions in a different order than the old one (say `z, x, y` instead of `x, y, z`), the old starting point gets applied to the wrong axes. You end up far outside the data, looking at an empty screen. That's what #885 reports — it shows up the second time you press apply in the JSON state editor, because the first apply leaves a position behind.

### What changed

01-master-apply1-loading
02-master-apply1-CORRECT
03-master-apply2-BROKEN-z-out-of-bounds
04-fixed-apply1-loading
05-fixed-apply1-CORRECT
06-fixed-apply2-STILL-CORRECT

- Neuroglancer now keeps track of which starting points it worked out for you, versus which ones you chose yourself — by saving them in a link, or by panning and zooming.
- Positions you chose are always kept, exactly as before, including ones deliberately outside the data.
- Only the automatically-chosen ones get recalculated when a view is reloaded.
- Added 17 tests: the reported case, plus edge cases for reordering, rank changes, panning, unbounded dimensions, and rounding.

### How to see it

Run `npm run dev-server`, open the viewer, and in the browser console apply the state from #885 twice:

```javascript
viewer.state.restoreState(state);              // loads correctly
viewer.state.reset(); viewer.state.restoreState(state);   // the second apply
```

Checked against the dataset in the issue (`fafb_v14`, bounds `z: 7063`, `x: 248832`, `y: 134144`):

- **Before:** position `[124416.5, 67072.5, 3531.5]` — `z` is `124416.5` against a `z` limit of `7063`. Outside the data, blank panels.
- **After:** position `[3531.5, 124416.5, 67072.5]` — inside the data.

### Note for reviewers

This tracks a *copy* of the inferred coordinates rather than a boolean flag, because several callers move the position by writing into the array returned by `position.value` and dispatching `changed` directly, never going through the setter — `NavigationState.translateVoxelsRelative` and the slice-view zoom-at-mouse handler in `src/sliceview/panel.ts`. A flag would go stale on pan or zoom and silently discard the move.

There is a plausible alternative I'd be glad to switch to if you prefer it: remember the last *valid* coordinate space and remap by dimension id across the invalid gap. That would carry a panned position through a reorder rather than re-centring it.

Fixes #885

### DIFF
--- a/src/navigation_state.ts
+++ b/src/navigation_state.ts
@@ -169,12 +169,18 @@ function makeSimpleLinked<T extends RefCounted & { changed: NullarySignal }>(
 export class Position extends RefCounted {
   private coordinates_: Float32Array = vector.kEmptyFloat32Vec;
   private curCoordinateSpace: CoordinateSpace | undefined;
-  // Indicates that `coordinates_` was inferred from the bounds of
-  // `curCoordinateSpace` rather than specified explicitly.  Inferred
+  // The coordinates that were inferred from the bounds of `curCoordinateSpace`,
+  // or `undefined` if the coordinates were specified explicitly.  Inferred
   // coordinates are only meaningful for the coordinate space they were computed
   // from, and must be recomputed rather than reused if the coordinate space is
   // replaced while invalid.
-  private coordinatesAreDefault_ = false;
+  //
+  // This is a copy of the coordinates rather than a boolean flag because
+  // several callers, such as `NavigationState.translateVoxelsRelative`, move
+  // the position by writing into the array returned by `value` instead of going
+  // through the setter.  Comparing against the copy detects those moves as
+  // well.
+  private inferredCoordinates_: Float32Array | undefined;
   changed = new NullarySignal();
   constructor(
     public coordinateSpace: WatchableValueInterface<CoordinateSpace>,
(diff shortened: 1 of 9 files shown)


src/navigation_state.ts
```diff
@@ -169,6 +169,18 @@ function makeSimpleLinked<T extends RefCounted & { changed: NullarySignal }>(
...
+  // or `undefined` if the coordinates were specified explicitly.  Inferred
+  // coordinates are only meaningful for the coordinate space they were computed
+  // from, and must be recomputed rather than reused if the coordinate space is
+  // replaced while invalid.
+  //
+  // This is a copy of the coordinates rather than a boolean flag because
+  // several callers, such as `NavigationState.translateVoxelsRelative`, move
+  // the position by writing into the array returned by `value` instead of going
+  // through the setter.  Comparing against the copy detects those moves as
+  // well.
```
chrisj: it might be best to update those usages because that still opens up the case for one of those callers to navigate away and then back to the same value as inferredCoordinates_ and have it treated as coordinatesAreInferred.

Whereas in your setter, you purposely clear out inferredCoordinates_. However maybe that usage isn't relevant.

AKnassa: good catch, this turned out to be relevant. stepping rounds to the voxel centre, so an arrow key away and back lands exactly on inferredCoordinates_ and got treated as inferred again. I updated those usages: the in-place movers now report via markMoved (only when a coordinate actually changed, so no-op writes still re-infer), and the two channel widgets just go through the setter now. kept the comparison as a fallback for any caller that doesn't report. added tests driving each caller, including the away-and-back cases.

PR #1088, opened 2026-07-28
docs: fix segment selection in demo notebook

### LINKED ISSUE #609: Jupyter Notebook Demo outdated and causes errors
(Amazing project!)

I was trying to work through https://github.com/google/neuroglancer/blob/master/python/examples/jupyter-notebook-demo.ipynb, and discovered that some of the code no longer works. In particular,

```python
with viewer.txn() as s:
    s.layers['segmentation'].segments.update([1752, 88847])
    s.layers['segmentation'].visible = True
```

causes the following error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[13], line 2
      1 with viewer.txn() as s:
----> 2     s.layers['segmentation'].segments.update([1752, 88847])
      3     s.layers['segmentation'].visible = True

AttributeError: 'VisibleSegments' object has no attribute 'update'
```

AKnassa:
Fixes #609

**What this does**

Fixes one broken line in the Jupyter notebook demo.

**Why**

The "Select a couple segments" cell called `segments.update([...])`. That stopped working when `layer.segments` became a set-like view of the starred segments, so running the cell raised `AttributeError: 'VisibleSegments' object has no attribute 'update'`.

**What changed**

- One line: `segments.update([1752, 88847])` becomes `segments |= [1752, 88847]`, which the set view does support and which accepts any iterable.

`update` itself still exists on `StarredSegments`, if starring rather than selecting is what is wanted.

**How to see it**

Run the notebook. The cell now selects the two segments instead of raising an error. This was the only stale call in the notebook — the later `segments.add` cell is still correct.

### DIFF
--- a/python/examples/jupyter-notebook-demo.ipynb
+++ b/python/examples/jupyter-notebook-demo.ipynb
@@ -167,7 +167,7 @@
    "outputs": [],
    "source": [
     "with viewer.txn() as s:\n",
-    "    s.layers['segmentation'].segments.update([1752, 88847])\n",
+    "    s.layers['segmentation'].segments |= [1752, 88847]\n",
     "    s.layers['segmentation'].visible = True"
    ]
   },


seankmartin (approved):
Thanks for the change, looks reasonable to me, but alternatively we could add back in `update` as a method on VisibleSegments to the Python API if @jbms would prefer

PR #1091, opened 2026-07-30
design draft: adds a status warning when neuroglancer hits memory limits

fcollman:
Recently that Allen Institute was hosting a set of 30 educators who were learning to teach neuroscience using neuroglancer.  During the workshop, one point of confusion popped up when they ended up in states of neuroglancer which bumped up into their memory limits, and they were unaware of what was happening and why data was not loading.  There is no clear visual indication presently of when this is happening, or what the fix might be (i.e. increasing memory limits)

This PR is one suggestion for how we might mitigate this issue for new users. 

A different or complementary idea would be adding a visual indicator to the status bar for a layer (say turning it red) when that layer has stopped loading data because of memory limits as opposed to waiting for responses from the server.

Interested what people think of this from a design perspective versus other ideas.

### DIFF
--- a/src/chunk_manager/backend.ts
+++ b/src/chunk_manager/backend.ts
@@ -30,6 +30,7 @@ import {
   ChunkState,
   getChunkDownloadStatisticIndex,
   getChunkStateStatisticIndex,
+  MemoryLimitFlags,
   numChunkMemoryStatistics,
   numChunkStatistics,
   REQUEST_CHUNK_STATISTICS_RPC_ID,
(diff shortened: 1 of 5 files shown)


tarelli:
@syamace-metacell could you take a look?

seankmartin:
Overall it seems pretty reasonable to me, and a global indicator about visible chunks seems a good starting point. If it's of any use you might find https://github.com/MetaCell/neuroglancer/pull/158 interesting where we tried out one strategy for per layer memory pressure indicators of interest.

It still suffers a bit from what we chatted about before which is that some layers can degrade well while others can't. For e.g. a 2D multi-scale image slice can degrade quite well. If possible to share, which type of data were the users hitting the limit on?

chrisj:
@seankmartin you showed that to us a while back right? We were discussing that as an alternative.

I was hoping that a per layer indicator could lead us to better prioritization of memory. Perhaps user configurable depending on their workflow. For example, I would imagine it would be generally good to prioritize segmentation over image.

syamace-metacell:
Agree here, a global indicator per the PR is definitely a good start, already solving that user confusion. From a design perspective, a per layer indicator would be ideal to give the user better and more accurate indicators of exactly which data has stopped loading due to memory limits - extending on the loading status bar is natural and what I'd suggest too + a tooltip as to why it turned orange/red. Something like this
CleanShot 2026-08-05 at 14 00 46@2x

PR #1094, opened 2026-08-11
fix(drag_and_drop): Fix drag and drop issues under Wayland

jbms:
Under Wayland, neither Firefox nor Chrome report changes to modifier keys during
drag events.  Chrome reports modifiers for the initial `dragstart` event and no
modifiers for the subsequent events (https://issues.chromium.org/issues/40138974).
Firefox saves the modifiers that were pressed during the initial `dragstart` and
continues to report the same modifiers for all subsequent events.

Previously, Neuroglancer determined the drop effect based on the modifier keys
reported to the `dragover` event, and under Linux Wayland on Chrome no modifiers
are ever reported.  With this change, the Firefox behavior under Wayland is
essentially emulated on Chrome: the modifiers reported to `dragstart` are saved
and apply throughout the drag event (even if the user later releases them).

This commit also updates the drag status message to indicate the
Wayland-specific limitations when Wayland is detected.

Additionally, Chrome under Wayland sets `effectAllowed` to `copyMove` rather
than `all` by default.  Previously, in cases such as dropping a layer onto a
drop zone in order to create a new layer group, Neuroglancer selected a default
dropEffect of `link`, which led to the drop being rejected by Chrome.  This
commit fixes that problem by reporting a fake dropEffect, if necessary, to
ensure it is one of the values allowed by `effectAllowed`.  Regardless of the
reported drop effect, the correct drop operation is still performed.

### DIFF
--- a/src/util/drag_and_drop.ts
+++ b/src/util/drag_and_drop.ts
@@ -81,6 +81,29 @@ export function decodeParametersFromDragTypeList(
 
 let savedDropEffect: DataTransfer["dropEffect"] | undefined;
 
+// Chrome on Wayland seems to report an `effectAllowed` of `copyMove` regardless
+// of what the dragstart handler sets.  Convert the actual drop effect to an
+// allowed value to avoid Chrome rejecting the drop.  The actual drop effect is
+// still stored separately by `setDropEffect` to ensure the correct drop action
+// is performed.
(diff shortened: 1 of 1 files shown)


seankmartin (approved):
Looks good thanks

PR #1096, opened 2026-08-13
Synchronized cross-section hover cursor + move-layers-between-panels   by dragging

cgoina:
## Summary

This branch adds two independent cross-section cursor / layer-panel usability features plus some unit tests. I implemented them as part of the same PR because for the cross-section cursor I had a hard time moving an existing layer under a different root. Initially when I it kept the layer in both section which was confusing when IU under both roots which moved it kept 2 copies which for me created a lot of confusion especially if I had to adjust the rendering controls for the layer. I played with a layer-tab context menu but I found dragging more intuitive.

## 1. Synchronized cross-section hover cursor

A toggle (`showCrossSectionHoverPosition`) that echoes the mouse-hover position from the cross-section panel. I found this feature useful when I check registered images - sometime is more convenient being able to hover over the fixed and moving images simultaneously  instead of toggling the corresponding layers on/off.

- Bound to the **`c`** key and exposed in the viewer settings dialog.
- Only has an effect when two or more cross-section panels are shown side by side; a single panel never draws it (the active cursor is always in that panel).
- Serialized in the viewer state, the Python API, and the JSON schema.

## 2. Move layers between panels (and back to the master layer list) by dragging

- Fixes drag-and-drop of layer tabs across split (multi layer-group) layouts:

- Dragging a layer tab from one group panel to another now **moves** it (removes it from the panel it came from) by default, instead of linking it into both. Hold **Ctrl** to link (show in both panels), **Shift** to copy.
- Dragging a layer from a group panel onto the master **Layer List** side panel is now allowed and removes it from its source group — mirroring the layer tab's
    × close button: it is archived if it was its last group, and left untouched in any other groups that still show it. Previously this was silently a no-op / disallowed.

### DIFF
--- a/docs/json_schema/viewer_state.yml
+++ b/docs/json_schema/viewer_state.yml
@@ -41,6 +41,12 @@ properties:
     type: boolean
     title: "Indicates whether to show the red/green/blue axis lines."
     default: true
+  showCrossSectionHoverPosition:
+    type: boolean
+    title: |
+      Indicates whether to show the position hovered in one cross-section panel
+      as a marker in the other cross-section panels.
+    default: false
   wireFrame:
     type: boolean
     title: "Indicates whether to enable wireframe rendering mode (for debugging)."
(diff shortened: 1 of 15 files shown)


jbms:
Please split this into two PRs since they are entirely different features.

cgoina:
closing this - I will create 2 PRs

PR #1099, opened 2026-08-25
fix find path with new lvl2ids

fcollman:
when the find path tool runs with precision mode on a newly formed ID, it is likely a subset of level2graph nodes might not have attributes yet, but this fixes a problem with the current implementation where the makes all the nodes not use the pcgl2cache point, rather than just the ones that weren't included in the response.

### DIFF
--- a/src/datasource/graphene/frontend.ts
+++ b/src/datasource/graphene/frontend.ts
@@ -2127,13 +2127,26 @@ class GrapheneGraphSource extends SegmentationGraphSource {
         // maybe a small fraction have no info
         // sometime l2 is so small (single voxel), it is ignored by l2
         // best to just drop those points
-        centroids = l2_path
+        // note that the l2 cache reports a missing id as an empty object
+        // rather than omitting it, so an absent rep_coord_nm also means
+        // missing; treating it as malformed would throw away the whole
+        // refinement over a single uncached id (common right after an edit)
+        const refined = l2_path
           .map((id) => {
             return verifyOptionalObjectProperty(attributes, id, (x) => {
-              return verifyFloatArray(x["rep_coord_nm"]);
+              return verifyOptionalObjectProperty(
+                x,
+                "rep_coord_nm",
+                verifyFloatArray,
+              );
             });
           })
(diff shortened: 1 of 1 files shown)


src/datasource/graphene/frontend.ts
```diff
@@ -2127,13 +2127,26 @@ class GrapheneGraphSource extends SegmentationGraphSource {
         // maybe a small fraction have no info
         // sometime l2 is so small (single voxel), it is ignored by l2
         // best to just drop those points
-        centroids = l2_path
+        // note that the l2 cache reports a missing id as an empty object
+        // rather than omitting it, so an absent rep_coord_nm also means
```
chrisj: does the l2 cache sometimes omit ids? If not, we can replace with `verifyOptionalObjectProperty` on the id with `verifyObjectProperty`

fcollman: it will omit IDs for which it has not calculated properties yet.  In particular, right after an edit is done, the properties are not yet calculated.  So when doing a multi-soma split, and the find path tool is automatically retriggered when there is a split and the result does not produce two IDs, then there is going to likely be a level2 node that does not yet have a property, and that results in the entire path getting downgraded to a coarse path, when really just an individual point could be left off or replaced with a the chunk level approximation of the point.

chrisj: the current code does expect that some ids will not exist but it does expect that for every id, there will be a "rep_coord_nm". If not, the entire path will be thrown away. But the comment in the code implies that missing ids are always included as empty objects so it seems like it makes sense to change to optional for "rep_coord_nm" but also change the previous optional on the id to required. But if we want both optional for defensive purposes or for later changes to the l2 cache api, I think that is fine.

fcollman: i have made the outer ID required, and the inner attribute optional, and then made it fallback to the chunk based coordinate if an individual point is missing from the pcgl2cache response, rather than falling back the entire path.

chrisj (approved):

PR #1100, opened 2026-08-25
feat: add support for input/output in ome transforms as object

seankmartin:
Previously from the transforms RFC the `input/output` field of a transform was a string, where the string could be the name or the path and based on the location of the transform the appropriate meaning was inferred. This has changed in the ome spec so that the `input/output` is an object with `name` and `path` as optional strings. Generally speaking only one of `name` and `path` is intended to be filled for each `input/output`. For example, the `path` is expected to be filled on the `input` while `name` is expected to be filled on the `output` of transforms inside of `multiscales.datasets`. This PR is to update support in neuroglancer for that.

We continue to allow strings because we support version `0.6.dev1` which was built to the old spec. If there is a string input or output, we interpret that as an object with `name` and `path` both filled in to the value of the provided string. Because of this we also remain slightly looser than some of the MUST clauses in the spec around input and output, because of how we implemented the initial support. In summary this leaves us compliant when provided valid metadata, and we do verify a lot of the key metadata for correctness, but we don't enforce compliance with some MUST clauses for every part of the spec. 

In addition it also adds a new way of finding the intrinsic coordinate space and output coordinate space. Since the ome spec now includes a way to determine the intrinsic space as the space that is named as the output of each transform in the multiscales.datasets, we can iterate the names to verify all have the same output name - and if so, use that as the intrinsic space.

We don't follow the recommendation from the spec (not a MUST) that suggests viewers to use the intrinsic space unless instructed otherwise. Since we don't have a coordinateSpace selector, we follow the existing code in neuroglancer which applies the `multiscales.coordinateTransforms`, if present, in order. So our output coordinateSpace is the intrinsic space if `multiscales.coordinateTransforms` is not present. Otherwise it is the coordinateSpace which has the same name as the `output` from the last transform in `multiscales.coordinateTransforms` - which may be the intrinsic space still.

### DIFF
--- a/src/datasource/zarr/ome.ts
+++ b/src/datasource/zarr/ome.ts
@@ -549,6 +549,48 @@ function parseMultiscaleOutput(obj: unknown) {
   return parseOmeInputOutput(transformationInputs[0]);
 }
 
+// Find each coordinate transform with an input of the intrinsic system
+// as those are the only ones which could be applied directly
+function findTransformationsWithIntrinsicInput(
+  transformations: unknown,
+  intrinsicCoordinateSystemName: string,
+): { transformation: unknown; outputCoordinateSystemName: string }[] {
+  if (transformations === undefined) return [];
+  const parsed = parseArray(transformations, (transformation) => {
+    verifyObject(transformation);
+    return {
+      transformation,
+      input: verifyOptionalObjectProperty(
+        transformation,
+        "input",
+        parseOmeInputOutput,
+      ),
+      output: verifyOptionalObjectProperty(
+        transformation,
+        "output",
+        parseOmeInputOutput,
+      ),
+    };
+  });
+  const transforms = [];
+  for (const { transformation, input, output } of parsed) {
+    if (input?.name !== intrinsicCoordinateSystemName) continue;
+    if (output?.name === undefined) continue;
+    // Avoid finding child labels groups
+    if (
+      output.path !== undefined &&
+      output.path.includes("labels") &&
+      output.path !== output.name
+    )
+      continue;
+    transforms.push({
+      transformation,
+      outputCoordinateSystemName: output.name,
+    });
+  }
+  return transforms;
+}
+
 function parseMultiscaleScale(
   rank: number,
   url: string,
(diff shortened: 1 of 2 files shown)


chrisj:
can you add a test to ome.spec.ts for this?

seankmartin:
> can you add a test to ome.spec.ts for this?

Yes I added a test and updated a few others to better reflect the status of the transforms spec

src/datasource/zarr/ome.ts
```diff
@@ -471,35 +461,35 @@ function validateCoordinateTransformations(
...
-            verifyString,
+            parseOmeInputOutput,
           );
 
           // First transform in sequence should have input matching the sequence's input
           if (
             i === 0 &&
             innerInput !== undefined &&
-            innerInput !== expectedInput
+            innerInput.name !== expectedInput
```
chrisj: I believe `.path` should be checked instead of `.name` as the following valid metadata is rejected as it only uses path to identify the input
```json
{
  "type": "sequence",
  "input": { "path": "s0" },
  "output": { "name": "physical" },
  "transformations": [
    {
      "type": "scale",
      "scale": [2, 2],
      "input": { "path": "s0" },
      "output": { "name": "scaled" }
    },
    {
      "type": "translation",
      "translation": [10, 20],
      "input": { "name": "scaled" },
      "output": { "name": "physical" }
    }
  ]
}
```

chrisj:
This is another edge case that seems to be an issue

```json
{
  "ome": {
    "version": "0.6",
    "multiscales": [
      {
        "coordinateSystems": [
          {
            "name": "physical",
            "axes": [
              { "name": "y", "type": "space", "unit": "micrometer" },
              { "name": "x", "type": "space", "unit": "micrometer" }
            ]
          },
          {
            "name": "world",
            "axes": [
              { "name": "y", "type": "space", "unit": "micrometer" },
              { "name": "x", "type": "space", "unit": "micrometer" }
            ]
          }
        ],
        "datasets": [
          {
            "path": "s0",
            "coordinateTransformations": [
              {
                "type": "identity",
                "input": { "path": "s0" },
                "output": { "name": "physical" }
              }
            ]
          }
        ],
        "coordinateTransformations": [
          {
            "type": "translation",
            "input": { "name": "world" },
            "output": { "name": "physical" },
            "translation": [10, 20]
          }
        ]
      }
    ]
  }
}
```

the coordinateTransform should be ignored when outputing s0 to physical as s0 is already in the physical coordinate system.

src/datasource/zarr/ome.ts
seankmartin: Agreed, great spot thank you for that

seankmartin:
> This is another edge case that seems to be an issue
> 
> ```json
> {
>   "ome": {
>     "version": "0.6",
>     "multiscales": [
>       {
>         "coordinateSystems": [
>           {
>             "name": "physical",
>             "axes": [
>               { "name": "y", "type": "space", "unit": "micrometer" },
>               { "name": "x", "type": "space", "unit": "micrometer" }
>             ]
>           },
>           {
>             "name": "world",
>             "axes": [
>               { "name": "y", "type": "space", "unit": "micrometer" },
>               { "name": "x", "type": "space", "unit": "micrometer" }
>             ]
>           }
>         ],
>         "datasets": [
>           {
>             "path": "s0",
>             "coordinateTransformations": [
>               {
>                 "type": "identity",
>                 "input": { "path": "s0" },
>                 "output": { "name": "physical" }
>               }
>             ]
>           }
>         ],
>         "coordinateTransformations": [
>           {
>             "type": "translation",
>             "input": { "name": "world" },
>             "output": { "name": "physical" },
>             "translation": [10, 20]
>           }
>         ]
>       }
>     ]
>   }
> }
> ```
> 
> the coordinateTransform should be ignored when outputing s0 to physical as s0 is already in the physical coordinate system.

To be honest this implied a larger change to fix than I was expecting, because it means we can't just apply all the transforms in `coordinateTransformations` in order anymore. So it's another place where <0.6 differs from 0.6. I left some details in https://github.com/google/neuroglancer/pull/1100/commits/6836ae74a8f0cc7bcd03ca4ca7dbb490fc855267 commit message but happy to discuss more.

chrisj (approved):
looks good! thanks @seankmartin

seankmartin:
> looks good! thanks @seankmartin

Thank you very much for the review Chris, appreciate you spotting those problems, I'll merge it in now

PR #1103, opened 2026-09-04
fix: find groups in global search of command palette

### LINKED ISSUE #1102: Command Palette does not allow searching by group label
1. visit: https://neuroglancer-demo.appspot.com
2. open the command palette
3. enter "toggle layer vis"

Expected:
Results list should show a single entry: "Toggle Layer Visibility"

Actual:
Result list is empty

seankmartin:
Should fix #1102. Extracts the search and rank out of the catalog, and then uses the same search and rank across both the catalog commands and the catalog groups. 

Also fixes another smaller thing which is to expose the open command palette command to the catalog of commands, and just hide that command in the palette. It's an available command so I think it should be in the catalog, but showing it in the palette doesn't make much sense since the palette is already open.


chrisj (approved):
thanks!

PR #1107, opened 2026-09-11
Fix Safari state share clipboard copy

### LINKED ISSUE #872: Share state button does not copy in Safari
The "Share" button for Neuroglancer states does not actually put the state in the clipboard in Safari. All of the other copy buttons work as expected, for example lists of root ids or positions.

Developer tools shows the following error:

```
[Error] Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
	(anonymous function) (main.b12594c7a7ed221d.js:2209:31848)
```

at `state_share.ts:108`.

fatih-uzlmz:
Fixes #872.

Safari requires clipboard writes to begin within the user-activation handler. The state share flow previously called `navigator.clipboard.writeText` only after the state-server request completed, at which point Safari rejected it with `NotAllowedError`.

This change:

- starts `navigator.clipboard.write` during the Share button click;
- supplies the asynchronously generated link through a promised `ClipboardItem` blob;
- displays the generated link with an explicit Copy button if automatic clipboard access is denied; and
- adds regression coverage for pending data, state-generation failures, and clipboard failures.

Tested with:

- `npx vitest --run src/util/clipboard.spec.ts`
- `npm run typecheck`
- `npm run lint:check`
- `npm run format:check`
- manual Safari and Chrome testing with a delayed local state-server response


chrisj (approved):
thanks @fatih-uzlmz

chrisj:
@fatih-uzlmz was adding "displays the generated link with an explicit Copy button if automatic clipboard access is denied; and" necessary from your testing or was it added defensively? If the former, is that related to the time it takes for the state server to respond?

seankmartin (approved):

